PR #133, opened 2022-05-10
fix quickstart redirects

olaszakos:
Old quickstart URL's don't work since their redirect is incorrect.

`/docs/quickstart/quickstart-intro.html` incorrectly redirects to `/docs/current/developer-docs/quickstart/hello10minutes` which doesn't exist.


aisconnolly:
Closing as they were caught in an earlier PR. Thanks!

PR #166, opened 2022-05-17
Fetch the cli-ref docs from the GitHub repository using submodules

hugolgst:
Initialising the external repository fetching using submodules and symlinks. Symlinks eliminate the need for any CI job/script to move the files to the correct location.

#### How to set fetch MD files from an external repository:
1. `git submodule add <git url> submodules/<name>`
2. **OPTIONAL**: add `branch = <name>` to fetch from a specific branch, then `git submodule update --remote`
3. `cd /docs/<path> && ln -s ../../<depends on path>/submodules/motoko/doc/md motoko`

Documentation is now linked and you can use files the standard way in `/sidebars.js`


crusso:
I'll investigate the motoko doc issues today. Longer term, would it be possible to add a build step so we can generate the markdown rather that have it reside  in the git submodule. I'd prefer that for the motoko base library, and motoko itself, since the sources are in mo/adoc and checking in generated files is not great.

docs/developer-docs/build/languages/motoko-index.md
```diff
@@ -0,0 +1,163 @@
...
+ }
+ ```
+
+ Because `value` was declared `stable`, the current state, *n*, of the service is retained after the upgrade. Counting will continue from *n*, not restart from `0`.
+
+ Because the new interface is compatible with the previous one, existing clients referencing the canister will continue to work, but new clients will be able to exploit its upgraded functionality (the additional `reset` function).
+
+ For scenarios that can�t be solved using stable variables alone, Motoko provides user-definable upgrade hooks that run immediately before and after upgrade, and allow you to migrate arbitrary state to stable variables.
+
+ ## And more �?
```
crusso, line 157: What's with the weird chars?

docs/developer-docs/build/languages/motoko-index.md
```diff
@@ -0,0 +1,163 @@
...
+
+ Motoko provides many other developer productivity features, including subtyping, arbitrary precision arithmetic and garbage collection.
+
+ Motoko is not, and is not intended to be, the only language for implementing canisters. If it doesn�t suit your needs, there is a canister development kit (CDK) for the Rust programming language. Our goal is to enable any language (with a compiler that targets WebAssembly) to be able to produce canisters that run on the Internet Computer and interoperate with other, perhaps foreign, canisters through language neutral Candid interfaces.
```
crusso, line 161: More weird chars here.

docs/developer-docs/build/languages/motoko-index.md
```diff
@@ -0,0 +1,163 @@
...
+ var sum = 0;
+ iterTree<Nat>(tree, func (leaf) { sum += leaf });
+ sum
+ ```
+
+ ## Autogenerated IDL files
+
+ A Motoko actor always presents a typed interface to its clients as a suite of named functions with argument and (future) result types.
+
+ The Motoko compiler (and SDK) can emit this interface in a language neutral format called Candid, so other canisters, browser resident code and smart phone apps that support Candid can use the actor�s services. The Motoko compiler can consume and produce Candid files, allowing Motoko to seamlessly interact with canisters implemented in other programming languages (provided they support Candid).
```
crusso, line 81: Another weird char.

crusso: Why does this file exist? It seems to replicate the content of motoko/doc/md/motoko.md from the motoko repo. Is it to avoid a clash with file `motoko` above?

hugolgst: Ah. It got renamed and I missed it.

docs/developer-docs/build/languages/motoko-index.md
hugolgst: Solved by removing and using doc/md/motoko

docs/developer-docs/build/languages/motoko-index.md
hugolgst: Solved by removing and using doc/md/motoko

crusso:
I'm working on fixes in the motoko directory. Part of the problem is that markdown is too braindead to display the kind of tables that adoc can (table with multi-line entries).  I'll have to work around that limitation by re-authoring the adoc to avoid that.

crusso:
@hugolgst, @aisconnolly please try again with the most recent version of the branch. I've fixed some adoc bugs and rewrote some tables that don't surve the translation to md.

crusso:
I see errors remain. 
How do I trigger a rebuild of your branch so I can get some feedback on my changes?

crusso:
Try bumping motoko again - my experiments here https://github.com/dfinity/portal/pull/171 suggest it will not fall over on the motoko md anymore...

crusso:
Once you bump motoko, there's a new file to include in sidebars.js : message-inspection(.md)

crusso:
Eventually, you should update submodule motoko again to pick up new doc for moc 0.6.28 (when that ships with the next version of dfx).

crusso:
I'm curious why has the motoko submodule been removed? Do you have a different solution for pulling in the docs from motoko repo?

hugolgst:
Actually, at that time we had issues with links and couldn't build. I did not have the time to look at it again unfortunately

crusso:
> Actually, at that time we had issues with links and couldn't build. I did not have the time to look at it again unfortunately

I would be happy to try to fix the links on my end if I knew what the problem was. 

The docs could actually use updating so the motoko samples have the correct language attributes for the interpreter in #247, plus the new doc for method inspection is not live.

aisconnolly:
Closed as this task is resolved in a recent PR

PR #213, opened 2022-06-02
update local ledger installation instructions

### LINKED ISSUE #212: Ledger Wasm too big ... again!
Why does this page (https://internetcomputer.org/docs/current/developer-docs/functionality/ledger/ledger-local-setup) yet again point to a ledger canister that is too big and not deployable? This is still a year after launch and yet the public docs do not have a ledger canister that is deployable locally...despite 3 interventions from this forum thread... This is getting embarrassing now. 

Please put in an automated check to ensure the size of the ledger is under 2mb  before tagging. 

https://forum.dfinity.org/t/ledger-wasm-too-big/11274/6

sesi200:
Mirroring this commit: https://github.com/dfinity/ic/commit/fec4a891d46b7ed606bf71f4acca66ce7676212f
Fixes #212

### DIFF
--- a/docs/developer-docs/functionality/ledger/ledger-local-setup.md
+++ b/docs/developer-docs/functionality/ledger/ledger-local-setup.md
@@ -37,7 +37,17 @@ Follow the steps below to deploy your copy of the ledger canister to a local rep
         }
       }
     }
-    ```
+    
+6. Configure your replica to run a `System` subnet. Modify `dfx.json` to include:
+     ```json
+     {
+       "defaults":{
+         "replica": {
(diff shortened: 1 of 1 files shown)


domwoe:
Hey @sesi200 ,

I added this as a note in step 9. So, please remove that note. I added it as a note initially because
1) not everyone uses dfx > 0.10.0
2) We might change the commit version such that it is not needed.

But given #212 I think this approach is better.

sesi200:
Is removed, thanks for notifying me.

aisconnolly (approved):

PR #223, opened 2022-06-06
fix samples page links

olaszakos:
The page https://internetcomputer.org/samples/ has an issue where the "Developer Docs" links don't work: 
image

## Steps to reproduce:
1. go to https://internetcomputer.org
2. click `Sample code` in the top navbar
3. try the "Developer Docs" buttons, they work (notice there is no trailing slash in the URL)
4. refresh the page - notice Netlify added a trailing slash to the URL
5. try the "Developer Docs" buttons, they lead to 404

Proposed fix: use absolute path URL's for internal routing so the target pages don't depend on the current URL (ie. whether or not there is a trailing slash).

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,0 +1,3 @@
+{
+  "singleQuote": false
+}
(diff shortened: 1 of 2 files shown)


sesi200 (approved):
Thanks for the steps to reproduce!

PR #244, opened 2022-06-14
Update basic-concepts.md

mikhail-turilin-dfinity:
Highlighted that do blocks need enclosing parens.


crusso:
These are transitional copies of the Motoko generated docs and should not be edited. I'll fix the sources instead and then close this PR.

crusso:
Fixed in motoko doc branch here https://github.com/dfinity/motoko/commit/5c99f088131338854d413cecbf072ba4479cb426

mikhail-turilin-dfinity:
I agree that this might be not the best place to explain nested
expressions.

However, I don’t think regular developers will read the grammar definition.
Maybe we can find another place to explain the details of “do” blocks.

PR #249, opened 2022-06-16
typo in conversion rate

sesi200:


docs/developer-docs/deploy/computation-and-storage-costs.md
```diff
@@ -28,7 +28,7 @@ Note: System API calls are just like normal function calls from the WebAssembly
 
 Cycles Cost per Transaction (as of July 26, 2021)
 
-The $USD cost for transactions below is based on the above cycle costs. 1 XDR is equal to 1 Trillion cycles. As of July 26, 2021, the exchange rate for 1 XDR = $1.42. The exchange rate for USD \<\> XDR may vary and it will impact the conversion rate. For XDR exchange rates please visit: <https://www.imf.org/external/np/fin/data/rms_sdrv.aspx>
+The $USD cost for transactions below is based on the above cycle costs. 1 XDR is equal to 1 Trillion cycles. As of July 26, 2021, the exchange rate for 1 XDR = $1.42. The exchange rate for USD.XDR may vary and it will impact the conversion rate. For XDR exchange rates please visit: <https://www.imf.org/external/np/fin/data/rms_sdrv.aspx>
```
dprats, line 31: i could be ignorant but is "USD.XDR" a financial shorthand for currency? I was not aware... but maybe we should edit to make it clearer such as "the exchange rate between USD and XDR"?

sesi200: welp... I thought that was the canonical way to say it. My broker at least uses that format. Looking further on [the mother of all interwebz knowledge](https://en.wikipedia.org/wiki/Currency_pair) it says that the 'normal' way to express this is with a `/`. Should we use that then? I just would like to get off the `<` and `>` symbols as that clashes with html codes and broke rendering.

dprats: I would use `/`

PR #271, opened 2022-06-28
Example: demonstrate use of remark-code-import codeblock file inclusion 

crusso:
Example PR to show use of remark-code-import plugin enabled in #292

PR #292, opened 2022-07-06
Plugin remark-code-import

crusso:
Enable appropriate version of remark-code-import plugin to enable file inclusion from code blocks:
E.g.

<pre>
## Native canister smart contract support

Motoko has native support for Internet Computer canister smart contracts.

A canister smart contract (or canister for short) is expressed as a Motoko actor. An actor is an autonomous object that fully encapsulates its state and communicates with other actors only through asynchronous messages.

``` motoko name=counter file=./Counter.mo
```

This section of code defines the private state of the actor.

``` motoko name=counter file=./Counter.mo#L3
```

While, this section of code defines its shared, public functions:

``` motoko name=counter file=./Counter.mo#L5-L8
```
</pre>

NB: versions of remark-code-import newer than 0.4.0 may require ESM style modules and are no compatible with docusaurus.

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -48,6 +48,7 @@ const config = {
         path: "samples",
         routeBasePath: "samples",
         sidebarPath: require.resolve("./sidebarsSample.js"),
+        remarkPlugins: [import("remark-code-import")],
       },
     ],
     customDocusaurusPlugin,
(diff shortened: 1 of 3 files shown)


wackyleo459 (approved):
Everything looks great, as long as the plugin is doing what's expected.

The npm version specified will ensure the older version is installed, and not the latest.

Can you point to a specific code example where this is utilized, if there had been any changes made within any markdown files?

I am curious how the path is referenced for github submodule files (now that's working!).

crusso:
Cool. Thanks for checking. Submodule shouldn't be a problem since the doc dir in a module is just symlinked into the correct place. I'll put the examples in a subdirectory, I expect.

For an example, see motoko.md in the separate  draft PR #271 (not to be merged, just for show)

PR #320, opened 2022-07-19
Refactor: patches from dfinity/motoko

rvanasa:
Backfill from dfinity/motoko#3353. Should have no impact on current app functionality.

@chenyan-dfinity


static/load_moc.ts
```diff
@@ -7,22 +7,16 @@ async function addPackage(name, repo, version, dir) {
...
-        const content = await (await fetch(base_url + f.name)).text();
-        const stripped = name + f.name.slice(dir.length + 1);
-        fetchedFiles.push(stripped);
-        Motoko.saveFile(stripped, content);
-      })();
-      promises.push(promise);
+      const content = await (await fetch(base_url + f.name)).text();
+      const stripped = name + f.name.slice(dir.length + 1);
+      fetchedFiles.push(stripped);
+      await Motoko.saveFile(stripped, content);
```
chenyan-dfinity, line 16: Do we know if `Motoko` module methods are all async calls? I was under the impression that they are synchronous?

chenyan-dfinity:
Also patch in the `className` fix?

static/load_moc.ts
rvanasa: I added the "await" keyword in case these calls become asynchronous due to using a service worker or anything similar (as in the case of Motoko Playground).

chenyan-dfinity: Okay, so it's harmless to put `await` anywhere, right?

chenyan-dfinity (approved):
LGTM.

sesi200 (approved):

static/load_moc.ts
rvanasa: Yep, as long as the return value isn't expected to be a "[thenable](https://www.stefanjudis.com/today-i-learned/await-in-async-functions-works-for-any-thenable/)."

PR #324, opened 2022-07-21
Fix the dead link to Reproducible canisters

Takayuki-Suzuki:

### DIFF
--- a/docs/concepts/trust-in-canisters.md
+++ b/docs/concepts/trust-in-canisters.md
@@ -12,7 +12,7 @@ The answer to this question has two separate dimensions:
 
 ## The Canister does what it is Supposed to do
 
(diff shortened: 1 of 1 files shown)


docs/concepts/trust-in-canisters.md
```diff
@@ -12,7 +12,7 @@ The answer to this question has two separate dimensions:
...
+The correct behavior of a canister can be checked in two steps. First, inspect the source code used to generate the Wasm code deployed in a canister to ensure that it implements the expected/claimed functionality, and only this functionality. Second, ensure that the Wasm module the canister runs, has indeed been generated from the claimed source code. Here, reproducibility of the build is crucial: the developer should have constructed the Wasm module so that precisely the same Wasm can be rebuilt from scratch. The user can then compare the hash of the rebuilt Wasm module with the module hash reported by the IC. Developers and users can find guidance on ensuring reproducibility in [Reproducible canisters](https://internetcomputer.org/docs/current/developer-docs/build/backend/reproducible-builds).
```
sesi200, line 15:
```suggestion
The correct behavior of a canister can be checked in two steps. First, inspect the source code used to generate the Wasm code deployed in a canister to ensure that it implements the expected/claimed functionality, and only this functionality. Second, ensure that the Wasm module the canister runs, has indeed been generated from the claimed source code. Here, reproducibility of the build is crucial: the developer should have constructed the Wasm module so that precisely the same Wasm can be rebuilt from scratch. The user can then compare the hash of the rebuilt Wasm module with the module hash reported by the IC. Developers and users can find guidance on ensuring reproducibility in [Reproducible canisters](../developer-docs/build/backend/reproducible-builds.md).
```
We can actually make this a relative link so in case of another migration it would stay updated

sesi200 (approved):
Thank you very much for pointing this out!

sesi200:
I think if you merge master this will fix the build errors

Takayuki-Suzuki:
@sesi200 I just found it by chance:) You mean merge master into my branch? Looks like my branch is not behind by master

sesi200:
Ugh, must be something with master. Let me ping the relevant folks...

PR #337, opened 2022-07-25
Kyle/agent intro

krpeacock:


docs/developer-docs/build/using-an-agent.mdx
```diff
@@ -0,0 +1,56 @@
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+import CodeBlock from '@theme/CodeBlock';
+
+# Using an Agent
+
+If you have a canister running, either on your local machine or live on the Internet Computer, you will have two main ways to interact with your canister smart contract. You can talk to the canister using the v2 API using an `agent` that follows the interface specification, or you can use the canister's http interface.
```
lwshang, line 7: > or you can use the canister's http interface.

Does this refer to the common practice that implement a `http_request` method in backend canister and use it from frontend?

docs/developer-docs/build/using-an-agent.mdx
```diff
@@ -0,0 +1,56 @@
...
+
+A `call` to the Internet Computer can take two common forms - an `update` or a `query`. The `agent` submits a POST request to `/api/v2/canister/<effective_canister_id>/call`, and includes the following components: 
+* `request_type`
+* Authentication
+    * `sender`, `nonce`, and `ingress_expiry`
+* `canister_id`
+* `method_name`
+* `arg` - the rest of the payload
+
+By knowing the Candid interface of the canister, the `agent` will assemble the `"arg"` by `cbor`-encoding the data from the client application, and ensuring it matches the Candid interface for the method it will be calling. That encoded data will be turned into a buffer.
```
lwshang, line 25: IIUC, we `cbor` encode the whole authentication envelope, not only on the `arg`.

docs/developer-docs/build/using-an-agent.mdx
```diff
@@ -0,0 +1,56 @@
...
+### 1. Structuring Data
+
+A `call` to the Internet Computer can take two common forms - an `update` or a `query`. The `agent` submits a POST request to `/api/v2/canister/<effective_canister_id>/call`, and includes the following components: 
+* `request_type`
+* Authentication
+    * `sender`, `nonce`, and `ingress_expiry`
+* `canister_id`
+* `method_name`
+* `arg` - the rest of the payload
+
```
lwshang, line 24: `update` also need a `request_id` field which is basically a hash of all request information. Should we include it in this guide?

docs/developer-docs/build/using-an-agent.mdx
```diff
@@ -0,0 +1,56 @@
...
+Calls to the Internet Computer always need to have a cryptographic identity attached. That identity will either be Anonymous or Authenticated, using a cryptographic signature. Since identities are required, canisters can use the identity attached to a call to decide how to respond to that call. This enables contracts to use those identities for other purposes. 
+
+#### Accepted Identities
+
+The IC accepts calls using the following types of identities:
+
+* Ed25519
+* Secp256k1
+* ECDSA with a P-256 curve
```
lwshang, line 43: For the self-authenticating principle, should we use the words in the spec?
> ### Ed25519 and ECDSA signatures
> Plain signatures are supported for the schemes
> * [Ed25519](https://ed25519.cr.yp.to/index.html) or
> * [ECDSA](https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.186-4.pdf) on curve P-256 (also known as secp256r1), using SHA-256 as hash function, as well as on the Koblitz curve secp256k1.

docs/developer-docs/build/using-an-agent.mdx
```diff
@@ -0,0 +1,56 @@
...
+
+When encoding these identities as a Principal, agents attach a suffix byte, indicating whether the identity is self-authenticating or anonymous. 
+
+A self-authenticating identity using one of the above curves will have a suffix of 2, while anonymous identities receive a suffix of 4. The default anonymous identity is derived from an empty byte array, apart from the suffix 4. It resolves to `"2vxsx-fae"`, in its textual encoding.
```
lwshang, line 47:
```suggestion
A self-authenticating identity using one of the above curves will have a suffix of 2.

While the anonymous identity is a single byte 4. It resolves to `"2vxsx-fae"`, in its textual encoding.
```

Only `0x04` is anonymous identity. Other bytes ending in 4 is not categorized so far.

docs/developer-docs/build/using-an-agent.mdx
krpeacock: Yes, although it's an official interface via the boundary nodes, so I'm choosing to refer to it as a way that canisters can be interacted with

docs/developer-docs/build/using-an-agent.mdx
krpeacock: Yes, I'll add it

docs/developer-docs/build/using-an-agent.mdx
krpeacock: I modified it slightly, but yes, the precision is better

lwshang (approved):

aisconnolly (approved):

docs/developer-docs/build/using-an-agent.mdx
```diff
@@ -0,0 +1,63 @@
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+import CodeBlock from '@theme/CodeBlock';
+
+# Using an Agent
+
+If you have a canister running, either on your local machine or live on the Internet Computer, you will have two main ways to interact with your canister smart contract. You can talk to the canister using the v2 API using an `agent` that follows the interface specification, or you can use the canister's http interface.
```
ggreif, line 7: HTTP and CBOR are still acronyms and should be capitalised as such

docs/developer-docs/build/using-an-agent.mdx
```diff
@@ -0,0 +1,63 @@
...
+
+The IC accepts calls using the following types of signatures in identities:
+
+* Ed25519 and ECDSA signatures
+  * Plain signatures are supported for the schemes
+* Ed25519 or ECDSA on curve P-256 (also known as secp256r1)
+  * using SHA-256 as a hash function
+  * using the Koblitz curve in secp256k1.
+
+When encoding these identities as a Principal, agents attach a suffix byte, indicating whether the identity is self-authenticating or anonymous.
```
ggreif, line 50: `Principal`

docs/developer-docs/build/using-an-agent.mdx
ggreif: #348

docs/developer-docs/build/using-an-agent.mdx
ggreif: #348

PR #412, opened 2022-08-18
home hero update

olaszakos:


artkorotkikh-dfinty (approved):
LGTM as per Dom's request

PR #445, opened 2022-08-26
update links from tiles

dprats:
Goal of this branch is to make sure that every tile in the "whats cool" section of the homepage has a link to a a landing page or wiki page. Where o wiki page existed, I created it.


artkorotkikh-dfinty (approved):
lgtm

PR #474, opened 2022-09-09
Update deploying-your-first-bitcoin-dapp.md

dfx-json:
one line that should be appearing as a code block is not appearing as a code block. I believe this is due to a tab being used to prefix the line instead of spaces.

### DIFF
--- a/docs/samples/deploying-your-first-bitcoin-dapp.md
+++ b/docs/samples/deploying-your-first-bitcoin-dapp.md
@@ -23,7 +23,7 @@ and [Bitcoin API](https://internetcomputer.org/docs/current/references/ic-interf
 
 3. Initialize the git submodules
 
-	git submodule update --init --recursive
+        git submodule update --init --recursive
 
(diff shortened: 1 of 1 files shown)


docs/samples/deploying-your-first-bitcoin-dapp.md
```diff
@@ -23,7 +23,7 @@ and [Bitcoin API](https://internetcomputer.org/docs/current/references/ic-interf
 
 3. Initialize the git submodules
 
-	git submodule update --init --recursive
+        git submodule update --init --recursive
```
dprats, line 26: i may be crazy, but I do not see a difference in the change. Look like identical lines. Can you help me understand @dfx-json ?

sesi200: tabs vs spaces. It actually corrects the formatting.

sesi200 (approved):

PR #546, opened 2022-09-27
solve mismatched types error

KevynTang:
original search function leads to an `mismatched types` compile error, maybe modify it like this is the correct way to solve the issue?


aisconnolly:
@MurielPinho  or @olaszakos does it help the search?

MurielPinho:
> @MurielPinho or @olaszakos does it help the search?

This is related to the cdk-rs search function, would need @sesi200 opinion

sesi200:
`cdk-rs` is not my forte, I'll delegate this one to @lwshang

adamspofford-dfinity:
Thank you for noticing the error! But your version has the very thing that ManualReply exists to remove the need for - cloning borrowed values. The only actual needed change is moving the final return value inside the `with` closure.

KevynTang:
> Thank you for noticing the error! But your version has the very thing that ManualReply exists to remove the need for - cloning borrowed values. The only actual needed change is moving the final return value inside the `with` closure.

Thank you! Learned a lot of things via this PR 🥳

PR #567, opened 2022-09-29
move search to a netlify function

olaszakos:


docusaurus.config.js
```diff
@@ -28,12 +28,20 @@ const votingRewardsPlugin = require("./plugins/voting-rewards");
 const isDeployPreview =
   !!process.env.NETLIFY && process.env.CONTEXT === "deploy-preview";
 
+const isDfxBuild = !!process.env.DFX_VERSION;
+
+if (isDfxBuild) {
+  console.log(`Canister id: ${process.env.CANISTER_ID}`);
+}
```
MurielPinho, line 35: Probably left over from local debugging

olaszakos: I left this here preemptively because the IC build might not work correctly due to reliance on the env variable. Will remove in a future update if works as expected.

dprats (approved):

PR #585, opened 2022-10-05
Green updates

aisconnolly:


olaszakos:
the new green.png should be higher resolution if possible.

olaszakos (approved):

PR #600, opened 2022-10-10
/sns page

olaszakos:


src/pages/sns.tsx
```diff
@@ -0,0 +1,480 @@
...
+              <br className="hidden md:inline" />
+              Be unstoppable with an SNS.
+            </motion.p>
+            <motion.p
+              className="mb-0 flex flex-col items-start sm:flex-row gap-6 md:gap-8"
+              variants={transitions.item}
+            >
+              <Link
+                className="button-outline"
+                href="/docs/current/developer-docs/integrations/sns/sns-intro#how-to-get-and-maintain-a-dao---different-options"
```
LaraAS, line 128: I thinks this should point to https://internetcomputer.org/docs/current/tokenomics/sns/sns-intro-tokens 
Note that the currently linked page is not yet public

src/pages/sns.tsx
```diff
@@ -0,0 +1,480 @@
...
+                Liquid democracy. When a proposal is approved by the community,
+                the SNS implements changes to the code and upgrades itself
+                autonomously. This ensures permisionless and decentralized
+                governance of an SNS DAO.{" "}
+                <span className="italic">
+                  No single entity can stop the process.
+                </span>
+              </motion.p>
+              <MotionLink
+                className="link-external"
```
LaraAS, line 210: This currently links back to the same page. Should we rather remove the "learn more" for now?

src/pages/sns.tsx
```diff
@@ -0,0 +1,480 @@
...
+                that community members can stake to claim governance rights or
+                use to purchase services. The same token can be used to reward
+                community contributions to the SNS DAO. Such contributions can
+                range from engaging with the dapp to writing code, designing,
+                promoting or participating in governance.{" "}
+                <span className="italic">No fiat required.</span>
+              </motion.p>
+              <MotionLink
+                className="link-external"
+                href="https://internetcomputer.org/docs/current/tokenomics/sns/sns-intro-tokens"
```
LaraAS, line 232: Should this link to https://internetcomputer.org/docs/current/tokenomics/sns/tokenomics ? 
If I press on the current link, I end up on the top level "SNS" page..

src/pages/sns.tsx
```diff
@@ -0,0 +1,480 @@
...
+              >
+                Learn more
+              </MotionLink>
+            </Card>
+            <Card title="Scalable" icon={<IconScalable aria-hidden="true" />}>
+              <motion.p
+                className="mb-6 tw-paragraph"
+                variants={transitions.item}
+              >
+                Infnitely scalable. The Internet Computer is the only blockchain
```
LaraAS, line 243:
```suggestion
                Infinitely scalable. The Internet Computer is the only blockchain
```

src/pages/sns.tsx
```diff
@@ -0,0 +1,480 @@
...
+          />
+          <AnimateSpawn
+            className="mx-auto text-center sm:w-6/12 mb-16 md:mb-8"
+            variants={transitions.container}
+          >
+            <motion.h2
+              className="tw-heading-4 md:tw-heading-60 mb-3 md:mb-8"
+              variants={transitions.item}
+            >
+              Turn your App into a dApp with an SNS
```
LaraAS, line 281:
```suggestion
              Turn your app into a dapp with an SNS
```

src/pages/sns.tsx
```diff
@@ -0,0 +1,480 @@
...
+            </motion.h2>
+            <motion.p
+              className="tw-lead-sm md:tw-lead mb-8"
+              variants={transitions.item}
+            ></motion.p>
+            <MotionLink
+              variants={transitions.item}
+              className="button-outline-white"
+              href="/docs/current/developer-docs/integrations/sns/sns-intro#how-to-get-and-maintain-a-dao---different-options"
```
LaraAS, line 290: This page is not yet public.
Recommend to link to https://internetcomputer.org/docs/current/tokenomics/sns/sns-intro-tokens in the meantime

src/pages/sns.tsx
```diff
@@ -0,0 +1,480 @@
...
+              <div className="md:w-8/10">
+                <h3 className="tw-heading-6 md:tw-heading-5 mb-4">
+                  Not an ICO
+                </h3>
+                <p className="tw-paragraph mb-10">
+                  No funds are handed over to developers. The creation of an SNS
+                  is initiated by a decentralized entity (
+                  <Link
+                    className="text-infinite underline hover:no-underline hover:text-black"
+                    href="https://nns.ic0.app/"
```
LaraAS, line 415: I think we mix here different meanings of "NNS".
In the sentence we refer to the NNS as a DAO (i.e., the DAO that runs the IC).
However, we link to _one_ frontend that allows to interact with the NNS. 
I propose we rather link to an explanation of what the NNS is.

src/pages/sns.tsx
```diff
@@ -0,0 +1,480 @@
...
+                variants={transitions.item}
+              >
+                What an SNS DAO is <span className="text-razzmatazz">NOT</span>
+              </motion.h2>
+              <div className="md:w-8/10">
+                <h3 className="tw-heading-6 md:tw-heading-5 mb-4">
+                  Not an ICO
+                </h3>
+                <p className="tw-paragraph mb-10">
+                  No funds are handed over to developers. The creation of an SNS
```
LaraAS, line 411: Is this really true? developers can get initial neurons and put a considerable number of SNS tokens there

src/pages/sns.tsx
```diff
@@ -0,0 +1,480 @@
...
+                  </Link>
+                  ), and ICP raised by the decentralization sale is allocated to
+                  the treasury of the newly born SNS DAO.
+                </p>
+                <h3 className="tw-heading-6 md:tw-heading-5 mb-4">
+                  Not an IPO
+                </h3>
+                <p className="tw-paragraph">
+                  This is not a liquidity exit or a public listing. It’s the
+                  start of a new journey where you as the initial creator and
```
LaraAS, line 427: Initial creator of what?

It is a bit unclear to me what we mean here.
Specifically, we explicitly don't want to say the creator of the DAO, right?
I think it can currently be misunderstood in this way.

src/pages/sns.tsx
```diff
@@ -0,0 +1,480 @@
...
+                  prose-h3:tw-heading-6 prose-h3:md:tw-heading-5 prose-h3:mb-4
+                  prose-p:tw-paragraph prose-p:mb-3
+                  prose-ul:mb-4 prose-ul:list-none prose-ul:pl-0 prose-ul:tw-paragraph
+                  prose-li:bg-[url('/img/checkmark.svg')] prose-li:bg-no-repeat prose-li:bg-left-top prose-li:pl-8 prose-li:my-3 prose-li:leading-6
+                  "
+              >
+                <h3>Decentralization Sale</h3>
+                <p>
+                  Funds are put in a treasury, a portion of which is allocated
+                  to the founders in the form of locked neurons for developing a
```
LaraAS, line 454: I think neurons are not locked but rather tokens are locked in neurons. Should we reformulate accordingly?

src/pages/sns.tsx
```diff
@@ -0,0 +1,480 @@
...
+                className="
+                  md:w-8/10
+                  prose
+                  prose-h3:tw-heading-6 prose-h3:md:tw-heading-5 prose-h3:mb-4
+                  prose-p:tw-paragraph prose-p:mb-3
+                  prose-ul:mb-4 prose-ul:list-none prose-ul:pl-0 prose-ul:tw-paragraph
+                  prose-li:bg-[url('/img/checkmark.svg')] prose-li:bg-no-repeat prose-li:bg-left-top prose-li:pl-8 prose-li:my-3 prose-li:leading-6
+                  "
+              >
+                <h3>Decentralization Sale</h3>
```
LaraAS, line 451: This reads to me as "An SNS DAO is a decentralization sale".
This is not accurate as the sale is just part of it or maybe even just the way how to launch the DAO. 
Is there a way that we can avoid this misinterpretation?

src/pages/sns.tsx
```diff
@@ -0,0 +1,480 @@
...
+                <p>
+                  Funds are put in a treasury, a portion of which is allocated
+                  to the founders in the form of locked neurons for developing a
+                  service. Launching an SNS DAO requires the following:
+                </p>
+                <ul>
+                  <li>An existing dapp</li>
```
LaraAS, line 458: I think technically this is not a requirement. 
You could launch an SNS without a dapp and let the SNS create the dapp.
Do we still want to phrase it like this as we think this will be the prevalent way people will use it?

LaraAS (commented):
Thanks for the draft. I added some comments.
I think some of them should be addressed before releasing this page, mostly links to pages that should not yet be public.

src/pages/sns.tsx
harpa12: Done. Please create PR to change the link when the docs are live.

src/pages/sns.tsx
harpa12: Removing this link for now until all docs are live.

src/pages/sns.tsx
harpa12: We want App and dApp to stand out. To compromise, I suggest capitalizing "Your" instead because it's a title.

src/pages/sns.tsx
harpa12: How about no funds are "directly" handed over...? The way I understand it, the NNS has to approve the configuration of the dissolve delay of a neuron. I assume that multiple zero dissolve proposal would not be approved???

src/pages/sns.tsx
harpa12: Agreed that it is unclear. I will edit to: initial creator of a service.

src/pages/sns.tsx
harpa12: Yes, good catch. Reformulating to "locked tokens in voting neurons"

src/pages/sns.tsx
harpa12: I think the best solution here is to remove the subtitle "Decentralization Sale"

src/pages/sns.tsx
harpa12: I think we should keep it as such - otherwise it's too close to an ICO. We certainly don't want to promote empty SNSs with great promise coupled with zero dissolve delay proposals. WDYT?

src/pages/sns.tsx
LaraAS: Yes I would prefer that even though it is still a bit unclear probably what we mean by this.
But I think it is a step in the right direction!

src/pages/sns.tsx
LaraAS: This is fine for me. Resolving this thread. Thank you.

csepreghy (approved):

PR #634, opened 2022-10-18
Correct `href` of Bitcoin Integration Tutorials I and II

Benjamin-Loison:
Redirecting currently incorrectly when clicking on Bitcoin Integration Tutorials I and II thumbnails on [Bitcoin Integration](https://internetcomputer.org/bitcoin-integration/) webpage.


sesi200 (approved):
Thank you very much for the fix! And nice to still see you around, Benjamin!

PR #644, opened 2022-10-20
roadmap entry for canister groups

bogwar:


roadmap/1_core-protocol/upcoming/1_canister-groups/canister-groups.md
```diff
@@ -0,0 +1,13 @@
+-
+title: Canister groups
+links:
+  Forum Link: https://forum.dfinity.org/t/canister-groups/16015
+  Proposal: 
+eta:
+is_community: false
+---
+
+Communication between canisters on the same subnetwork is very fast since the messages that they exchange do not need to go through consensus. Since intercanister communication latency is less of a concern, one can scale Dapps through multi-canister architectures, as long as the canisters are guaranteed to be collocated.
```
sesi200, line 10:
```suggestion
Communication between canisters on the same subnet is very fast since the messages that they exchange do not need to go through consensus. Since inter canister communication latency is less of a concern, one can scale dapps through multi-canister architectures, as long as the canisters are guaranteed to be collocated. 
```

roadmap/1_core-protocol/upcoming/1_canister-groups/canister-groups.md
```diff
@@ -0,0 +1,13 @@
...
+links:
+  Forum Link: https://forum.dfinity.org/t/canister-groups/16015
+  Proposal: 
+eta:
+is_community: false
+---
+
+Communication between canisters on the same subnetwork is very fast since the messages that they exchange do not need to go through consensus. Since intercanister communication latency is less of a concern, one can scale Dapps through multi-canister architectures, as long as the canisters are guaranteed to be collocated. 
+
+Since upcoming load-balancing mechanisms need to move canisters between subnets, we need to avoid splitting up canisters that are part of the same Dapp.   This feature would allow a Dapp developer to explicitly indicate which canisters "belong" together and which should always be located on the same subnet.
```
sesi200, line 12:
```suggestion
Since upcoming load-balancing mechanisms need to move canisters between subnets, we need to avoid splitting up canisters that are part of the same dapp. This feature would allow a dapp developer to explicitly indicate which canisters "belong" together and which should always be located on the same subnet.
```

roadmap/1_core-protocol/upcoming/1_canister-groups/canister-groups.md
sesi200, line 10: `since the messages that they exchange do not need to go through consensus.` This is too easy to misunderstand and needs to be reformulated. This sounds like a single replica decides what happens.

sesi200 (changes requested):

manudrijvers:
The path should not be `roadmap/1_core-protocol/upcoming/1_canister-groups/canister-groups.md` but `roadmap/1_core-protocol/upcoming/canister-groups.md`. Please also make sure things look good on the preview that is generated for each PR (currently it's not actually shown due to the path issue).

roadmap/1_core-protocol/upcoming/1_canister-groups/canister-groups.md
Dfinity-Bjoern: Maybe it is sufficient to just state the fact that communication on the same subnet is faster than between subnets – all that is needed for the motivation of this feature is *that* this is the case, not necessarily why.

roadmap/1_core-protocol/upcoming/1_canister-groups/canister-groups.md
Dfinity-Bjoern: I'd probably rephrase the last sentence as "Canister groups will [...]" – since it's on the roadmap so I would expect it to be built.

Dfinity-Bjoern (approved):
Overall looks good, some suggestions in the comments.

roadmap/1_core-protocol/upcoming/1_canister-groups/canister-groups.md
bogwar: implemented the suggestion to remove the mentioning of consensus and just state the fact.

roadmap/1_core-protocol/upcoming/canister-groups.md
```diff
@@ -0,0 +1,13 @@
+-
```
sesi200, line 1:
```suggestion
---
```

sesi200 (approved):

PR #654, opened 2022-10-24
improve the rust-counter tutorial

KevynTang:
fixed those confusing inconsistencies on the file paths and canister names between this tutorial and a real project.


sesi200 (approved):
Thank you so much! It's so hard finding all the possible places of that change

PR #656, opened 2022-10-24
Create financial-spam-prevention.md

bjoernek:
This commit adds financial spam prevention as a pending/upcoming project under core protocol to the public roadmap.

### DIFF
--- a/roadmap/1_core-protocol/upcoming/financial-spam-prevention.md
+++ b/roadmap/1_core-protocol/upcoming/financial-spam-prevention.md
@@ -7,8 +7,8 @@ eta:
 is_community: true
 ---
 
-This change suggests a conversion from a system-based rewards model to a voter-based reward model. Voting rewards are allocated under the assumption that every neuron voted on everything. 
+In the current voting reward set-up, there is a financial incentive for spam because the overall reward pot is always handed out and by submitting governance spam propoals you can skew the voting reward allocation. 
 
-This change also includes moving the governance weight back to 20 without incentivising spam. We propose this occur as part of this proposal.
+This change removes this incentive by allocating voting rewards under the assumption that every neuron voted on everything. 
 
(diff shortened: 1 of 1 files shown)


sesi200 (approved):

roadmap/1_core-protocol/upcoming/financial-spam-prevention.md
```diff
@@ -0,0 +1,14 @@
...
+  Proposal: https://dashboard.internetcomputer.org/proposal/80970
+eta:
+is_community: true
+---
+
+This change suggests a conversion from a system-based rewards model to a voter-based reward model. Voting rewards are allocated under the assumption that every neuron voted on everything. 
+
+This change also includes moving the governance weight back to 20 without incentivising spam. We propose this occur as part of this proposal.
+
+A result of this proposal will be a reduction in total minted ICP due to the fact that some voters do not vote on all proposals or follow a voter for all proposals. In a follow-on proposal, the NNS can determine what to do with that “abandoned” ICP. We suggest an NNS treasury but will leave it to a future proposal to finalize that.
```
manudrijvers, line 14: I would stick to what is agreed upon and not say anything else here. This feature is about exactly what was proposed, and an NNS treasury is a future discussion (and a controversial one), so I think we should not mix things up.

manudrijvers: Even though the problem is pretty well known, I would maybe also add one line explaining that currently there is a financial incentive for spam because the rewards are always handed out, and we remove the incentive by changing that.

bjoernek: Very good points, which are now incorporated.

manudrijvers:
@bjoernek sorry I missed this earlier, but this one is currently in the "core protocol" directory, while I believe it should be under "governance".

bjoernek:
moved to governance folder

Issue #665, opened 2022-10-25
Floating point arithmetic and Decimal


andrew-lee-work:
Add note on floating point arithmetic. It should not be used for finance. On division by zero, f64 gives -inf, inf, or NaN.

Use rust_decimal::Decimal instead.

This is regarding 
`docs/references/security/rust-canister-development-security-best-practices.md`

PR #666, opened 2022-10-25
Add note in security best practices on floats

andrew-lee-work:
Addressing #665 .

### DIFF
--- a/docs/references/security/rust-canister-development-security-best-practices.md
+++ b/docs/references/security/rust-canister-development-security-best-practices.md
@@ -447,6 +447,17 @@ Integers in Rust may overflow. While such overflows lead to panics in the debug 
 
 -   See also the [Rust security guidelines on integer overflows](https://anssi-fr.github.io/rust/04_language.html#integer-overflows).
 
+### Rust: Avoid floating point arithmetic for financial information
+
+#### Security Concern
+
+Floats in Rust may behave unexpectedly. There can be undesirable loss of precision under certain circumstances. When dividing by zero, the result could be `-inf`, `inf`, or `NaN`. When converting to integer, this can lead to unexpected results.
+
+#### Recommendation
+
+Use fixed point numbers instead. For example, [`rust_decimal::Decimal`](https://docs.rs/rust_decimal/latest/rust_decimal/). One still has to manage precision, but it is easier to reason about.
+
+
 ### For expensive calls, consider using captchas or proof of work
 
 #### Security Concern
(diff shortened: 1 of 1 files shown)


daniel-wong-dfinity-org:
LGTM

sesi200 (approved):
Thank you very much! That's a really good point to add to the list

docs/references/security/rust-canister-development-security-best-practices.md
```diff
@@ -447,6 +447,17 @@ Integers in Rust may overflow. While such overflows lead to panics in the debug
...
 
+### Rust: Avoid floating point arithmetic for financial information
+
+#### Security Concern
+
+Floats in Rust may behave unexpectedly. There can be undesirable loss of precision under certain circumstances. When dividing by zero, the result could be `-inf`, `inf`, or `NaN`. When converting to integer, this can lead to unexpected results. (There is no `checked_div` for floats.)
+
+#### Recommendation
+
+Use fixed point numbers instead. For example, [`rust_decimal::Decimal`](https://docs.rs/rust_decimal/latest/rust_decimal/). One still has to manage precision, but it is easier to reason about. Use `checked_div` to handle division by zero.
```
robin-kunzler, line 458: ``` One still has to manage precision, but it is easier to reason about. ```

I think an example would help me to understand why it is easier to reason about. Would it make sense to add an example / or explain _why_ it is easier to reason about the decimals?

robin-kunzler (commented):
@andrew-lee-work Thanks a lot for adding this to the best practices! Looks good to me, have one question, see below.

docs/references/security/rust-canister-development-security-best-practices.md
andrew-lee-work: How about this:

---
Use [`rust_decimal::Decimal`](https://docs.rs/rust_decimal/latest/rust_decimal/) or [`num_rational::Ratio`]( https://docs.rs/num-rational/latest/num_rational/). Use `checked_div` to handle division by zero.

Numbers in common use like 0.1 and 0.2 can be represented more intuitively in Decimal, and can be represented exactly with Ratio. Oddities like `0.1 + 0.2 != 0.3`, which happen with floats in Rust, do not arise with Decimal (see https://0.30000000000000004.com/ ). With Ratio, the desired precision can be made explicit. With either Decimal or Ratio, although one still has to manage precision, the above make arithmetic is easier to reason about.

robin-kunzler: This looks nice, thanks for adding more context!

andrew-lee-work: @robin-kunzler Thanks for the feedback. I've made the update in PR #676. Your approval would be appreciated.

PR #668, opened 2022-10-27
suggest one improvement on rust_profile tutorial

KevynTang:
In the **Update interface description file** part of the [rust profile](https://internetcomputer.org/docs/current/developer-docs/build/cdks/cdk-rs-dfinity/rust-profile/) tutorial, I use Edge browser and I find I always trigger download when clicking the provided (.did file)[https://internetcomputer.org/assets/files/profile-54ab75a0d7a1a7b3b3b38861dba2dd6a.did], in order to make this part more reader friendly, I suggest just putting this piece of code inside the text instead of providing a link (just like [the next tutorial](https://internetcomputer.org/docs/current/developer-docs/build/cdks/cdk-rs-dfinity/rust-counter/)). Besides, the original type declaration file has some meanless code (maybe due to versioning reason?), I also suggest removing them to improve readers' experience.


sesi200:
Thanks for the suggestion, Kevyn! Can we delete the .did file then? I'd prefer not leaving it in if it isn't referenced anymore. CI will catch it if a link breaks because of that

KevynTang:
> Thanks for the suggestion, Kevyn! Can we delete the .did file then? I'd prefer not leaving it in if it isn't referenced anymore. CI will catch it if a link breaks because of that

It should be ok to delete the `profile.did`, I've searched for `profile.did` in the `portal` repo and I think only `profile-rs` tutoral uses this link.

sesi200:
Nice! Can you do that in your branch, please? I don't think I have write-access to it

KevynTang:
> Nice! Can you do that in your branch, please? I don't think I have write-access to it

Done it. 🥳

sesi200 (approved):
Thank you very much! ❤️

PR #784, opened 2022-11-17
added cdk optimizer dependency

letmejustputthishere:

### DIFF
--- a/docs/developer-docs/build/cdks/cdk-rs-dfinity/rust-profile.md
+++ b/docs/developer-docs/build/cdks/cdk-rs-dfinity/rust-profile.md
@@ -33,7 +33,7 @@ Before you start your project, verify the following:
 
     The Rust tool chain must be at version 1.46.0, or later.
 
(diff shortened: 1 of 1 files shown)


sesi200:
AFAIK with the latest versions of dfx the optimizer is not necessary anymore. Does it still work if we leave this out?

letmejustputthishere:
let me find out. if this is the case then instead of adding it to all of the tutorials, we have to remove the step from the quickstart for the rust cdk

ggreif:
> AFAIK with the latest versions of dfx the optimizer is not necessary anymore. Does it still work if we leave this out?

@chenyan-dfinity ^^

letmejustputthishere:
i can confirm that the `shrink WASM module size` step is present even after running `cargo uninstall ic-cdk-optimizer`.

also i digged [this](https://forum.dfinity.org/t/wasm-module-contains-a-function-that-is-too-complex/15407/18?u=cryptoschindler) up in the forum.

if it is true, we should edit/remove the entire `optimize rust` section. will make a pr for that

PR #944, opened 2022-12-23
Included ios integration sample to portal

keplervital:

### DIFF
--- a/src/components/Common/sampleItems.ts
+++ b/src/components/Common/sampleItems.ts
@@ -268,4 +268,23 @@ export const sampleItems: SampleItem[] = [
       livePreview: "https://uf2fn-liaaa-aaaal-abeba-cai.ic0.app/",
     },
   },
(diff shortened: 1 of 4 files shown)


src/components/Common/sampleItems.ts
```diff
@@ -268,4 +268,23 @@ export const sampleItems: SampleItem[] = [
       livePreview: "https://uf2fn-liaaa-aaaal-abeba-cai.ic0.app/",
     },
   },
+  {
+    index: 13,
+    title: "IOS Integration",
+    image: require("../../../static/img/samples/default.gif").default,
```
keplervital, line 274: I've used the default image, but would be great to maybe add another.

sesi200 (commented):
Links to smartcontracts.org are broken. Please use relative links wherever possible

sesi200 (approved):

PR #947, opened 2022-12-23
Fixing the Bitcoin integration pricing table

THLO:
This PR reflect the Bitcoin integration pricing changes, in particular for Bitcoin testnet, in the pricing table.


artkorotkikh-dfinty (changes requested):
@THLO do we need to change package-lock.json for this change? If not, can you remove from PR please? I afraid we might broke something with this and I won't be able to fix it :)

THLO:
> @THLO do we need to change package-lock.json for this change? If not, can you remove from PR please? I afraid we might broke something with this and I won't be able to fix it :)

Good catch! I removed the file from the PR.

PR #985, opened 2023-01-17
Fix sns bottom nav

lopezi:
- Bottom nav at docs/current/developer-docs/integrations/sns/integrate-sns/ledger-integration enters a loop because of current sidebar.js structure. Added a *root page [*needs better description] to make the navigation flow
- Corrected some misspellings along the way

- [✅] You have registered new documents to [`/sidebars.js`](https://github.com/dfinity/portal/blob/master/sidebars.js) if any.


sesi200:
Thank you very much for your PR, @lopezi!

I'm not a fan of having an almost-empty page. What do you think of defining the `Integrate with an SNS` category like this?
```
            {
              type: 'category',
              label: 'Integrate with an SNS',
              link: {
                type: "doc",
                id: "developer-docs/integrations/sns/integrate-sns/index-integration",
              },
              items: [
                { type: "ref", id: "developer-docs/integrations/sns/integrate-sns/index-integration" },
                "developer-docs/integrations/sns/integrate-sns/ledger-integration"
                //                "developer-docs/integrations/sns/integrate-sns/frontend-integration"
              ],
            },
```

lopezi:
Thanks @sesi200 

Fine for me for now. If there's no need for a base page within` integrate-sns` the most important is to keep the nav flowing.

sesi200 (approved):
Nice, thanks!

PR #1032, opened 2023-02-03
doc: Candid Ref: add colon after `service` keyword

Web3NL:
`didc` tool fails to validate `service` code snippets.

`moc` is not consistent about the spaces around the colon.
`service: { }` 
vs
`service : { }`


sesi200 (approved):
Thanks a lot!

PR #1145, opened 2023-02-28
chore: EXC-1193 Add costs for local development

dfinity-berestovskyy:


dsarlis:
Isn't it better to have the absolute numbers for easy consumption by end users? As opposed to having the formula? Like X * 13 / 34?

dfinity-berestovskyy:
> Isn't it better to have the absolute numbers for easy consumption by end users? As opposed to having the formula? Like X * 13 / 34?

I think it's more user friendly representation of data. For example, `100B` vs `261,538,461,538` vs `100B / 13 * 34`. People are not that good at multiplication and division by 13, so glancing at the `100B` and `261,538,461,538` people don't see the linear scaling of costs. Instead they see random numbers, and the details are 3 paragraphs above.

With the explicit formula, not only we give a more precise data (i.e. there is no rounding), but at the same time we explicitly show that those numbers are not random. Seeing the `100B` and `100B / 13 * 34` a clear message about linear scaling, or a signal to look for the details around.

WDYT?

sesi200 (approved):

sesi200:
I actually like the formulaic approach. It's easy enough to estimate what `/ 13 * 34` is (around `* 2.5`) and it punches you in the face with how the logic works so that even the least attentive person will know how prices scale between subnets of different sizes.

PR #1175, opened 2023-03-07
chore: change landing page for motoko docs

dfx-json:
Changes the landing page for the Motoko docs to the "Motoko Introduction".

### DIFF
--- a/sidebars.js
+++ b/sidebars.js
@@ -155,7 +155,7 @@ const sidebars = {
       label: "Motoko Language Tour",
       link: {
         type: "doc",
-        id: "motoko/intro/index",
+        id: "motoko/main/motoko",
       },
       items: [
         {
--- a/sidebars.js
+++ b/sidebars.js
@@ -172,7 +172,7 @@ const sidebars = {
       },
       items: [
         "motoko/version",
-        "motoko/main/motoko-introduction",
+        "motoko/main/motoko",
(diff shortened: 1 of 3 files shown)


sidebars.js
```diff
@@ -172,7 +172,6 @@ const sidebars = {
       },
       items: [
         "motoko/version",
```
crusso, line 174:
```suggestion
        "motoko/version",
        "motoko/main/motoko",
```
I wonder if it makes sense to also have a duplicate link here, just as an intro to the following docs. I'll leave it up to you. You can just commit the suggestion if you want to change, or leave it if not.

crusso (commented):
LGTM. See minor suggestion you can take or leave.

BTW Develop -> Motoko Docs from the "World Computer is our future" initial page still takes you the first tutorial page by Mikhail rather than the landing page. Not sure if that is intended or not. I can live with it.

("World Computer is our future" makes me want to hurl, btw. Is it just me?)

sidebars.js
dfx-json: I agree in principal, but doing so causes two items to be highlighted at the same time:
Screenshot 2023-03-07 at 12 22 04 PM

crusso (approved):
Thanks for fixing those links!

olaszakos (approved):
LGTM

PR #1184, opened 2023-03-09
update contribute card

domwoe:
- Make card for Office hours more useful and change invalid link.
- Link to the DFINTY Grants program needs to be there. However, it's also a pity to remove the ICDevs Bounty card.


sesi200 (changes requested):
I don't see the change. Could it be that your changes are in the wrong file? If you change `DevelopersHome` to `DocsHome` in the path you should still see the old text

sesi200 (approved):

PR #1220, opened 2023-03-23
docs: clarify `AuthClient` documentation

rvanasa:
Adjusts an example involving `@dfinity/auth-client` to better match what I was hoping to find (from the perspective of a JavaScript developer looking for a copy-paste example) on this documentation page.


dfx-json (approved):
Nice 👍

docs/developer-docs/integrations/internet-identity/integrate-identity.md
```diff
@@ -37,18 +37,14 @@ DFINITY provides an [easy-to-use library (agent-js)](https://github.com/dfinity/
...
 // Call authClient.login(...) to login with Internet Identity. This will open a new tab
 // with the login prompt. The code has to wait for the login process to complete.
-// We can either use the callback functions directly or wrap in a promise.
-await new Promise((resolve, reject) => {
-  authClient.login({
-    onSuccess: resolve,
-    onError: reject,
-  });
-});
+await authClient.login();
```
krpeacock, line 47: This won't actually work, even though login returns a promise 😅

The promise wrapper will actually work, but even then, I recommend that apps pass `onSuccess` and `onError` handlers, and avoid `await`ing `login`

krpeacock (changes requested):

docs/developer-docs/integrations/internet-identity/integrate-identity.md
rvanasa: Okay I see what you're saying. Would it be feasible for this to work without a promise wrapper in v1.0? A precedent would be Auth0's SPA login flow (specifically [`loginWithRedirect()`](https://auth0.com/docs/libraries/auth0-single-page-app-sdk#use-promises)), which works exactly like this.

rvanasa: Taking another look, it seems like the current promise-wrapping example causes a type error in TS (at least for some tsconfigs) caused by a mismatch between the input values for `onSuccess()` and `resolve(value)`. Just wanted to mention this since it's relevant to this conversation, although I'm sure most people will be able to figure out how to fix this.

jessiemongeon1:
This PR appears to be outdated/not needed. Closing this PR to clean up the repo. Please feel free to reopen if you think this should be revisited.

PR #1259, opened 2023-04-01
fix: missing Rust code syntax highlighting

peterpeterparker:
### Fix

Rust code syntax is not higlighted because Docusaurus does not enable it per default.

See [Docusaurus documentation](https://docusaurus.io/docs/markdown-features/code-blocks#supported-languages).


sesi200 (approved):
Looks great! Thank you very much
image

PR #1260, opened 2023-04-01
Update trust-in-canisters.md

0kenx:
Update in accordance with NNS proposal #115067


sesi200:
Thanks for the suggestion! Closing this in favour of #1280, which is a bit more precise and extensive.

Petty nitpicking: `Following the adoption of NNS proposal` is not entirely correct - it has always been like that

PR #1289, opened 2023-04-04
update the trust in canisters page

nathanosdev:


docs/concepts/trust-in-canisters.md
```diff
@@ -26,25 +26,23 @@ Canister controllers, if not voluntarily decentralized, have complete control ov
...
 
-### Governed Mutability
+### Sole Control by NNS and Other Governance
 
 A more complex but powerful approach is to control the canister via a distributed governance mechanism. One can imagine different levels of complexity and control that such a governance mechanism may implement. An example is the (upcoming) [SNS feature](https://medium.com/dfinity/how-the-service-nervous-system-sns-will-bring-tokenized-governance-to-on-chain-dapps-b74fb8364a5c) which allows developers to set the controller of their canister to some governing canister.
```
andrew-lee-work, line 47:
```suggestion
A more complex but powerful approach is to set the controller of the canister to a distributed governance mechanism. In this case, the NNS still has ultimate control over the canister. The advantage is that the more specific governance mechanism can be more focused on the canister. One can imagine different levels of complexity and control that such a governance mechanism may implement. An example is the (upcoming) [SNS feature](https://medium.com/dfinity/how-the-service-nervous-system-sns-will-bring-tokenized-governance-to-on-chain-dapps-b74fb8364a5c) which allows developers to set the controller of their canister to some governing canister.
```

nathanosdev: Thanks!

sesi200 (approved):

PR #1311, opened 2023-04-11
icpcoins showcase addition

infu:
Icpcoins showcase

PR #1313, opened 2023-04-11
adding Bitshop community project

martinagrzybowska:
adding Bitshop community project as requested


sesi200:
Did you want to add this to `showcase.json` like e.g. #1311? I don't think the current change displays anywhere

martinagrzybowska:
> Did you want to add this to `showcase.json` like e.g. #1311? I don't think the current change displays anywhere

It seems to me that `showcase.json` is a different section, but I will add it there as well.

martinagrzybowska:
> Did you want to add this to `showcase.json` like e.g. #1311? I don't think the current change displays anywhere

added as requested

sesi200 (approved):

PR #1315, opened 2023-04-12
Metamob showcase

av1ctor:
Metamob showcase, as requested by Dfinity's grant checklist.


sesi200 (approved):
Glad to have you on board!

PR #1383, opened 2023-04-28
corrected single tick for a backtick

apophis51:
line 55 and 56 incorrectly used single quotations '   instead of backticks  `
   
I could not get the code to run as it was written. However, with those changes the code executes flawlessly


apophis51:
disreguard this pull request... it looks like the root cause is the space in export TOKEN_NAME="My Token" In the document https://internetcomputer.org/docs/current/developer-docs/integrations/icrc-1/deploy-new-token
I will need to submit a different pull request

PR #1418, opened 2023-05-08
Add network=ic to tutorial of top-up cycles

zire:
The original tutorial misses '--network ic' and would lead to an error message as it could only transfer cycles on the local machine


dprats (approved):

dprats:
Good catch. Thanks @zire !

PR #1449, opened 2023-05-13
feat: add subnav to dev docs

dfx-json:
This PR reverts the docs-specific navbar in favor of a dedicated subnav menu that appears below the ubiquitous nav. Logic has been added to ensure the subnav has behavior consistent with the nav bar with respect to autohide functionality. The menu also received treatment and computes the padding dynamically based on whether the navbar is visible or hidden. The result is a full navigation experience that feels seamless without having to eject the nav component.

Note - I've opted to disable the auto-hide functionality for navbar on mobile since toggling the menu via the breadcrumbs would require some pretty ugly hacks - the mobile menu is hidden within the nav element; if the navbar is hidden, we need to make it visible __somehow__ - this could be through a programmatic scroll or something else. Still ugly regardless.

https://github.com/dfinity/portal/assets/98767015/e1ef5f14-4ab9-4bda-bc3f-8c564982d35b


krpeacock:
image
Styling glitch on Desktop Chrome - small window sizes
Also verified on Firefox

The padding is also off prior to the initial scroll animation
image

src/components/Basics/Hero/index.module.css
```diff
@@ -28,7 +28,7 @@
   bottom: -680px;
   right: -750px;
   width: 1500px;
-  z-index: -1;
+  @apply -z-1;
```
krpeacock, line 31: Does this work cross-browser? https://caniuse.com/sr_css-apply

src/components/Common/DevDocsBreadcrumbs.tsx
```diff
@@ -0,0 +1,55 @@
...
+    } else {
+      return [parts[0]];
+    }
+  }, [location]);
+
+  const toggleMenu = useCallback((event: React.MouseEvent) => {
+    document
+      .getElementsByClassName("navbar__toggle")
+      .item(0)
+      ?.dispatchEvent(
```
krpeacock, line 34: A click event triggering a click event is always a little suspicious, but I'm sure this was needed. I'd add a comment here explaining the interaction

krpeacock (changes requested):

src/components/Basics/Hero/index.module.css
dfx-json: this is a [tailwind directive](https://tailwindcss.com/docs/functions-and-directives#apply)

src/components/Common/DevDocsBreadcrumbs.tsx
dfx-json: I agree completely, however, in this case, I'm not sure of another way to trigger the mobile menu to activate. Docusaurus doesn't expose the navbar's API through a global context or hook. It seems the only we can activate the mobile menu is to synthesize a click event on the `navbar__toggle` element. I'll continue to look for a cleaner way to tackle this.

olaszakos (approved):
Introducing tailwind directives in .css files doesn't help with maintenance, especially if the classes are custom which have a higher chance of getting changed or removed. I'd keep the CSS as is, or completely reimplement it with tailwind in the jsx files. Good to go nonetheless.

olaszakos:
Found an issue on desktop chrome, mac, the dropdowns are cut off: 

image

olaszakos:
Issue on desktop chrome, mac, if you go back to the main menu on the mobile nav, it's empty 
image

krpeacock (approved):

PR #1526, opened 2023-06-02
Add Node.js dependency to Poll tutorial

aubort:
In the Poll tutorial, there is no mention of Node.js being a prerequisite. If a user doesn't complete the "hello" tutorial first, they might not have node installed. In this case, frontend files will not be created and following the tuto is not possible.


jessiemongeon1 (approved):

aubort:
Thanks @jessiemongeon1 and @sesi200!

PR #1553, opened 2023-06-08
fix: wrong trillion cycles in table

Luca8991:
Fixed the count from 10 billion in the table


dprats (approved):

dprats:
Good catch @Luca8991 ! Thank you for catching this and submitting a fix.

PR #1593, opened 2023-06-17
add: dsign to showcase

cybrowl:


cybrowl:
Wondering if I can move the position of where it displays? Can I choose to be displayed next to a particular project like Codebase?

sesi200 (approved):
I don't think there is a way. AFAIK the order is random

PR #1620, opened 2023-06-26
Add two showcase from ICLighthouse

elephant-w:


sesi200:
The CLA bot broke once again... Would you mind going through the flow over here too? https://github.com/dfinity/cla/issues/279

showcase.json
```diff
@@ -1958,5 +1958,46 @@
...
+      "DeFi"
+    ],
+    "usesInternetIdentity": false,
+    "github": "",
+    "twitter": "",
+    "youtube": "",
+    "submittableId": "36483996"
+  },
+  {
+    "name": "BTC integration in ICDex",
```
sesi200, line 1983: ICDex is already in the showcase and this tile seems somewhat empty. In this state I don't think it should be added

elephant-w: Hello, what we added was not the ICDex project, but projects related to BTC and ETH integration. These projects are primarily intended for grant applications.

jennifertrin: Hello @elephant-w We allow only one listing per core product on the showcase. Would you be open to highlighting that you offer a BTC and ETH integration on the ICLighthouse listing instead?

jennifertrin:
Hello @elephant-w As mentioned above, we allow only one listing per core product on the showcase. If you would like to highlight that you offer a BTC and ETH integration, please update your description in a new PR.

PR #1635, opened 2023-06-30
PSEC-1360: Add security best practices for inspect_message

tmu0:


docs/developer-docs/security/rust-canister-development-security-best-practices.md
```diff
@@ -581,6 +581,16 @@ Canisters pay for their cycles which makes them inherently vulnerable to attacks
...
 
+### Do not rely on ingress message inspection
+
+#### Security concern
+
+The correct execution of [canister_inspect_message](../../references/ic-interface-spec#system-api-inspect-message) is not guaranteed in the presence of a malicious node, because it is executed as a [query call](../../references/ic-interface-spec#http-query).  
+
+#### Recommendation
+
+Your canisters should not rely on the correct execution of `canister_inspect_message`. This in particular means that no security relevant checks, such as authentication or authorization, should be solely performed in that method. Such checks **must** be performed as part of an update method to guarantee reliable execution. Also note that for inter-canister calls `canister_inspect_message` is not invoked, so these calls can't be handled there.
```
roelstorms, line 592:
```suggestion
Your canisters should not rely on the correct execution of `canister_inspect_message`. This in particular means that no security critical code, such as access control checks, should be solely performed in that method. Such checks **must** be performed as part of an update method to guarantee reliable execution. Ideally, they are executed both in the `canister_inspect_message` function and a guard function. 

Also note that for inter-canister calls `canister_inspect_message` is not invoked which is another reason to execute the code as part of the update call by using a guard.
```

docs/developer-docs/security/rust-canister-development-security-best-practices.md
```diff
@@ -581,6 +581,18 @@ Canisters pay for their cycles which makes them inherently vulnerable to attacks
 
 Consider monitoring, early authentication, rate limiting on canister level to mitigate this. Also, be aware that an attacker will aim for the call consuming most cycles. See the "Cycle balance drain attacks section" in [how to audit an Internet Computer canister](https://www.joachim-breitner.de/blog/788-How_to_audit_an_Internet_Computer_canister).
 
+### Do not rely on ingress message inspection
```
robin-kunzler, line 584: Nice, thanks for adding this @tmu0 ! LGTM. 

A minor point: Would it make sense to link this section from the section "Make sure any action that only a specific user should be able to do requires authentication" and vice versa?

robin-kunzler (approved):
Approving this as I only left one minor comment :-) Thanks again @tmu0

PR #1647, opened 2023-07-05
WebSockets Blog Post

r-birkner:


blog/features/websockets-poc.md
harpa12, line 1: Great post! Just a few things: 

- I committed a few minor grammatical/ stylistic changes already.
- Would it makes sense to add a line about how the proof of concept differs from HTTPS outcalls on the IC? In terms of functionality, one is bidirectional and the other not, but how does the architecture differ and will the use cases be the same? Perhaps this is more something for the Medium blog??
- What about adding a link to the Forum discussion at the end??

r-birkner: Thanks a lot for your edits and comments. Sorry it took me some time to get back to it.
* I have added the link to the forum discussion (I just forgot).
* I tried to add two sentences about HTTPS outcalls: In the beginning, I mention that WebSockets are the continuation of the outcalls and in the architecture section, I mention that we could have used HTTPS outcalls as an alternative for polling.
I hope that's along the lines of what you had in mind.

harpa12 (approved):

PR #1693, opened 2023-07-17
Update rust-canister-development-security-best-practices.md

levifeldman:
Clarify that a response by a callee-canister is guaranteed to come back to the caller-canister.


alin-at-dfinity:
I guess the accurate thing to say here is that a request only ever results in a single response. That response may be:
* a system-generated reject, before the call reaches the callee;
* a response from the callee; or
* a system-generated reject after the call has reached the callee; and might have resulted in changes to the state of the callee.

This unique response is then guaranteed to be delivered to the caller.

So I believe that the existing description is accurate, modulo the "generally guarantees" bit. A reject response guarantees that the call has failed at some point. That point might have been before anything got executed or at least persisted; or it may be after some changes (possibly all changes, depending on the callee's implementation) have been persisted. So "generally guarantees" is technically correct, but I would rather have it say that "a reject response guarantees that the call has not been processed successfully by the called canister". Note the "guarantees" and "call" as opposed to "message": the request message may have been processed successfully and the trap (or system error) may have occurred while processing a different message, e.g. a response to a downstream call.

levifeldman:
> @levifeldman `Such system-generated rejects can occur at any time before the call reaches the callee-canister.` is not accurate. They can also occur if the request has reached the callee-canister and the canister traps during processing.

@tmu0 If the canister traps during processing, I consider that as coming from the called canister and not 'system-generated' since a canister trap during processing can only happen because of something the canister is doing wrong, as oppose to system-generated rejects which can happen before the call reaches the callee even if the callee is doing everything correct.

levifeldman:
@tmu0
> the message that should come across is that errors can happen even after the message has arrived at the receiver 

I added this point into the PR. Let me know if it is good or needs more clarification.

> and that a reject response means that the message hasn't been successfully processed by the receiver but doesn't guarantee that the receivers state wasn't changed

 I took out that last line in the first commit because of the ambiguity around the words 'generally guarantees' and 'successfully' because of the fact that the receiver can still commit state while returning a manual reject, so now with this bit at the end I put it in.

PR #1708, opened 2023-07-20
Adding dfx deps documentation page

jessiemongeon1:


docs/developer-docs/setup/pulling-canister-dependencies.md
```diff
@@ -0,0 +1,430 @@
...
+        "wasm_url": "https://github.com/lwshang/pullable/releases/latest/download/service.wasm",
+        "init_guide": "A natural number, e.g. 1"
+      }
+    }
+  }
+}
+```
+
+:::caution
+The wasm module of a `pullable` canister must be hosted via an URL so that service consumers can download it when pulling the dependency.
```
dfx-json, line 60: I wonder if we should also add a footnote indicating that a future evolution of this feature will likely leverage direct wasm downloads from the replica.

docs/developer-docs/setup/pulling-canister-dependencies.md
```diff
@@ -0,0 +1,430 @@
...
+    }
+}
+```
+
+The `pullable` object will be serialized as a part of the `dfx` metadata and attached to the wasm.
+
+To better understand the `pullable` object, let's look at each property in depth. 
+
+- `wasm_url`: A URL used to download the canister wasm module which will be deployed locally.
+- `wasm_hash` A SHA256 hash of the wasm module located at `wasm_url`. This field is optional. In most cases, the wasm module at `wasm_url` will be the same as the on-chain wasm module. This means that dfx can read the state tree to obtain and verify the module hash. In other cases, the wasm module at `wasm_url` is not the same as the on-chain wasm module. For example, the Internet Identity canister provides a `development` flavor to be integrated locally. In these cases, `wasm_hash` provides the expected hash, and `dfx` verifies the downloaded wasm against this.
```
dfx-json, line 96:
```suggestion
- `wasm_hash` A SHA256 hash of the wasm module located at `wasm_url`. This field is optional. In most cases, the wasm module at `wasm_url` will be the same as the on-chain wasm module. This means that dfx can read the state tree to obtain and verify the module hash. In some cases, the wasm module at `wasm_url` is not the same as the on-chain wasm module. For example, the Internet Identity canister provides a `development` variant to be integrated locally. In these cases, `wasm_hash` provides the expected hash, and `dfx` verifies the downloaded wasm against this.
```

I also think we should expand on the implication of not receiving the expected hash. Will dfx print a warning? Should developers be concerned? When should they ignore the warning, and in what circumstances should they heed the warning?

docs/developer-docs/setup/pulling-canister-dependencies.md
```diff
@@ -0,0 +1,430 @@
...
+The canister wasm downloaded from `wasm_url` should have the following metadata (public or private):
+
+- `candid:service`
+- `candid:args`
+- `dfx`
+
+All metadata sections are handled by `dfx` when the canister is built.
+
+### Deployment process
+Service providers will to use the following deployment process to deploy their `pullable` canister.
```
dfx-json, line 113:
```suggestion
Service providers will use the following deployment process to deploy their `pullable` canister.
```

docs/developer-docs/setup/pulling-canister-dependencies.md
```diff
@@ -0,0 +1,430 @@
...
+```
+dfx start --clean --background
+dfx deps deploy
+dfx deploy app
+```
+
+- #### Step 7: Test the app by making the following call:
+  
+```
+dfx canister call app double_service
```
dfx-json, line 396: a little confusing here. why call `app` and not `service`? Does `app` in turn call `service`? Can we explain what is happening here?

docs/developer-docs/setup/pulling-canister-dependencies.md
jessiemongeon1: Updated this bullet point with information from @lwshang via Slack DM conversation.

docs/developer-docs/setup/pulling-canister-dependencies.md
jessiemongeon1: Updated this step with clarification from @lwshang via Slack DM conversation.

docs/developer-docs/setup/pulling-canister-dependencies.md
```diff
@@ -0,0 +1,441 @@
...
+In this example, the `app` canister defines a method called `double_service` which makes an inter-canister call to the `service` canister.
+
+- #### Step 1: First, assure that you have installed the [IC SDK](https://github.com/dfinity/sdk) and [git](https://git-scm.com/downloads).
+
+- #### Step 2: Then, open a terminal window and use `git clone` to download the following repository:
+
+```
+git clone https://github.com/lwshang/pullable.git
+```
+
```
lwshang, line 362: Clone the service provider's repo is not a required step.

In this demo, a service consumer should `dfx new` to get new project and edit the `dfx.json` as in step 3 below.

dfx-json (approved):
👌

PR #1724, opened 2023-07-24
Adding VetKD encrypted notes and QR code dapp examples to sample projects page

jessiemongeon1:


docs/samples/vetkd-encrypted-notes.md
```diff
@@ -0,0 +1,137 @@
+# VetKD encrypted notes dapp
+
+## Overview
+
+This example dapp is a new version of the [encrypted notes dapp](./encrypted-notes.md) that has been altered to use the new [proposed vetKD feature](https://github.com/dfinity/interface-spec/pull/158).
+
+Particularly, the dapp has been altered where instead of creating a principal-specific AES key and syncing it across devices (by means of device-specific RSA keys), the notes are encrypted with an AES key that is derived (directly in the browser) from a principal-specific vetKey obtained from the backend canister. THis key is obtained in an encrypted form, using an ephemeral transport key, which obtains itself from the vetKD system API. This way, there is no need for any device management in the dapp.
```
dprats, line 7:
```suggestion
Particularly, the dapp has been altered where instead of creating a principal-specific AES key and syncing it across devices (by means of device-specific RSA keys), the notes are encrypted with an AES key that is derived (directly in the browser) from a principal-specific vetKey obtained from the backend canister. This key is obtained in an encrypted form, using an ephemeral transport key, which obtains itself from the vetKD system API. This way, there is no need for any device management in the dapp.
```

dprats (approved):

PR #1743, opened 2023-07-28
Initial version of dev documents

Dfinity-skaestle:


docs/developer-docs/integrations/composite-query/composite-query.md
```diff
@@ -0,0 +1,135 @@
...
+```
+
+:::note
+The first call to put might be slow to respond because the data partition canisters have to be created first.
+:::
+
+The output should resemble the following indicating that no value has previously been registered for this key:
+```(null)```
+
+Step 2: Retrieve the value associated with a key using composite queries with the command:
```
jessiemongeon1, line 101:
```suggestion
### Step 2: Retrieve the value associated with a key using composite queries with the command:
```

docs/developer-docs/integrations/composite-query/composite-query.md
```diff
@@ -0,0 +1,135 @@
...
+dfx canister call kv_frontend get '(1)'
+```
+
+The output should resemble the following:
+```(opt (1337 : nat))```
+
+This workflow displays the ability to fetch the value using composite queries with very low latency.
+
+### Comparing composite queries to calls from update functions
+Let’s now compare the performance of composite query calls with those of an equivalent implementation that leverages calls from update functions. To do this, we will use the get_update method, which contains the exact same logic, but is implemented based on update calls. Run the following command in your terminal window:
```
jessiemongeon1, line 113:
```suggestion
Let’s now compare the performance of composite query calls with those of an equivalent implementation that leverages calls from update functions. To do this, we will use the `get_update` method, which contains the exact same logic, but is implemented based on update calls. Run the following command in your terminal window:
```

jessiemongeon1 (approved):

PR #1784, opened 2023-08-11
Update local-development.md

letmejustputthishere:

### DIFF
--- a/docs/developer-docs/integrations/bitcoin/local-development.md
+++ b/docs/developer-docs/integrations/bitcoin/local-development.md
@@ -105,6 +105,10 @@ The port used can be changed in `dfx.json`.
 :::
 
 - #### Step 2: Deploy the example canister:
+    If you're running the Rust example, use
+
+        dfx deploy basic_bitcoin --argument '(variant { regtest })'
+    If you're running Motoko, use
 
         dfx deploy basic_bitcoin --argument '(variant { regtest })'
 


docs/developer-docs/integrations/bitcoin/local-development.md
```diff
@@ -105,6 +105,10 @@ The port used can be changed in `dfx.json`.
 :::
 
 - #### Step 2: Deploy the example canister:
+    If you're running the Rust example, use
```
jessiemongeon1, line 108:
```suggestion
    If you're running the Rust example, use the command:
```

docs/developer-docs/integrations/bitcoin/local-development.md
```diff
@@ -105,6 +105,10 @@ The port used can be changed in `dfx.json`.
...
+
+        dfx deploy basic_bitcoin --argument '(variant { regtest })'
+    If you're running Motoko, use
```
jessiemongeon1, line 111:
```suggestion

    If you're running the Motoko example, use the command:
```

jessiemongeon1 (approved):

ielashi (changes requested):
We don't need separate commands for the different languages. We just need to fix the Motoko examples's dfx.json. I'll take care of it.

letmejustputthishere:
> We don't need separate commands for the different languages. We just need to fix the Motoko examples's dfx.json. I'll take care of it.

@ielashi the initialisation arguments are still different for the two examples so i don't think there's anything you can do about that in `dfx.json`?

motoko expects [upper case variants](https://github.com/dfinity/examples/blob/d54cd27712450e77e4a398e35f0f0118eee8a47e/motoko/basic_bitcoin/src/basic_bitcoin/basic_bitcoin.did#L9) and rust expects [lower case variants](https://github.com/dfinity/examples/blob/d54cd27712450e77e4a398e35f0f0118eee8a47e/rust/basic_bitcoin/src/basic_bitcoin/basic_bitcoin.did#L9) for the init args.

so imo we still need to merge this PR

ielashi:
> > We don't need separate commands for the different languages. We just need to fix the Motoko examples's dfx.json. I'll take care of it.
> 
> @ielashi the initialisation arguments are still different for the two examples so i don't think there's anything you can do about that in `dfx.json`?
> 
> motoko expects [upper case variants](https://github.com/dfinity/examples/blob/d54cd27712450e77e4a398e35f0f0118eee8a47e/motoko/basic_bitcoin/src/basic_bitcoin/basic_bitcoin.did#L9) and rust expects [lower case variants](https://github.com/dfinity/examples/blob/d54cd27712450e77e4a398e35f0f0118eee8a47e/rust/basic_bitcoin/src/basic_bitcoin/basic_bitcoin.did#L9) for the init args.
> 
> so imo we still need to merge this PR

Both examples should use lower-case invariants, as that what is specified in the interface specification of the Bitcoin API. I addressed this, along with updating dfx.json here: https://github.com/dfinity/examples/pull/606, so we don't the change in this PR.

PR #1802, opened 2023-08-15
Roadmap entry: ckETH and ckERC-20 tokens

dietersommer:


jessiemongeon1 (approved):

manudrijvers:
Our plan is that we have ckETH in 2023 but we will do ckERC20 after that, and there I think 2023 for ckERC20 is optimistic. Maybe we can split it into two entries, where we put 2023 on ckETH, but something later on ckERC20?

dietersommer:
> Our plan is that we have ckETH in 2023 but we will do ckERC20 after that, and there I think 2023 for ckERC20 is optimistic. Maybe we can split it into two entries, where we put 2023 on ckETH, but something later on ckERC20?

Just split it into the features on the roadmap and adjusted the timeline to 2024 for ckERC-20.

PR #1806, opened 2023-08-16
Extend recommendation to protect against cycles draining attacks in t…

robin-kunzler:
Extends the recommendations to protect against cycles draining attacks in the security best practices.


docs/developer-docs/security/rust-canister-development-security-best-practices.md
```diff
@@ -581,7 +581,11 @@ Canisters pay for their cycles which makes them inherently vulnerable to attacks
...
+* Consider monitoring, early authentication, rate limiting on canister level to mitigate this. Also, be aware that an attacker will aim for the call consuming most cycles. See the "Cycle balance drain attacks section" in [how to audit an Internet Computer canister](https://www.joachim-breitner.de/blog/788-How_to_audit_an_Internet_Computer_canister).
+
+* For query calls that cause significant computation and don't modify the state, it is advisable to not execute the expensive computation if the method is called as update. Unfortunately, the execution mode of the query (whether it was called as query or update) is currently not directly exposed to the user code. However, one can e.g. call  `ic0.data_certificate_present()` which returns `1` when called as query, and `0` for update methods. See the Interface Specification [section on certified data](/references/ic-interface-spec.md#system-api-certified-data).
```
domwoe, line 586: OTOH this opens up the security issue for the client to get data from a single malicious node. You also (typically?) can't use certified data to mitigate in the case of a computationally heavy query.

Probably still good to mention it here though.

robin-kunzler: Good point, thanks. I mentioned this trade-off and linked to the query certification best practice.

robin-kunzler:
Thanks for the review and approval @tmu0 !

> Regarding the 3rd point one could mention that it is also possible to require payment for ingress messages. It is just not supported by the platform and has to be custom designed.

Good point, makes sense. I added an additional bullet describing this. It is very vague but I guess that's what it means that it would need to be designed ;-)

domwoe (approved):

domwoe:
I approved, but I'd like to know how this would work in practice @tmu0?  The only way I see would be a pre-payment since otherwise, you'd always have to make an inter-canister call to check for a payment.

robin-kunzler:
Thanks for the approval and merging @domwoe !

PR #1917, opened 2023-09-19
Sns testing 2

dprats:
Updates to the testing pages for SNS

Screenshot 2023-09-19 at 1 46 59 PM


docs/developer-docs/integrations/sns/launching/launch-steps-1proposal.md
```diff
@@ -28,21 +28,22 @@ ls bazel-bin/rs/sns/cli/sns
 ```
 ## Stages
 
-### 1. Dapp developers choose the initial parameters of the SNS for a dapp.
```
LaraAS, line 31: I think we deliberatley wanted to have the choosing of the parameters as a first step in the launch as this is actually what might take multiple weeks in reality, so I would not recommend to already add NNS as controller before starting this process. Is there any reason why you think it would be important to change this order?

docs/developer-docs/integrations/sns/testing/testing-locally.md
```diff
@@ -5,13 +5,14 @@ sidebar_position: 2
 
 ## Overview
 
-To help developers, DFINITY has created the `sns-testing` repo which has scripts that help developers test the SNS process. Developers can run a local version of the Internet Computer on their local machine, deploy their dapp locally and run through [the stages](../launching/launch-summary.md) of decentralizing their dapp. 
+To help developers, DFINITY has created the [sns-testing repo](https://github.com/dfinity/sns-testing) which has scripts that help developers test the SNS process. Developers can run a local version of the Internet Computer on their local machine, deploy their dapp locally and run through [the stages](../launching/launch-summary.md) of decentralizing their dapp.
```
LaraAS, line 8: Should we already say here that this might not work for any dapp?
The repo is just supposed to show this on one example dapp and potentially you can do something similar with your own dapp, but as soon as you need something more customized it would probably be good to following the new guildelines here

docs/developer-docs/integrations/sns/testing/testing-locally.md
```diff
@@ -28,91 +29,119 @@ These scripts have been most tested with a stack that looks like this:
...
 
-## Testing via `sns-testing` repo
-To use the `sns-testing` commands and their arguments properly, please consult the `sns-testing` [README](https://github.com/dfinity/sns-testing#sns-lifecycle).
+## Getting started
+
+To use the `sns-testing` commands and their arguments properly, you need to first set up the `sns-testing` repo properly. Please see [README](https://github.com/dfinity/sns-testing#sns-lifecycle).
 
 :::info
-If you have started using `sns-testing` before August 2023 and are using the old legacy flow of launching an SNS (please see [here](../launching/index.md) for more context), please continue your work with the legacy documentation on [this old README](https://github.com/dfinity/sns-testing/blob/v1-legacy/README.md). In particular, 
+* All of the commands in this article require `sns-testing` to be set up properly.
```
LaraAS, line 37: Is it really the case that the commands here will basically be run "on top of `sns-testing`"? 
I thought here we would just describe the commands in terms of dfx commands (not relying on sns-testing as a tool) 
I assume that some of the setup that was done in sns-testing is also needed for this to work though but I am not quite sure which part.

LaraAS (commented):
Thanks a lot Diego, that already looks great! 
I am wondering if we need more clarification of how to use sns-tesing and this together. Let me ask if one of the NNS engineers can maybe chime in here.

docs/developer-docs/integrations/sns/testing/testing-locally.md
dprats: done

docs/developer-docs/integrations/sns/testing/testing-locally.md
dprats: yes,

1. you need to follow the install README of `sns-testing`. 

2. `dfx` commands are not enough, there is quite a bit of boiler plate within the bash scripts.

jessiemongeon1 (approved):

aterga:
First of all, having these docs is infinitely better than not having any docs, and this MR is IMHO a step forward. 

Thanks for the work, @dprats !

However, there's IMHO a conceptual problem with the approach that we've seem to be taking so far. Here's my understanding of what the relationship should be between sns-testing and the docs.

1. We shouldn’t say that sns-testing will cover all the SNS testing needs of your custom Dapp. We cannot guarantee this, and also people looking into the repo might get a false impression that all those scripts are supposed to always work regardless of their setup. This is not the case, as it's extremely hard to get a set of Bash scripts to always work for everyone (unless we define the exact scenarios and run them on CI).
2. The way I would rather position sns-testing: it’s a demo of how the SDK tooling can be used to solve the problem for a few typical (simplified) scenarios. 
3. The documentation should first and foremost explain how to solve the problem using the official SDK, not just linking to sns-testing.
4. If we can use the steps of the sns-testing's README file to illustrating the SDK-based instructions (which would be less concrete / more general than the scripts), that would be a great (but secondary) milestone.

So, in my mental model, the ideal documentation should be 

| Description | How to do that via the SDK tools | Concrete example from sns-testing (if available) |
|---|---|---|
| Call SNS Root's `list_sns_canisters` endpoint to get IDs of all canisters belonging to this SNS. | `dfx canister --network <network> call <sns-root-canister-id> [--candid <sns-root-canister-did>] list_sns_canisters '(record {})'` | [`./get_sns_canisters.sh`](https://github.com/dfinity/sns-testing/blob/main/get_sns_canisters.sh) |
| ... | ... | ... |


Wdyt, @dprats, @LaraAS  ?

LaraAS:
@aterga I agree with what you said and also that the last column in your table is less of a priority. 
We could show this as a table or, more similarly to launch commands, just as paragraphs (this would allow for indexing / referring to the individual steps). 
But no super strong preference here.

dprats:
> The documentation should first and foremost explain how to solve the problem using the official SDK, not just linking to sns-testing.

> If we can use the steps of the sns-testing's README file to illustrating the SDK-based instructions (which would be less concrete / more general than the scripts), that would be a great (but secondary) milestone.

@aterga does the current implementation NOT already do that for testing.

What I see is:

1. Step 1: has a bash script to deploy a dummy dapp
2. All other steps are standard `dfx` commands
3. The last step runs a bash script to simulate users participating in a swap

Screenshot 2023-09-28 at 10 40 15 AM

aterga:
> does the current implementation NOT already do that for testing.

I mainly wanted to agree on what we should aim to converge to, not to claim that the current thing is wrong.

> The first step has a bash script to deploy a dummy dapp

Yes, this one we cannot know upfront. But even then, maybe we could refer to some page explaining how to deploy dapps via DFX. This is IMHO a low priority thing, as the the devs preparing for an SNS will already likely be familiar with how to deploy non-decentralized dapps.

> The last step runs a bash script to simulate users participating in a swap

Note that for some use cases, that script won't work (depending on what DFX identities the user already has). Generally, the scripts make some implicit assumptions that work for sns-testing but might not work for an arbitrary community dev.

So ideally, we should also unpack that into something like:

1. Create a test identity via DFX
2. Form an SNS participation query via Quill
3. Send the signed query via Quill

I think it's also fine to merge this PR and keep refining the docs iteratively, so I approve the PR (but let's wait till tomorrow before we merge, just in case @LaraAS would have something else to add).

In a future iteration, it would also be good to explain how to validate the steps using the SDK tools. In principle, this also concerns the steps that are marked (Automated), as it's important to know where to look if something didn't go as planned.

aterga (approved):

PR #1935, opened 2023-09-23
Adding API examples when methods are mentioned

jessiemongeon1:
This PR adds API examples when an API method is mentioned to give additional context about how an API method can be used. This edit is in response to recent feedback given regarding the Dev Docs. 


dfx-json (approved):
Nice one Jessie

PR #2099, opened 2023-10-31
Revert capitalization change for Juno project

jessiemongeon1:

### DIFF
--- a/showcase.json
+++ b/showcase.json
@@ -542,7 +542,7 @@
   {
     "id": "juno",
     "name": "Juno",
-    "oneLiner": "Build Web3 dapps like Web2",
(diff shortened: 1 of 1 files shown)


showcase.json
```diff
@@ -542,7 +542,7 @@
   {
     "id": "juno",
     "name": "Juno",
-    "oneLiner": "Build Web3 dapps like Web2",
+    "oneLiner": "Build Web3 dApps like Web2",
```
dfx-json, line 545: was there another PR recently where all instances of `dApp` were normalized to `dapp`?

jessiemongeon1: There was, but David has requested that this single instance be reverted back to `dApp` to reflect the Juno branding properly.

PR #2113, opened 2023-11-02
Replace system and platform

yvonneanne:


robin-kunzler (approved):
Thanks for improving the wording @yvonneanne ! LGTM

PR #2160, opened 2023-11-07
Add redirects for smartcontract.org embedded URLs 

jessiemongeon1:

### DIFF
--- a/plugins/utils/redirects.js
+++ b/plugins/utils/redirects.js
@@ -177,6 +177,8 @@ const redirects = `
   /docs/current/developer-docs/production/instruction-limits /docs/current/developer-docs/production/resource-limits
   /docs/current/developer-docs/backend/resource-limits /docs/current/developer-docs/production/resource-limits
   /docs/current/concepts/data-centers /docs/current/concepts/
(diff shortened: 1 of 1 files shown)


plugins/utils/redirects.js
```diff
@@ -177,6 +177,8 @@ const redirects = `
   /docs/current/developer-docs/production/instruction-limits /docs/current/developer-docs/production/resource-limits
   /docs/current/developer-docs/backend/resource-limits /docs/current/developer-docs/production/resource-limits
   /docs/current/concepts/data-centers /docs/current/concepts/
+  /docs/current/developer-docs/build/languages/rust/rust-intro/ /docs/current/developer-docs/backend/rust/
```
dfx-json, line 180: I think we want to omit the trailing slash, otherwise it will error out.

dfx-json: (sorry for giving you an imprecise mapping earlier)

dfx-json (approved):

plugins/utils/redirects.js
jessiemongeon1: No worries, thanks for catching it! 😃

PR #2272, opened 2023-12-11
chore: update motoko doc to 0.10.2, matching dfx 0.15.2

crusso:


crusso:
I checked out the preview and it looks ok to  me. The Regions docs are there, compiler version is 0.10.2 and the embedded interpreter seems to work fine.

PR #2297, opened 2023-12-18
Overhaul Gas & Cycles cost page

jessiemongeon1:
Thank you for your contribution to the IC Developer Portal.
Before submitting your Pull Request, please make sure that:

- [ ] You have updated the [`.github/CODEOWNERS`](https://github.com/dfinity/portal/blob/master/.github/CODEOWNERS) file according to your changes.
- [ ] You have registered new documents to [`/sidebars.js`](https://github.com/dfinity/portal/blob/master/sidebars.js) if any.

PR #2353, opened 2024-01-17
Additional Revisions: Gas cost page

jessiemongeon1:

### DIFF
--- a/docs/developer-docs/gas-cost.md
+++ b/docs/developer-docs/gas-cost.md
@@ -24,9 +24,9 @@ Canisters pay for consumed resources by burning cycles. These resources include 
 
 ## Units and fiat value
 
-The price of cycles is fixed against the price of [XDR](/docs/current/references/glossary#xdr), where **1 trillion cycles equals 1 XDR**. As of December 18, 2023, the exchange rate for 1 XDR = $1.336610, which is used on this page. The exchange rate for USD/XDR may vary, and it will impact the conversion rate. You can view XDR exchange rates [here](https://www.imf.org/external/np/fin/data/rms_sdrv.aspx).
+The price of cycles is fixed against the price of [XDR](/docs/current/references/glossary#xdr), where **1 trillion cycles equals 1 XDR**. As of December 18, 2023, the exchange rate for 1 XDR = $1.336610. The exchange rate for USD/XDR may vary. Learn more about [XDR exchange rates](https://www.imf.org/external/np/fin/data/rms_sdrv.aspx).
 
-This documentation will use the following units to measure and calculate the number of cycles and their associated fiat value:
+This documentation will use the following units:
 
 | Abbreviation  | Name      | In numbers   | Cycles XDR value  | Cycles USD value |
 |-------------- | --------- | ------------ | ----------------- | ---------------- |
@@ -38,26 +38,28 @@ This documentation will use the following units to measure and calculate the num
 
 ## The ICP reverse gas model
 
-In contrast to other blockchains where users pay for their transactions, ICP charges the smart contract canister for the resources it consumes. This allows developers to provide a smoother user experience, as users are free from tedious tasks such as signing and approving every transaction they perform within the application. This cost model is known as ICP's 'reverse gas model'.
+ICP charges the smart contract canister for the resources it consumes. This allows developers to provide a smoother user experience, as users are free from tedious tasks such as signing and approving every transaction they perform. This cost model is known as ICP's 'reverse gas model'.
 
 :::info
-While this may be referred to as the ICP reverse 'gas' model, it is important to note that ICP does not use 'gas' -- it uses cycles.
+It is important to note that ICP does not use 'gas' -- it uses cycles.
 :::
 
-You can learn more about the ICP reverse gas model [here](https://internetcomputer.org/capabilities/reverse-gas).
+You can learn more about the [ICP reverse gas model](https://internetcomputer.org/capabilities/reverse-gas).
 
-One downside of the ICP reverse gas model in comparison to other models, such as the EVM gas model, is that it requires prerequisite steps and ongoing maintenance for developers. Canisters must have their cycles balances maintained and regularly topped up as they continuously use resources; otherwise, the canister will be removed from the network if it runs out of cycles. However, a [freezing threshold](/docs/current/tutorials/developer-journey/level-1/1.6-managing-canisters#setting-the-canisters-freezing-threshold) can be set that pauses a canister's executions if the cycles amount is expected to fall below a certain amount, and there are several community tools that have been developed to automate managing a canister's cycles, such as [CycleOps](https://cycleops.dev/).
+One downside of the ICP reverse gas model is that it requires prerequisite steps and ongoing maintenance for developers. Canisters must have their cycles balances maintained and regularly topped up as they continuously use resources. The canister will be removed from the network if it runs out of cycles. However, a [freezing threshold](/docs/current/tutorials/developer-journey/level-1/1.6-managing-canisters#setting-the-canisters-freezing-threshold) can be set that pauses a canister's executions if the cycles amount is expected to fall below a certain amount. There are several community tools that have been developed to automate managing a canister's cycles, such as [CycleOps](https://cycleops.dev/).
 
-You can learn how to query a canister's cycles balance [here](/docs/current/tutorials/developer-journey/level-1/1.6-managing-canisters#checking-the-cycles-balance-of-a-canister), and you can learn how to top up a canister [here](/docs/current/tutorials/developer-journey/level-1/1.6-managing-canisters#topping-up-a-canisters-cycles-balance).
+Learn how to query a [canister's cycles balance](/docs/current/tutorials/developer-journey/level-1/1.6-managing-canisters#checking-the-cycles-balance-of-a-canister). 
+
+Learn how to [top up a canister](/docs/current/tutorials/developer-journey/level-1/1.6-managing-canisters#topping-up-a-canisters-cycles-balance).
 
 
 ## Canister operations and resources
 
-Canisters pay for consumed resources as well as for performed operations using features such as HTTPS outcalls, ECDSA signing, and the Bitcoin integration API. At a high level this can be visualized using the following diagram:
+Canisters pay for consumed resources and performed operations using features such as HTTPS outcalls, ECDSA signing, and the Bitcoin integration API. At a high level this can be visualized using the following diagram:
 
 ![Canister calls overview](./_attachments/transaction-overview.svg)
 
-Each type of canister operation and resource has a different cycles cost associated with it, and the canister responsible for paying the cycles varies based on the type of operation.
+Each type of canister operation and resource has a different cycles cost associated with it. The canister responsible for paying the cycles varies based on the type of operation.
 
 ### Canister creation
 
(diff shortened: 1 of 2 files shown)


docs/developer-docs/gas-cost.md
```diff
@@ -24,9 +24,9 @@ Canisters pay for consumed resources by burning cycles. These resources include
 
 ## Units and fiat value
 
-The price of cycles is fixed against the price of [XDR](/docs/current/references/glossary#xdr), where **1 trillion cycles equals 1 XDR**. As of December 18, 2023, the exchange rate for 1 XDR = $1.336610, which is used on this page. The exchange rate for USD/XDR may vary, and it will impact the conversion rate. You can view XDR exchange rates [here](https://www.imf.org/external/np/fin/data/rms_sdrv.aspx).
+The price of cycles is fixed against the price of [XDR](/docs/current/references/glossary#xdr), where **1 trillion cycles equals 1 XDR**. As of December 18, 2023, the exchange rate for 1 XDR = $1.336610. The exchange rate for USD/XDR may vary. Learn more about [XDR exchange rates](https://www.imf.org/external/np/fin/data/rms_sdrv.aspx).
```
samuelburri, line 27: I propose to shorten sentences with links by not separating the description of where the visitor might want to go and the link (see also my comment on Slack):

Instead of "You can view XDR exchange rates [here](https://www.imf.org/external/np/fin/data/rms_sdrv.aspx)"
Write "More [XDR exchange rates](https://github.com/dfinity/portal/pull/2297/link)".

docs/developer-docs/gas-cost.md
```diff
@@ -24,9 +24,9 @@ Canisters pay for consumed resources by burning cycles. These resources include
...
 
-This documentation will use the following units to measure and calculate the number of cycles and their associated fiat value:
+This documentation will use the following units:
```
samuelburri, line 29: We can be even shorter: "The following table shows units of cycles and the respective fiat values".

docs/developer-docs/gas-cost.md
```diff
@@ -38,26 +38,28 @@ This documentation will use the following units to measure and calculate the num
 
 ## The ICP reverse gas model
```
samuelburri, line 39: Can we remove "ICP" or did you do that with a specific intention?

docs/developer-docs/gas-cost.md
```diff
@@ -38,26 +38,28 @@ This documentation will use the following units to measure and calculate the num
...
 
-In contrast to other blockchains where users pay for their transactions, ICP charges the smart contract canister for the resources it consumes. This allows developers to provide a smoother user experience, as users are free from tedious tasks such as signing and approving every transaction they perform within the application. This cost model is known as ICP's 'reverse gas model'.
+ICP charges the smart contract canister for the resources it consumes. This allows developers to provide a smoother user experience, as users are free from tedious tasks such as signing and approving every transaction they perform. This cost model is known as ICP's 'reverse gas model'.
```
samuelburri, line 41: - "where users pay for their transactions" ... I'm wondering whether we should write "end users". Or "the users of the smart contract". Because the dev can also be seen as a user.
- 2 years ago, Dom insisted that we should write "canister smart contract". Let's please stick to it, i.e. not write "smart contract canister"

docs/developer-docs/gas-cost.md
```diff
@@ -38,26 +38,28 @@ This documentation will use the following units to measure and calculate the num
...
 
 :::info
-While this may be referred to as the ICP reverse 'gas' model, it is important to note that ICP does not use 'gas' -- it uses cycles.
+It is important to note that ICP does not use 'gas' -- it uses cycles.
```
samuelburri, line 44: ... ICP does not use the term 'gas', it uses the term 'cycles'.

docs/developer-docs/gas-cost.md
```diff
@@ -38,26 +38,28 @@ This documentation will use the following units to measure and calculate the num
...
 :::
 
-You can learn more about the ICP reverse gas model [here](https://internetcomputer.org/capabilities/reverse-gas).
+You can learn more about the [ICP reverse gas model](https://internetcomputer.org/capabilities/reverse-gas).
```
samuelburri, line 47: See comment above on "ICP reverse gas model" vs "reverse gas model"

docs/developer-docs/gas-cost.md
```diff
@@ -83,18 +85,18 @@ Note that query messages are currently free, but this may change in the future.
```
samuelburri, line 85: Comment was about the sentence "An additional factor of cost includes the number of nodes in the subnet the canister is deployed on.":

Is this scaling factor not applicable to all operations? If so, we might want to state that in the beginning of the section. For example after the sentence "Each type of canister operation and resource has a different cycles cost associated with it, and the canister responsible for paying the cycles varies based on the type of operation."
we could write:
"The following subsections explain the pricing for the different operation and resource types for a 13 node subnet. The same operations and resources cost linearly more on subnets with more nodes."

docs/developer-docs/gas-cost.md
```diff
@@ -83,18 +85,18 @@ Note that query messages are currently free, but this may change in the future.
...
+By default canisters are scheduled for execution in a "best-effort" manner. Canisters that require guaranteed execution can get a share of compute capacity by setting `compute_allocation` in their canister settings. Compute allocation is expressed in percents and denote the percentage of an execution core reserved for the canister. For example, compute allocation of 50% means that the canister will get 50% of an execution core. It will be scheduled at least every other round. Compute allocation of 100% means that the canister will be scheduled every round. The current fee for 1 percent computer allocation per second is 10M cycles (or $0.0000133661 USD).
 
 ### Storage
 Canisters pay for storage consumed by their Wasm memory and stable memory. The storage cost depends on the number of consumed bytes and the time duration. The current fee is 127K cycles per second (or $0.00000016974947 USD).
```
samuelburri, line 93: - Can we add a link to an explanation of WASM memory and stable memory?
- Can bytes in k cycles per second is precise but very hard to map to the real world. I propose to
   - rewrite "Canisters pay for storage consumed by their Wasm memory and stable memory. The storage cost depends on the number of consumed bytes and the time duration. The current fee is 127K cycles per second (or $0.00000016974947 USD)."
   - to "Canisters pay for storage consumed by their Wasm memory and stable memory per time. Storing 1 byte for 1 second costs 127k cycles, wich amounts to X T cycles (Y USD) for storing 1 MB for 1 year."

docs/developer-docs/gas-cost.md
```diff
@@ -109,47 +111,45 @@ The amount of reserved cycles depends on how many bytes are allocated and on the
...
 
-- **HTTPS outcalls**: The cost of HTTPS outcalls is calculated using a unique formula since the feature includes a quadratic component in its implementation. The cost for an HTTPS outcall is calculated using the formula `(3_000_000 + 60_000 * n) * n` for the base fee and `400 * n` each request byte and `800 * n` for each response byte, where `n` is the number of nodes in the subnet. These costs are included in the chart found below.
+- **HTTPS outcalls**: The cost for an HTTPS outcall is calculated using the formula `(3_000_000 + 60_000 * n) * n` for the base fee and `400 * n` each request byte and `800 * n` for each response byte, where `n` is the number of nodes in the subnet. These costs are included in the chart found below.
 
 - **Bitcoin API**: Pricing for the **Bitcoin API** is available in the [Bitcoin API documentation](./integrations/bitcoin/bitcoin-how-it-works.md).
 
 - **Chain-key signing API**: Pricing for the **Chain-key signing API** is available in the [Chain-key signing / threshold ECDSA documentation](./integrations/t-ecdsa/t-ecdsa-how-it-works.md).
 
 ### Replication
```
samuelburri, line 126: I wonder whether we should move the replication subsection to the top

docs/developer-docs/gas-cost.md
```diff
@@ -109,47 +111,45 @@ The amount of reserved cycles depends on how many bytes are allocated and on the
...
 
-Replication refers to the number of times a canister is replicated, which will depend on the subnet that the canister is deployed to.
+Replication refers to the number of times a canister is replicated, which is dependant on the canister's subnet.
 
 #### Local development
 
-When a canister is deployed to a local development environment, the canister is deployed to a single node (the locally running instance of the replica). Cycles charged to locally deployed canisters have a cost that is 1/13th the cost when deployed to a 13-node subnet. In local development environments, cycles can be fabricated using dfx.
+When a canister is deployed to a local development environment, the canister is deployed to a single node. Cycles charged to locally deployed canisters have a cost that is 1/13th the cost when deployed to a 13-node subnet. In local development environments, cycles can be fabricated using dfx.
```
samuelburri, line 132: Makes me think about possible improvements in our tooling.
* It seems very complicated to have a 1/13th of the cost locally. Can we not just completely remove any explanation of the local setup? Do I need to care about this as a developer? 
* Could we implement more of a prediction where the dev could be told how much he would need to pay if they were to run what they have locally on an n-node subnet on the IC.

docs/developer-docs/gas-cost.md
```diff
@@ -109,47 +111,45 @@ The amount of reserved cycles depends on how many bytes are allocated and on the
...
+Canisters are responsible for paying cycles for their own canister creation, compute resources, storage resources, and execution resources. For certain canister calls, the canister responsible for paying the cycles may vary.
 
-For inter-canister calls and the associated bytes, the canister sending this call is responsible for paying the required cycles. If there is an inter-canister call chain, for example, canister A → canister B → canister C, then canister A pays for the sending request-response pair between canister A and canister B, and canister B pays for sending the request-response pair between canister B and canister C. Additionally, each canister pays for processing that happens locally, meaning canister A pays for processing the response from canister B while canister B pays for processing the request from canister A.
+A canister sending an inter-canister call is responsible for paying the required cycles. In inter-canister call chains, each canister is responsible for the call that it sends. Each canister pays for processing that happens locally.
```
samuelburri, line 144: How about structuring such explanations more as bullet lists. I think ingress is most important and should therefore be on top.
* **Ingress messages:** bla, bla
* **Inter-canister calls:** bla, bla
* **Child canisters:** bla, bla

docs/developer-docs/gas-cost.md
```diff
@@ -109,47 +111,45 @@ The amount of reserved cycles depends on how many bytes are allocated and on the
...
 
 Canisters that are created from another canister (child canisters) are responsible for paying for themselves.
 
 ## Cycles price breakdown
 
-The chart below details the cost of compute and storage transmissions as well as management canister calls. A thorough example of how the cost of running a canister on a 13-node app subnet is computed can be found [here](https://wiki.internetcomputer.org/wiki/Comparing_Canister_Cycles_vs_Performance_Counter).
-
-The USD cost for canister transmissions below is based on the cycle costs indicated for 13-node and 34-node subnets. 1 XDR is equal to 1 trillion cycles. The cost is shown in USD (XDR/USD exchange rate as of December 18, 2023). The exchange rate for USD/XDR may vary, and it will impact the conversion rate. You can view XDR exchange rates [here](https://www.imf.org/external/np/fin/data/rms_sdrv.aspx).
+The chart below details the cost of compute, storage transmissions and canister calls. A thorough example of how the cost of running a canister on a 13-node subnet is computed can be found [on the wiki](https://wiki.internetcomputer.org/wiki/Comparing_Canister_Cycles_vs_Performance_Counter).
```
samuelburri, line 152: - is this a `chart`? isn't it rather a `table`?
- avoid the here-link-pattern

docs/developer-docs/gas-cost.md
```diff
@@ -109,47 +111,45 @@ The amount of reserved cycles depends on how many bytes are allocated and on the
...
 
 | Canister transmission | Description | Who is responsible for paying the cycles fee? | Local development ([IC SDK](./setup/index.md)) | 13-node application subnets cycles cost | 13-node application subnets USD cost | 34-node application subnets | 34-node application subnets USD cost |
```
samuelburri, line 154: - What's the point of having the local development in the table?
- We write "application subnet" but didn't explain on the page what an application subnet actually is. I would simply drop "application" in all column titles (yes, I know for system subnets it's free, but devs will not deploy to these subnets anyway".

docs/developer-docs/gas-cost.md
```diff
@@ -177,22 +177,22 @@ The following table shows the calculated storage cost per GB for a 30-day month
 
 ## Monitoring canister cycle usage
```
samuelburri, line 178: IMO these last 4 sections could be summarized under "Further readings" and then use a bullet list

## Further readings
* **Monitoring cycle usage**
   * [Check your canister's cycles balance](link)
   * [Monitor your canister's cycles usage](link)
   * [ Motoko cycles management library](link)
* **Topping up canisters**
   * ...

samuelburri (changes requested):
It seems my earlier feedback is not visible to everybody. I copy it over to this PR. Some feedback was already addressed by @jessiemongeon1  . For simplicity, I still copy over most of my earlier feedback.

**Feedback on the title:** 
I'm not totally happy with the title yet. transmission seems to only cover a subset of the resources that we cover on the page. I guess we landed here starting from transactions.

How about "Resource pricing in cycles", "Cycle prices for resource consumption", or "Paying for resources in cycles"? I think we want Cycles in the title and because most people have no clue what cycles mean we should combine it with words that have an obvious meaning, e.g. paying, resource usage, ...

docs/developer-docs/gas-cost.md
jessiemongeon1: Revised to " Learn more about [XDR exchange rates](link)"

docs/developer-docs/gas-cost.md
jessiemongeon1: Revised to "The following table shows units of cycles and the respective fiat values"

docs/developer-docs/gas-cost.md
jessiemongeon1: Revised to "ICP does not use the term 'gas', it uses the term 'cycles'."

docs/developer-docs/gas-cost.md
jessiemongeon1: Revised to 'end users'

Revised to 'canister smart contract' and added a note to the Dev Docs Style Guide that this is the correct order of words.

docs/developer-docs/gas-cost.md
jessiemongeon1: Removed 'ICP'

docs/developer-docs/gas-cost.md
jessiemongeon1: Removed "An additional factor of cost includes the number of nodes in the subnet the canister is deployed on."

Added "The following subsections explain the pricing for the different operation and resource types for a 13 node subnet. The same operations and resources cost linearly more on subnets with more nodes."

docs/developer-docs/gas-cost.md
jessiemongeon1: Revised 'chart' to 'table'

docs/developer-docs/gas-cost.md
jessiemongeon1: Revised to "Canisters pay for storage consumed by their Wasm memory and stable memory per time. Storing 1 byte for 1 second costs 127k cycles, which amounts to 3.684B cycles (or $5.35 USD) for storing 1 MB for 1 year."

Added links to Wasm and Stable Memory docs.

docs/developer-docs/gas-cost.md
jessiemongeon1: I think it's important to keep information on the local setup, since cycles are still charged in local development. This has been a point that the DevRel team has made efforts to emphasis, so I think it should be kept in to not give the impression that cycles aren't charged locally.

docs/developer-docs/gas-cost.md
jessiemongeon1: Removed 'Application' from the titles.

Removed 'local development' from the table, since while cycles are still charged locally, they can be fabricated and therefore have no real 'cost'.

docs/developer-docs/gas-cost.md
jessiemongeon1: Revised to use a bulleted list under 'Further readings'

docs/developer-docs/gas-cost.md
jessiemongeon1: Moved Replica subsection to the top of the page

docs/developer-docs/gas-cost.md
jessiemongeon1: Revised to use a bulleted list.

samuelburri (approved):
LGTM. Thank you!

dfx-json (approved):

PR #2356, opened 2024-01-18
Update security best practices with journaling

andrew-lee-work:
- Update how handle risks of outcalls to possibly untrusted canisters.
- Update how to handle traps.
- Add journaling section


docs/developer-docs/security/rust-canister-development-security-best-practices.md
```diff
@@ -262,8 +262,147 @@ Every inter-canister call is guaranteed to receive a response, either from the c
...
+- Value: list of references to Transaction journal structs
+
+Transaction journal struct
+- Meta data
+  - transaction, including memo
+  - timestamp
+- List of events, e.g. initialized, returned/success, returned/error with error condition
+- Number of times attempted and failed to progress
+
+#### Example Journaling Flows
```
domwoe, line 387: @andrew-lee-work Can we also link to implementations of this journaling?

andrew-lee-work: There is an implementation of journaling in https://github.com/GoldDAO/gldt-swap/blob/develop/canister/gldt_core/src/lib.rs#L654. I will add it.

docs/developer-docs/security/rust-canister-development-security-best-practices.md
```diff
@@ -262,8 +262,151 @@ Every inter-canister call is guaranteed to receive a response, either from the c
...
+
+We proceed to discuss “journaling”, which is currently our recommended way of addressing the problem at hand. 
+
+### Journaling
+
+Journaling can be used for ensuring that tasks are completed correctly in an asynchronous context, where any instruction or async task can fail. Journaling is generally useful in any security critical application canister on the IC. The journaling concept we describe here is inspired and adapted from journaling in file systems.
+
+A journal is a chronological list of the records and is kept in the canister’s storage. It keeps track of tasks before they begin and when they are completed. Before each failable task, the journal records the intent to execute the task, and after the task, the journal records the result. Creating a record in the journal is called “journaling”.  For example, to make an unreliable async call to a ledger:
```
oggy-dfin, line 308: How is a journal replay triggered? Is it by a user message retrying the procedure? If that's the case, should the user message also be logged in the journal, so that the successfully executed parts of handling the call are not done twice? If not (e.g., if it's a heartbeat that triggers the recovery), how does the user learn about the outcome of its call?

docs/developer-docs/security/rust-canister-development-security-best-practices.md
```diff
@@ -262,8 +262,151 @@ Every inter-canister call is guaranteed to receive a response, either from the c
...
+
+   - An “in progress” transfer would show in the journal as an entry containing intent to do the transfer without an entry containing the result of the transfer call.
+
+1. Call ledger to transfer 1 token from A to B.
+
+1. Journal the result of the transfer.
+
+   - On failure, record the error.
+
+   - On success, record success.
```
oggy-dfin, line 320: Doesn't this require ensuring that the callback doesn't trap, which we said we wanted to avoid? If it did, you wouldn't be able to record success, correct? It seems we need idempotency on the callee, right? (which concretely for the ledger we already have to an extent, as there exists a 24-hour dedup window)

oggy-dfin (commented):
I like the approach in theory but I'm not sure if it can be easily put into practice with the main IC canisters as they stand right now - see comments below.

docs/developer-docs/security/rust-canister-development-security-best-practices.md
```diff
@@ -262,8 +262,151 @@ Every inter-canister call is guaranteed to receive a response, either from the c
...
+1. The app calls the ledger to determine whether a transaction with the journaled parameters has succeeded on the ledger. Note that with the ICP ledger, there is a small possibility the transaction is stuck in the message queue, so one needs to wait 24 hours from the deduplication time stamp to ensure the message has either succeeded or failed and is not in transit. Therefore if the ledger does not contain the transaction and it is still within 24 hours of the deduplication timestamp, this step must be repeated.
+1. The app journals the result of the transfer call.
+1. The app journals the intention to update internal state according to the result of the transfer call, then updates the internal state, and finally journals the result of the attempt to update the internal state. (Journaling this step is still useful even if it does not contain outcalls, because outcalls may be introduced later, and the step could conflict with other processes that are not atomic.)
+
+Note that recovery can be complex in many cases. Therefore, it may be expedient to rely on manual recovery, then progressively implement automated recovery.
+
+#### Example Journaling Structures
```
robin-kunzler, line 366: Haven't reviewed the example sections here in full detail yet. Did you intend to push these as is? It seems they could benefit from some context, e.g. explaining that this is just a sketch/idea of how a journal could be implemented for the example discussed above. Also happy to review more closely and give more concrete suggestions. Maybe we could turn it into a paragraph just outlining this design on a high level.

robin-kunzler (commented):
Thanks a lot @andrew-lee-work ! LGTM, left a few comments / questions, see below. I think we need to decide what to do given @oggy-dfin 's concerns.

docs/developer-docs/security/rust-canister-development-security-best-practices.md
```diff
@@ -262,8 +262,151 @@ Every inter-canister call is guaranteed to receive a response, either from the c
...
+   - Examples of blocked tasks:
+
+     - On failure, log the failure in a user-visible log, and if less than 5 failures have occurred, make a new transfer outcall with the same parameters.
+
+     - On success, update the internal accounting of assets to conform to the result of the transfer.
+
+   - Note that any independent task does not need to wait for any part of this flow.
+
+The critical property of the journal is that at any point, if there is a failure, the journal is sufficient to determine what the next safe step should be. If, after step 1 (journal the intent),
+there is a failure in step 2 or 3, and step 3 (record the result) has not been completed, then the application should complete step 3 by finding out what happened to the call in step 2.
```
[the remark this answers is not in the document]
andrew-lee-work: Yes, although it is tedious, I think there is a way to correctly check the status of a transaction.

We could make sure the ledger is past the 24 hour limit by resending the transaction and seeing if it says it is too old.

Then we should be able to scan the blocks backwards in time to check for our transaction. Asynchrony should not be a big problem because we can keep track of the block height as we scan. (It may be necessary to have checked the block height before sending the transaction, so we know when to stop scanning.) If we don't find the transaction then it must have failed.

docs/developer-docs/security/rust-canister-development-security-best-practices.md
```diff
@@ -262,8 +262,151 @@ Every inter-canister call is guaranteed to receive a response, either from the c
...
+
+Extending the ledger example above, the recovery process can look like the following
+
+1. There is a panic and the status of the ledger call is unknown. However, the journal has recorded that a call to transfer with particular parameters and memo has been made, and the deduplication timestamp of the transfer. 
+1. The app calls the ledger to determine whether a transaction with the journaled parameters has succeeded on the ledger. Note that with the ICP ledger, there is a small possibility the transaction is stuck in the message queue, so one needs to wait 24 hours from the deduplication time stamp to ensure the message has either succeeded or failed and is not in transit. Therefore if the ledger does not contain the transaction and it is still within 24 hours of the deduplication timestamp, this step must be repeated.
```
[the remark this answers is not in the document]
andrew-lee-work: The transaction could be also stuck at some point before entering the message queue. In that case, it seems this step is still necessary. If we agree, I will change the text to describe the case where the transaction message is stuck before entering the ledger's message queue.

docs/developer-docs/security/rust-canister-development-security-best-practices.md
andrew-lee-work: If the call should be idempotent, the journal should contain enough information so that if the user retries, the app can determine which subtasks were successful and do not need to be redone. The app should then attempt to recover and continue the task. 

A recovery process could also be triggered on heartbeat. In this case, I would consider two options:
* The recovery process only resolves the status of unresolved entries and takes no actions. It is up to the user to decide if they want to continue the task or abort/clean up.
* The recovery process not only resolves unresolved entries, but it also attempts to continue the task. In this case the app would need to allow the user to check the status of the task. If the task is idempotent, the user could resubmit and see the status in the response.

docs/developer-docs/security/rust-canister-development-security-best-practices.md
andrew-lee-work: Yes I agree that it would be nice to have better commit atomicity, and that without it, we would need to either commit with an update call or simply allow that the success record is erased on a trap. If it is lost on trap, the next step in the journaling procedure would be to check the status of the transfer again.

docs/developer-docs/security/rust-canister-development-security-best-practices.md
oggy-dfin: Please use a permalink so that it stays stable; the link is already broken for me.

docs/developer-docs/security/rust-canister-development-security-best-practices.md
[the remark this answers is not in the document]
oggy-dfin: > Agree this seems tedious, also I guess we have no guarantees that a transaction will eventually make it to the archive, right? So one can never be sure that a transaction did _not_ happen?

@robin-kunzler I'm not sure I understood this point. Assuming that we did have a "has this transaction happened" endpoint on the ledger, our message ordering guarantees would ensure that once a "has the transaction happened" message makes it to the ledger, any prior "perform this transaction" message must also have made it to the ledger - or will never make it. So unless executing "perform this transaction" message is spread over multiple message handlers (because there are downstream calls involved), and unless the effects of the message are delayed, the second message would give you a definite answer.

docs/developer-docs/security/rust-canister-development-security-best-practices.md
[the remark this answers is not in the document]
andrew-lee-work: I see, I misunderstood your earlier message as being that execution of messages is the same as ordering in the message queue.

That aside, I'd rephrase the question of interest as: Does the guaranteed ordering apply all the way from the caller message execution to the callee message execution? I.e. does "(order of caller execution of a message `M1[i]` that makes a call `M2[i]`) = (order of callee message execution of `M2[i]`)"? It sounds like is the case, but I would like to confirm, since it is a distinct statement from "order of execution = order of sending assuming both messages are executed".

docs/developer-docs/security/rust-canister-development-security-best-practices.md
andrew-lee-work: OK, changing to https://github.com/GoldDAO/gldt-swap/blob/ledger-v1.0.0/canister/gldt_core/src/lib.rs . This is a tag, so is more permanent.

docs/developer-docs/security/rust-canister-development-security-best-practices.md
```diff
@@ -262,8 +262,151 @@ Every inter-canister call is guaranteed to receive a response, either from the c
...
+- Remove transaction from both Maps
+- Consider creating an audit log event
+
+Find transactions in unfinished state: 
+- Iterate through Transactions by timestamp map (oldest first)
+- Transactions older than e.g. 10min back can be considered to be in error state
+
+#### Example implementation of journaling
+
+GoldDAO's GLDT-swap has an implementation of journaling. See https://github.com/GoldDAO/gldt-swap/blob/develop/canister/gldt_core/src/lib.rs#L654. In their case the journal entries are recorded in the "registry". Note that in GLDT-swap there is also a separate concept of "record" which is a permanent audit trail, and is not used for journaling.
```
andrew-lee-work, line 404:
```suggestion
GoldDAO's GLDT-swap has an implementation of journaling. See https://github.com/GoldDAO/gldt-swap/blob/ledger-v1.0.0/canister/gldt_core/src/lib.rs#L654. In their case the journal entries are recorded in the "registry". Note that in GLDT-swap there is also a separate concept of "record" which is a permanent audit trail, and is not used for journaling.
```

docs/developer-docs/security/rust-canister-development-security-best-practices.md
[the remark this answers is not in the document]
andrew-lee-work: GoldDAO uses journaling to keep track of the process of acquiring a gold NFT and minting GLDT tokens. This happens in `notify_sale_nft_origyn`. AFAIK GoldDAO also requires manual recovery on trap. However, it may be that automated recovery can be implemented too much trouble. It depends on the Origyn NFT interface for determining the status of one's attempt to acquire an NFT. I would expect it's easier than checking the status of a ledger transaction. At least GoldDAO can determine if they own the NFT.

We also know of another project that does journaling and also uses manually recovery on failures. There is special functionality for principals with admin privileges to refund funds. In this project the journal is for dealing with ledger call failures, so it would tedious to resolve automatically. (Note that this is not yet open sourced so I won't reference it explicitly. We could consider adding the link later when the project is published.)

I agree we should at least recommend journaling with manual recovery. Early-stage projects that we've looked at that contain journaling appear to do only manual recovery. However, as in the case of GoldDAO, it can be practical to do automated recovery as the project matures.

Note that there is an index canister for SNSs. I would need to check but there might be one for the ICP ledger too. The index canister indexes ledger transactions for easier searching.

docs/developer-docs/security/rust-canister-development-security-best-practices.md
```diff
@@ -262,8 +262,151 @@ Every inter-canister call is guaranteed to receive a response, either from the c
...
+
 ### Avoid traps after await
 
+*This section is deprecated in favor of [Securely handle traps in callbacks](#Securely-handle-traps-in-callbacks)*
```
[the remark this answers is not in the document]
andrew-lee-work: I would like to remove it in a separate PR. It would make it easier to reinstate the section in case we want to do that later.

andrew-lee-work:
Follow-up issue: https://github.com/dfinity/portal/issues/2727

docs/developer-docs/security/rust-canister-development-security-best-practices.md
andrew-lee-work: I've added a description of retries.

docs/developer-docs/security/rust-canister-development-security-best-practices.md
andrew-lee-work: I've given more context and clarified this section.

docs/developer-docs/security/rust-canister-development-security-best-practices.md
```diff
@@ -262,8 +262,166 @@ Every inter-canister call is guaranteed to receive a response, either from the c
...
+Note that in canisters implemented in Rust with Rust CDK version 0.5.1, any local variables still go out of scope if a callback traps. The CDK actually calls into the ic0.call_on_cleanup API to release these resources. This helps to prevent issues with locks not being released, as it is possible to use Rust's Drop implementation to release locked resources, as we discuss in [Be aware that there is no reliable message ordering](https://internetcomputer.org/docs/current/developer-docs/security/rust-canister-development-security-best-practices#be-aware-that-there-is-no-reliable-message-ordering).
+
+#### Recommendation
+
+Recall that the responses to inter-canister calls are processed in the corresponding callback. If the callback traps, the cleanup (ic0.call_on_cleanup) is executed. When making an inter-canister call, the ICP reserves sufficiently many cycles to execute the response callback or cleanup (up to the instruction limit). A fixed fraction of the reservation is for the cleanup. Thus, a response or cleanup execution can never “run out of cycles”, but they can run into the instruction limit and trap.
```
oggy-dfin, line 284: Do you actually know which percentage is reserved? This sounds like something we should document, maybe [here](https://internetcomputer.org/docs/current/developer-docs/gas-cost#canister-calls)?
Can you still hit that percentage limit in the `call_on_cleanup`?

docs/developer-docs/security/rust-canister-development-security-best-practices.md
```diff
@@ -262,8 +262,166 @@ Every inter-canister call is guaranteed to receive a response, either from the c
...
+
+   - An “in progress” transfer would show in the journal as an entry containing intent to do the transfer without an entry containing the result of the transfer call.
+
+1. Call ledger to transfer 1 token from A to B.
+
+1. Journal the result of the transfer.
+
+   - On failure, record the error.
+
+   - On success, record success. (In order to commit the record, intercanister call can be made to an endpoint on the same canister that does nothing. Otherwise a trap could erase the journaled result, complicating recovery.)
```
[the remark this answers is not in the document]
oggy-dfin: "an inter-canister call"?
More importantly, should we recommend this call to be made? As-is, this sounds optional, but as discussed previously, without it a trap would render the journal useless for automated recovery (it would still be useful for manual debugging).

docs/developer-docs/security/rust-canister-development-security-best-practices.md
```diff
@@ -262,8 +262,166 @@ Every inter-canister call is guaranteed to receive a response, either from the c
...
+the ledger outcall. The lock will not get stuck assuming the application can always find out what happened to a call. Enough context about the call should be recorded in the intent to ensure
+that this is the case. For the ICP ledger, an ID can be generated and recorded in the journaled intent and the ledger can be called with the ID included in the memo so that the result of the
+call can be queried later.
+
+#### Journaling is Robust to Panics
+
+Continuing the above example, let us consider a panic at any point. 
+1. If there is panic before the async outcall, then the journaled intent will be lost, but no state change occurred internally and no outcalls were made so the app is in a safe state. The next step is to record a new intent.
+1. If there is a panic after the async outcall, then the journaled result (step 3) will be lost. This means the app will need to determine the result and journal it before continuing to step 4. As long as it is possible to determine the result, the app is in a safe state even.
```
oggy-dfin, line 349:
```suggestion
1. If there is a panic after the async outcall and no self-call was used to commit the journal, the journaled result (step 3) will be lost. This means the app will need to determine the result and journal it before continuing to step 4. As long as it is possible to determine the result, the app can be brought back to a consistent state.
```

docs/developer-docs/security/rust-canister-development-security-best-practices.md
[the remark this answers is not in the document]
oggy-dfin: In particular, the second paragraph suggests in this section suggests that the journal can be automatically replayed. Let's make it more nuanced?

docs/developer-docs/security/rust-canister-development-security-best-practices.md
```diff
@@ -262,8 +262,166 @@ Every inter-canister call is guaranteed to receive a response, either from the c
...
+1. The app calls the ledger to determine whether a transaction with the journaled parameters has succeeded on the ledger. Due to the guarantee that any pair of messages that are both executed are always executed in the order issued, if the ledger indicates that the transaction has not occurred, then the transaction will never occur.
+1. The app journals the result of the transfer call.
+1. The app journals the intention to update internal state according to the result of the transfer call, then updates the internal state, and finally journals the result of the attempt to update the internal state. (Journaling this step is still useful even if it does not contain outcalls, because outcalls may be introduced later, and the step could conflict with other processes that are not atomic.)
+
+Note that querying the ICP ledger or an ICRC ledger to determine whether a transaction has succeeded is not straightforward to automate, so it could be done manually.
+
+#### Example journaling structures and flows
```
oggy-dfin, line 369: Is it possible to illustrate these on the GoldDAO example, it might be more concrete?

docs/developer-docs/security/rust-canister-development-security-best-practices.md
```diff
@@ -262,8 +262,166 @@ Every inter-canister call is guaranteed to receive a response, either from the c
...
+- Currently, Motoko does not allow setting custom cleanup callbacks. Cleanup is used internally by Motoko to do some cleanups.
+- As discussed above, the Rust CDK has a feature that automatically releases local variables in cleanup, which [can be used to release locks](https://internetcomputer.org/docs/current/developer-docs/security/rust-canister-development-security-best-practices#be-aware-that-there-is-no-reliable-message-ordering). Since only one cleanup callback can be defined, any custom cleanup would currently have to implement that feature itself if needed, making this currently hard to use and understand.
+
+We proceed to discuss “journaling”, which is currently our recommended way of addressing the problem at hand. 
+
+#### Journaling
+
+Journaling can be used for ensuring that tasks are completed correctly in an asynchronous context, where any instruction or async task can fail. Journaling is generally useful in any security critical application canister on the IC. The journaling concept we describe here is inspired and adapted from journaling in file systems.
```
roelstorms, line 306:
```suggestion
Journaling can be used to ensure that tasks are completed correctly in an asynchronous context, where any instruction or async task can fail. Journaling is generally useful in any security critical application canister on the IC. The journaling concept we describe here is inspired and adapted from journaling in file systems.
```

docs/developer-docs/security/rust-canister-development-security-best-practices.md
```diff
@@ -262,8 +262,166 @@ Every inter-canister call is guaranteed to receive a response, either from the c
...
+
+Conceptually, a journal is a chronological list of records kept in a canister’s storage. It keeps track of tasks before they begin and when they are completed. Before each failable task, the journal records the intent to execute the task, and after the task, the journal records the result. The journal supports idempotent task flows by providing the necessary information for the canister to resume flows that failed to complete, report progress for ongoing flows, and report results for completed flows. Retries can be initiated by calls or automatically on heartbeat. If the task flow was completed in a heartbeat, a user can take advantage of idempotency to check the result.
```
roelstorms, line 308:
```suggestion
Conceptually, a journal is a chronological list of records kept in a canister’s storage. It keeps track of tasks before they begin and when they are completed. Before each failable task, the journal records the intent to execute the task, and after the task, the journal records the result. The journal supports idempotent task flows by providing the necessary information for the canister to resume flows that failed to complete, report progress for ongoing flows, and report results for completed flows. Retries can be initiated by calls, automatically on a [heartbeat](https://internetcomputer.org/docs/current/developer-docs/smart-contracts/advanced-features/periodic-tasks/#heartbeats) or using [timers](https://internetcomputer.org/docs/current/developer-docs/smart-contracts/advanced-features/periodic-tasks/#timers). If the task flow was completed in a heartbeat, a user can take advantage of idempotency to check the result.
```

roelstorms: > If the task flow was completed in a heartbeat, a user can take advantage of idempotency to check the result.
Isn't this also the case if the flow was initiated by a call or timer?

roelstorms:
I believe all references to IC documentation need to be made relatively. So all https://internetcomputer.org/docs/current/... need to be replaced. If had someone change my PR to fix this but maybe we can try to do that ourselves?

docs/developer-docs/security/rust-canister-development-security-best-practices.md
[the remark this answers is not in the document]
roelstorms: ICDex had a manual resolution mechanism that allowed some admin to retry failed transactions. I believe that is in line with our philosophy here.

> (In order to commit the record, intercanister call can be made to an endpoint on the same canister that does nothing. Otherwise a trap could erase the journaled result, complicating recovery.)

This sounds like we would benefit from a way to commit state or split messages using a system call to avoid having to use this hack? Is this a feature we want to add to the IC?

docs/developer-docs/security/rust-canister-development-security-best-practices.md
```diff
@@ -262,8 +262,166 @@ Every inter-canister call is guaranteed to receive a response, either from the c
...
+
+Creating a record in the journal is called “journaling”.  For example, to make an unreliable async call to a ledger:
+
+1. Check the journal to ensure the transfer is not already in progress. If it is already in progress, go into recovery (see [Recovery](#recovery) section below). Otherwise, journal the intent to call a ledger to transfer 1 token from A to B. The journaled intent should contain sufficient context to later identify what happened to the call.
+
+   - An “in progress” transfer would show in the journal as an entry containing intent to do the transfer without an entry containing the result of the transfer call.
+
+1. Call ledger to transfer 1 token from A to B.
+
+1. Journal the result of the transfer.
```
roelstorms, line 318:
```suggestion
2. Journal the result of the transfer.
```

Is the numbering messed up?

docs/developer-docs/security/rust-canister-development-security-best-practices.md
```diff
@@ -262,8 +262,166 @@ Every inter-canister call is guaranteed to receive a response, either from the c
...
+
+Creating a record in the journal is called “journaling”.  For example, to make an unreliable async call to a ledger:
+
+1. Check the journal to ensure the transfer is not already in progress. If it is already in progress, go into recovery (see [Recovery](#recovery) section below). Otherwise, journal the intent to call a ledger to transfer 1 token from A to B. The journaled intent should contain sufficient context to later identify what happened to the call.
+
+   - An “in progress” transfer would show in the journal as an entry containing intent to do the transfer without an entry containing the result of the transfer call.
+
+1. Call ledger to transfer 1 token from A to B.
```
roelstorms, line 316:
```suggestion
2. Call ledger to transfer 1 token from A to B.
```

docs/developer-docs/security/rust-canister-development-security-best-practices.md
roelstorms, line 349:
```suggestion
2. If there is a panic after the async outcall, then the journaled result (step 3) will be lost. This means the app will need to determine the result and journal it before continuing to step 4. As long as it is possible to determine the result, the app is in a safe state even.
```

docs/developer-docs/security/rust-canister-development-security-best-practices.md
```diff
@@ -262,8 +262,166 @@ Every inter-canister call is guaranteed to receive a response, either from the c
...
+#### Recovery
+
+The journal ensures the application knows that recovery from an error is needed, and aids in making recovery decisions. In order to support the recovery process, the journal should support querying all unresolved tasks of a certain type, and tasks of a certain type that resulted in an error. Given an intent, the journal should also be able to return the result if it exists and indicate if it does not exist.
+
+Note that recovery can often be complex to automate. In such cases, the journal can support a manual recovery process.
+Extending the ledger example above, a recovery process can look like the following
+
+1. There is a panic and the status of the ledger call is unknown. However, the journal has recorded that a call to transfer with particular parameters and memo has been made, and the deduplication timestamp of the transfer. 
+1. The app calls the ledger to determine whether a transaction with the journaled parameters has succeeded on the ledger. Due to the guarantee that any pair of messages that are both executed are always executed in the order issued, if the ledger indicates that the transaction has not occurred, then the transaction will never occur.
```
roelstorms, line 363:
```suggestion
2. The app calls the ledger to determine whether a transaction with the journaled parameters has succeeded on the ledger. Due to the guarantee that any pair of messages that are both executed are always executed in the order issued, if the ledger indicates that the transaction has not occurred, then the transaction will never occur.
```

docs/developer-docs/security/rust-canister-development-security-best-practices.md
```diff
@@ -262,8 +262,166 @@ Every inter-canister call is guaranteed to receive a response, either from the c
...
+1. The app journals the result of the transfer call.
```
roelstorms, line 364:
```suggestion
3. The app journals the result of the transfer call.
```

docs/developer-docs/security/rust-canister-development-security-best-practices.md
```diff
@@ -262,8 +262,166 @@ Every inter-canister call is guaranteed to receive a response, either from the c
...
+1. The app journals the intention to update internal state according to the result of the transfer call, then updates the internal state, and finally journals the result of the attempt to update the internal state. (Journaling this step is still useful even if it does not contain outcalls, because outcalls may be introduced later, and the step could conflict with other processes that are not atomic.)
```
roelstorms, line 365:
```suggestion
4. The app journals the intention to update internal state according to the result of the transfer call, then updates the internal state, and finally journals the result of the attempt to update the internal state. (Journaling this step is still useful even if it does not contain outcalls, because outcalls may be introduced later, and the step could conflict with other processes that are not atomic.)
```

docs/developer-docs/security/rust-canister-development-security-best-practices.md
```diff
@@ -262,8 +262,166 @@ Every inter-canister call is guaranteed to receive a response, either from the c
...
+Here is a sketch of how the above structures could be used.
+
+* When executing a transaction:
+  1. Look up the transaction in the audit log. If an entry exists, return it and exit this flow.
+  2. Look up the transaction flow in the `Journal by caller` map.
+  3. If no entry exists in `Journal by caller`, then:
+     1. Create a new `Transaction journal struct`.
+     2. Add an entry for the `Transaction journal struct` to `Journal by caller`,
+     3. Add an entry for the `Transaction journal struct` to `Transactions by timestamp`.
+  5. Initiate or resume the task flow according to the list of events in the `Transaction journal struct` that was previously obtained or created.
```
roelstorms, line 407:
```suggestion
  4. Initiate or resume the task flow according to the list of events in the `Transaction journal struct` that was previously obtained or created.
```

docs/developer-docs/security/rust-canister-development-security-best-practices.md
```diff
@@ -262,8 +262,166 @@ Every inter-canister call is guaranteed to receive a response, either from the c
...
+The naïve recommendation to address the security concern described above would be to avoid traps. However, that can be very difficult to achieve due to the following reasons: 
+
+- The implementation can be involved and could panic due to bugs, such as index out-of-bounds errors or panics (expect, unwrap) that should supposedly never happen.
+
+- It is hard to make sure the callback or cleanup doesn’t run into the instruction limit (and thus traps), because the number of instructions required can in general not be predicted and may e.g. depend on the data being processed. 
+
+Due to these reasons, while it is easy to recommend “avoiding traps”, this is actually hard to achieve in practice. So in our view, code should be written so that it can deal even with unexpected traps due to bugs or hitting the instruction limits. We discuss two approaches:
+
+1. Do simple cleanups 
+1. Do “journaling”
```
roelstorms, line 295:
```suggestion
2. Do “journaling”
```

docs/developer-docs/security/rust-canister-development-security-best-practices.md
roelstorms: Also fix the numbering here.

docs/developer-docs/security/rust-canister-development-security-best-practices.md
```diff
@@ -262,8 +262,166 @@ Every inter-canister call is guaranteed to receive a response, either from the c
...
+#### Journaling And Audit Events
+
+The journal can be used to augment the audit trail for recent events. However, it is probably too detailed for long term storage. After a while, journal entries could be compressed and incorporated into long term audit events. The process for creating audit events could itself be journaled.
+
+#### Recovery
+
+The journal ensures the application knows that recovery from an error is needed, and aids in making recovery decisions. In order to support the recovery process, the journal should support querying all unresolved tasks of a certain type, and tasks of a certain type that resulted in an error. Given an intent, the journal should also be able to return the result if it exists and indicate if it does not exist.
+
+Note that recovery can often be complex to automate. In such cases, the journal can support a manual recovery process.
+Extending the ledger example above, a recovery process can look like the following
```
roelstorms, line 360:
```suggestion
Extending the ledger example above, a recovery process could look as follows:
```

docs/developer-docs/security/rust-canister-development-security-best-practices.md
```diff
@@ -262,8 +262,166 @@ Every inter-canister call is guaranteed to receive a response, either from the c
...
+
+1. There is a panic and the status of the ledger call is unknown. However, the journal has recorded that a call to transfer with particular parameters and memo has been made, and the deduplication timestamp of the transfer.
```
roelstorms, line 362:
```suggestion
1. There is a panic and the status of the ledger call is unknown. However, the journal has recorded that a call to transfer with particular parameters and memo has been made, including the deduplication timestamp of the transfer. 
```

docs/developer-docs/security/rust-canister-development-security-best-practices.md
```diff
@@ -408,26 +566,29 @@ Not handling the error cases correctly is risky: for example, if a ledger transf
...
 
-### Only make inter-canister calls to trustworthy canisters
+### Be aware of the risks involved in calling untrustworthy canisters
 
 #### Security concern
 
 - If inter-canister calls are made to potentially malicious canisters, this can lead to DoS issues or there could be issues related to candid decoding. Also, the data returned from a canister call could be assumed to be trustworthy when it is not.
 
-- If a canister is called with a callback, the receiver can stall indefinitely if the peer does not respond, resulting in DoS. A canister can no longer be upgraded if it is in that state. Recovery would involve reinstalling, wiping the state of the canister.
+- If a canister is called with a callback, the receiver can stall indefinitely if the peer does not respond, resulting in DoS. A canister can no longer be upgraded if it is in that state. Recovery would involve reinstalling, wiping the state of the canister. Note that even a trustworthy canister could have a bug causing it to stall indefinitely. However, such a bug seems rather unlikely to occur.
```
roelstorms, line 575:
```suggestion
- When another canister is called with a callback being registered, and the receiver stalls the response indefinitely by not responding, the result would result in a DoS. Additionally, that canister can no longer be upgraded if it has callbacks registered. Recovery would require wiping the state of the canister by reinstalling it. Note that even a trustworthy canister could have a bug causing it to stall indefinitely. However, such a bug seems rather unlikely to occur. 
```

docs/developer-docs/security/rust-canister-development-security-best-practices.md
```diff
@@ -408,26 +566,29 @@ Not handling the error cases correctly is risky: for example, if a ledger transf
...
 
-- If a canister is called with a callback, the receiver can stall indefinitely if the peer does not respond, resulting in DoS. A canister can no longer be upgraded if it is in that state. Recovery would involve reinstalling, wiping the state of the canister.
+- If a canister is called with a callback, the receiver can stall indefinitely if the peer does not respond, resulting in DoS. A canister can no longer be upgraded if it is in that state. Recovery would involve reinstalling, wiping the state of the canister. Note that even a trustworthy canister could have a bug causing it to stall indefinitely. However, such a bug seems rather unlikely to occur. 
 
 - In summary, this can DoS a canister, consume an excessive amount of resources, or lead to logic bugs if the behavior of the canister depends on the inter-canister call response.
 
 #### Recommendation
 
-- Only make inter-canister calls to trustworthy canisters.
+- Making inter-canister calls to trustworthy canisters is safe, except for the rather unlikely case that there is a bug in the callee that makes it stall forever.
```
roelstorms, line 581: Actually, you also need to trust the target subnet. This is relevant for canisters on the NNS but might also apply in the future where we have rented subnets or subnets with smaller replication factors.

docs/developer-docs/security/rust-canister-development-security-best-practices.md
```diff
@@ -408,26 +566,29 @@ Not handling the error cases correctly is risky: for example, if a ledger transf
...
+- If a canister is called with a callback, the receiver can stall indefinitely if the peer does not respond, resulting in DoS. A canister can no longer be upgraded if it is in that state. Recovery would involve reinstalling, wiping the state of the canister. Note that even a trustworthy canister could have a bug causing it to stall indefinitely. However, such a bug seems rather unlikely to occur. 
 
 - In summary, this can DoS a canister, consume an excessive amount of resources, or lead to logic bugs if the behavior of the canister depends on the inter-canister call response.
 
 #### Recommendation
 
-- Only make inter-canister calls to trustworthy canisters.
+- Making inter-canister calls to trustworthy canisters is safe, except for the rather unlikely case that there is a bug in the callee that makes it stall forever. 
+
+- If you still want to call untrustworthy canisters, it’s possible to use a state-free proxy canister which could easily be re-installed if it is attacked as described above and is stuck. When the proxy is reinstalled, the caller obtains an error response to the open calls.
```
roelstorms, line 583:
```suggestion
- Interacting with untrustworthy canisters is still possible by using a state-free proxy canister which could easily be re-installed if it is attacked as described above and is stuck. When the proxy is reinstalled, the caller obtains an error response to the open calls.
```

docs/developer-docs/security/rust-canister-development-security-best-practices.md
```diff
@@ -262,8 +262,166 @@ Every inter-canister call is guaranteed to receive a response, either from the c
...
+
+* Map: Journal by caller
+  - Key: (caller, action type) – e.g. (Alice, Withdraw)
+  - Value: Journal struct
+
+* Map: Transactions by timestamp (ordered map, could use an additional list for the ordering in Motoko)
+  - Key: timestamp 
+  - Value: list of references to `Transaction journal struct`s
+
+* Transaction journal struct (compatible with `Journal struct`)
```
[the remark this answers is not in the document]
andrew-lee-work: The reason for distinguishing "Journal struct" from "Transaction Journal Struct" is that the journal can handle more than just transactions. If we make the journal only handle transactions, then the journal could be simplified to just a mapping from caller to transaction, and we wouldn't need to have the tuple (caller, action). 

We had made the initial decision when brainstorming this section that the journal should be able to handle more than one type of action. We could reverse that decision, but I do want to point out how deep this goes.

docs/developer-docs/security/rust-canister-development-security-best-practices.md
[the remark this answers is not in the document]
andrew-lee-work: > "an inter-canister call"? More importantly, should we recommend this call to be made? As-is, this sounds optional, but as discussed previously, without it a trap would likely render the journal useless for automated recovery (it would still be useful for manual debugging).

The journal is not useless even if a trap deletes the result. On resumption, the next task would be to obtain the result that was deleted. If it is cheap to obtain the result again and fill in the lost part of the journal, it could be beneficial to skip the extra call.

docs/developer-docs/security/rust-canister-development-security-best-practices.md
andrew-lee-work: IIRC @robin-kunzler made these changes, with the goal of making the rest of the doc consistent with the addition of the journaling section. I think it's somewhat of a scope creep to improve this section in this PR and I suggest putting this in a different ticket.

docs/developer-docs/security/rust-canister-development-security-best-practices.md
andrew-lee-work: @roelstorms Markdown automatically renders the numbering as "1., 2., ..." if you use `1.` for every item in a numbered list. You can check this in the markdown preview.

docs/developer-docs/security/rust-canister-development-security-best-practices.md
andrew-lee-work: One goal of this section is to provide a more abstract alternative to the GoldDAO example, to avoid having to explain too much background. It sounds like you are saying it is too abstract to be useful. In that case I would remove this section, and readers can look at GoldDAO.

docs/developer-docs/security/rust-canister-development-security-best-practices.md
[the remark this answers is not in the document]
andrew-lee-work: > This sounds like we would benefit from a way to commit state or split messages using a system call to avoid having to use this hack? Is this a feature we want to add to the IC?

@roelstorms this has been raised by various people, including @oggy-dfin. I'm not aware of any plan to add it. Perhaps @oggy-dfin would be the person best positioned to push for it.

docs/developer-docs/security/rust-canister-development-security-best-practices.md
andrew-lee-work: No, this is exercising a markdown WYSIWYM feature (see also my other comment on numbering).

docs/developer-docs/security/rust-canister-development-security-best-practices.md
andrew-lee-work: By using "for ensuring" instead of "to ensure", I was trying to make the subtle distinction that it is useful but not sufficient.

docs/developer-docs/security/rust-canister-development-security-best-practices.md
andrew-lee-work: I think trusting a canister implies trusting its subnet and various other dependencies. I'm not aware if we have documentation about what trusting a canister entails, but it if not it would be useful to create it and link to it from here.

docs/developer-docs/security/rust-canister-development-security-best-practices.md
roelstorms: I haven't found this info anywhere in the IC docs. If we don't add it we should definitely try to have it added somewhere as it is important for people that try to use this cleanup to comply with the best practice.

This reservation also wasn't obvious to me from the IC code so we might just ask @ulan

docs/developer-docs/security/rust-canister-development-security-best-practices.md
oggy-dfin: I thought it'd be helpful to illustrate the individual elements of the general journaling strategy on the GoldDAO example. Like point out where they implement support for the various queries that you mention below, or how they journal before and after inter-canister calls, how they do recovery, etc.

docs/developer-docs/security/rust-canister-development-security-best-practices.md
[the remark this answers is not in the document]
oggy-dfin: > The journal is not useless even if a trap deletes the result. On resumption, the next task would be to obtain the result that was deleted. If it is cheap to obtain the result again and fill in the lost part of the journal, it could be beneficial to skip the extra call.
You're right, it makes sense.

> Perhaps @oggy-dfin would be the person best positioned to push for it.
I will bring it up with the runtime/execution teams but I suspect it won't make it high up the priority list unless we can get some community requests for that.

docs/developer-docs/security/rust-canister-development-security-best-practices.md
roelstorms: Ok, makes sense. Then that last sentence also needs to specify timers. Updated my suggestion.

docs/developer-docs/security/rust-canister-development-security-best-practices.md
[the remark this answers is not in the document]
andrew-lee-work: This section will be removed. Resolving the convo.

docs/developer-docs/security/rust-canister-development-security-best-practices.md
andrew-lee-work: Add issue #2817. I recommend resolving this outside this PR.

docs/developer-docs/security/rust-canister-development-security-best-practices.md
andrew-lee-work: Fixed by using 1. for each item.

docs/developer-docs/security/rust-canister-development-security-best-practices.md
andrew-lee-work: Added a #2818. Will handle the issue outside of this PR.

docs/developer-docs/security/rust-canister-development-security-best-practices.md
```diff
@@ -262,8 +262,166 @@ Every inter-canister call is guaranteed to receive a response, either from the c
...
+* Map: Transaction audit log
+  - Key: transaction (hash)
+  - Value: result of (completed) transaction
+
+Note that `Journal by caller` is intended to support various types of flows, but only transaction flows are detailed explicitly. `Transactions by timestamp` is an auxiliary structure to be used for searching for transactions that require recovery.
+
+Here is a sketch of how the above structures could be used.
+
+* When executing a transaction:
+  1. Look up the transaction in the audit log. If an entry exists, return it and exit this flow.
```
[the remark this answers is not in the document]
andrew-lee-work: Added.

PR #2361, opened 2024-01-18
Dev Docs Accelerator: Revisions Part 1

jessiemongeon1:


docs/developer-docs/frontend/boilerplate-frontend.mdx
```diff
@@ -0,0 +1,470 @@
...
+
+- [x] Download and install the IC SDK package as described in the [installing the IC SDK](/docs/current/developer-docs/setup/install) page.
+
+- [x] Download and install [git](https://git-scm.com/downloads).
+
+- [x] Download and install [Node.js](https://nodejs.org/en) <v0.12.0.
+
+### Step 1: Clone the project repository.
+
+<Tabs>
```
dfx-json, line 22: nice!

dfx-json: I think we should remove the tabs for the code snippets that have no change between languages though

sidebars.js
```diff
@@ -411,10 +411,7 @@ const sidebars = {
       items: [
         "developer-docs/frontend/custom-frontend",
         "developer-docs/frontend/add-stylesheet",
-        "developer-docs/frontend/javascript-frontend",
```
dfx-json, line 414: do we need redirects for the old ones?

dfx-json (approved):

sidebars.js
jessiemongeon1: Yes, good catch thanks!

docs/developer-docs/frontend/boilerplate-frontend.mdx
jessiemongeon1: I think that may cause confusion for less-experienced devs who select one framework at the beginning and then the option to select disappears, as it might not be clear to them that the command will be the same for each one. I think there is a way to configure the tabs where if you select one, all the tabs on the page switch to the one you selected, which might be a good solution IMO.

PR #2410, opened 2024-01-29
Add Solidity to nav bar

jessiemongeon1:

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -85,6 +85,12 @@ const subnavItems = [
     label: "Python",
   },
   {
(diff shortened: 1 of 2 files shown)


dfx-json (approved):
I think we're reaching the maximum number of programming languages we should display horizontally on the nav. Should we consider making this a dropdown at some point?

jessiemongeon1:
> I think we're reaching the maximum number of programming languages we should display horizontally on the nav. Should we consider making this a dropdown at some point?

Yes, if we were to add more in the future, it would need to transition to a drop down I think.

PR #2425, opened 2024-01-31
Move what-is-ic.md to index.md

ulan:
This moves the contents of the 'What is IC' page to the introduction page.


jessiemongeon1:
Hadn't we discussed keeping the `what-is-ic` page name? Is there a specific reason for removing it and renaming it `index.md`? 

We had agreed that removing the current introduction page `index.md` was a good revision, but couldn't we just change the 'Build' tab link to go to `what-is-ic` instead of `index.md`? Removing `what-is-ic` is going to potentially break a lot of embedded links throughout the docs.

PR #2441, opened 2024-02-01
Revise NNS Intro

jennifertrin:


docs/tokenomics/nns/nns-intro.md
```diff
@@ -1,218 +1,53 @@
-# NNS intro
+# Network Nervous System (NNS)
 
-The NNS is the decentralized autonomous organization (DAO) that governs the Internet Computer Protocol. It is a fully on-chain, decentralized organization and is, for instance, responsible for making protocol level upgrades to continuously improve the Internet Computer Protocol. It does this via an implementation of liquid democracy in which ICP neuron holders vote on proposals that shape the development of the Internet Computer. Once such a proposal is accepted, it is autonomously executed.
+## Overview​
+The [NNS](https://nns.ic0.app/) is the decentralized autonomous organization (DAO) that governs the Internet Computer Protocol.
```
LaraAS, line 4: My preference is to not call the Frontend "the NNS" because others from the community might also build a frontend for the NNS. So I would add a sentence, "the on-chain frontend that allows you to interact with it is here" and link to the frontend there

docs/tokenomics/nns/nns-intro.md
```diff
@@ -1,218 +1,53 @@
...
+It is a fully on-chain, decentralized organization and is, for instance, responsible for making protocol-level upgrades to continuously improve the Internet Computer Protocol. 
 
-The NNS works by allowing users to stake ICP governance tokens to create voting neurons. Anyone can create a neuron, which together exert the will of the community, mediated through algorithms. Neurons are like accounts where notice of withdrawals must be given, where the length of the notice period is a configuration known as the “dissolve delay.” The voting power of neurons, and their relative claim to voting rewards, is proportional to the quantity of staked ICP, the length of the configured dissolve delay, and their “age.” Neurons can be made to vote manually, or automatically, by following other neurons in a form of liquid democracy.
+It operates under a liquid democracy in which those who have staked their ICP or [ICP neuron holders vote](../../concepts/governance.md) vote on proposals that shape the development of the Internet Computer.
```
LaraAS, line 8: I think
> those who have staked their ICP or [ICP neuron holders vote]

is confusing as it is not clear whether the "or" refers to two different concepts..
How about "those who have staked their ICP in a _neuron_ 

(and I think "vote" might be duplicated?)

docs/tokenomics/nns/nns-intro.md
```diff
@@ -1,218 +1,53 @@
...
+It operates under a liquid democracy in which those who have staked their ICP or [ICP neuron holders vote](../../concepts/governance.md) vote on proposals that shape the development of the Internet Computer. 
 
-Neuron holders are placed in a crypto-economic game, where they are incentivized to vote to adopt and reject proposals, or to configure neuron follows that cause them to vote automatically in a desirable way, according to what is most likely to drive the value of the Internet Computer protocol over the long term.
+Examples of proposals include:
+1. upgrade the protocol and software used by the node machines that host the network
+2. create new subnet blockchains to increase network capacity
+3. split subnets to divide their load
+4. configure economic parameters that control how much must be  paid by users for compute capacity
```
LaraAS, line 14:
```suggestion
4. configure economic parameters that determine the cost for compute capacity
```

docs/tokenomics/nns/nns-intro.md
```diff
@@ -1,218 +1,53 @@
...
+5. In extreme situations, it can freeze malicious canister smart contract software to protect the network; and many other things.
```
LaraAS, line 15: do we really want to point out this controversial topic? I would rather leave that out.

If you want one more once could mentioned the proposals that update the DAO itself (with voting rules etc)

docs/tokenomics/nns/nns-intro.md
```diff
@@ -1,218 +1,53 @@
...
+It operates under a liquid democracy in which those who have staked their ICP or [ICP neuron holders vote](../../concepts/governance.md) vote on proposals that shape the development of the Internet Computer. 
 
-Neuron holders are placed in a crypto-economic game, where they are incentivized to vote to adopt and reject proposals, or to configure neuron follows that cause them to vote automatically in a desirable way, according to what is most likely to drive the value of the Internet Computer protocol over the long term.
+Examples of proposals include:
+1. upgrade the protocol and software used by the node machines that host the network
+2. create new subnet blockchains to increase network capacity
+3. split subnets to divide their load
```
LaraAS, line 13: I think it is fine but we rarely (never) did this, so maybe not a super representative example

docs/tokenomics/nns/nns-intro.md
```diff
@@ -1,218 +1,53 @@
...
+The governance voting algorithm also applies to the "wait for quiet" period, which decides on proposals quickly if all voters agree, but increases the time that neurons can vote for more controversial proposals. Depending on the votes, the voting period can be dramatically increased.
```
LaraAS, line 22: I think this is a bit confusing. How about: 

```suggestion
For determining how long proposals are open for voting, the governance uses a so called "Wait-for-quiet" algorithm. This ensures that proposals are decided quickly if all voters agree, but have a longer voting period if they are controversial. Most proposals start with 4 days of voting period that can be extended up to 8 days.
```

docs/tokenomics/nns/nns-intro.md
```diff
@@ -1,218 +1,53 @@
...
+Once such a proposal is accepted, it is autonomously executed.
```
LaraAS, line 23:
```suggestion
Once a proposal is adopted, it is autonomously executed on-chain.
```

(note that this is not true for motion proposals so it would be more correct to say "for most" or "with the exception of..")

docs/tokenomics/nns/nns-intro.md
```diff
@@ -1,218 +1,53 @@
...
 
-The NNS decides whether to adopt or reject proposals by watching how neurons emit votes. Anyone can create a neuron by locking balances of the Internet Computer’s native utility token (ICP) that is hosted on a ledger inside the NNS. When a user creates a neuron, the locked balance of ICP can only be unlocked by disbursing (“destroying”) the neuron. Users are incentivized to create neurons because they earn rewards when they vote on proposals. The quantity of rewards disbursed to a neuron derive from such factors as the size of the locked balance, the minimum lockup period remaining (the “dissolve delay”), the neuron’s “age,” the proportion of possible votes it has participated in, and the sum of voting activity across all neurons, since the overall total rewards disbursed is capped and must be divided between voters.
+While other blockchains take weeks or months to upgrade (sometimes called hard fork) and typically require substantial manual work and coordination to do so, ICP upgrades itself every week (https://dashboard.internetcomputer.org/releases). Its ability to upgrade and iterate is a comparative "superpower."
```
LaraAS, line 25:
```suggestion
While other blockchains take weeks or months to upgrade and typically require substantial manual work and coordination to do so, ICP upgrades itself every week (https://dashboard.internetcomputer.org/releases). Its ability to upgrade and iterate is a comparative "superpower."
```

I would remove this: it is only a hard fork if a part of the network upgrades but the other part does not. So not all upgrades can be called that way I would say.

docs/tokenomics/nns/nns-intro.md
```diff
@@ -1,218 +1,53 @@
...
 
-Neuron owners may find it hard to manually vote directly on every proposal submitted to the NNS. Firstly, large volumes of proposals may be submitted to the NNS, often at awkward times, and owners may not be available or have the necessary time to evaluate each one. Secondly, neuron owners may lack the necessary expertise to evaluate proposals themselves. The NNS uses a form of liquid democracy to address these challenges. For any proposal topic, a neuron can be configured to vote automatically by following the votes of a group of neurons, voting to adopt proposals whenever more than half of the followees vote to adopt, and voting to reject whenever at least half of the followees vote to reject. A catch-all follow rule may also be defined to make a neuron vote automatically on proposals with topics for which no follow rule has been defined. It is assumed that neuron owners will manage how their neurons follow other neurons in the best interests of the network, which is also in their own economic interests, owing to their locked ICP balances.
+Proposals are adopted or rejected based on voting activity by “neurons” that network participants have created.
```
LaraAS, line 30: I wonder if neurons need to be introduced in 1 sentence before mentioning them here.
they appeared above but were not really "Defined" I think

docs/tokenomics/nns/nns-intro.md
```diff
@@ -1,218 +1,53 @@
...
 
-It is expected that a large proportion of the overall supply of ICP will be locked in order to earn rewards. This secures the Internet Computer’s network governance, by making it both difficult and exorbitantly expensive for an attacker to acquire a sufficiently large stake to gain significant influence. Since neuron owners may wish to maximize their rewards by voting on all proposals, most neurons will either be actively managed or configured to follow other neurons so that they can vote automatically.
+## Participation Rewards
+Network governance on the Internet Computer is incentivized through a system called "neurons".
+ICP token holders can create a “neuron” by locking balances of ICP up to eight years that are hosted on a ledger inside the NNS.
```
LaraAS, line 34:
```suggestion
ICP token holders can create a “neuron” by staking ICP up to eight years that are hosted on a ledger inside the NNS. 
```

docs/tokenomics/nns/nns-intro.md
```diff
@@ -1,218 +1,53 @@
...
+## Participation Rewards
+Network governance on the Internet Computer is incentivized through a system called "neurons".
+ICP token holders can create a “neuron” by locking balances of ICP up to eight years that are hosted on a ledger inside the NNS. 
 
-## Proposals
-### Format
-Each proposal submitted to the NNS has the following fields:
+When a user creates a neuron, the locked balance of ICP can only be unlocked by disbursing (“destroying”) the neuron. Each neuron is configured with “dissolve delay.” or setting that determines the minimum amount of time that you have to wait before you can unlock your original locked ICP utility tokens. Once a neuron has been placed into “dissolve mode,” its dissolve delay falls over time, rather like a kitchen timer, until it reaches zero, whereupon its owner can perform a final disburse action to unlock the balance of ICP.
```
LaraAS, line 36: Why is this explanation of neurons under the section "proposals"? (Not saying this can't work, but I didn't quite understand this structure)

```suggestion
When a user creates a neuron, the staked ICP can only be taken out of the neuron again by disbursing the neuron. Each neuron is configured with a “dissolve delay” that determines the minimum amount of time that you have to wait before you can unlock your staked ICP utility tokens. Once a neuron has been placed into “dissolving mode,” its dissolve delay falls over time, rather like a kitchen timer, until it reaches zero, whereupon its owner can perform a final disburse action to unlock the balance of ICP. 
```

docs/tokenomics/nns/nns-intro.md
```diff
@@ -1,218 +1,53 @@
...
-* **URL:** the web address of additional content required to evaluate the proposal, specified using HTTPS. For example, the address might describe content supporting the assignment of a DCID (data center id) to a new data center.
-* **Proposer:** the ID of the neuron that submitted the proposal. When a proposal is submitted, a “charge” is placed on its balance in case it is rejected. So the balance needs to be big enough to pay the charge on (all) rejection(s). We require a neuron to have a dissolve delay ≥ 6 months to vote, and this applies to submitting proposals too.
-* **Proposal Type:** the type of the proposal. This infers what topic it belongs to (e.g., #NodeAdmin), the system function that will process the proposal if it is adopted, and the type and structure of the parameters that will be passed to that function.
-* **Parameters:** the parameters that will be passed to the system function that will be invoked if the proposal is adopted, as determined by its type. When a proposal is submitted, the NNS checks the parameters.
+Neuron owners are economically incentivized to vote in the best interests of the network by locking their ICP tokens for longer periods.
```
LaraAS, line 38:
```suggestion
Neurons with a longer dissolve delay have more voting power and get more rewards so that they are economically incentivized to vote in the best long-term interests of the network.  
```

docs/tokenomics/nns/nns-intro.md
```diff
@@ -1,218 +1,53 @@
...
-* **Proposal Type:** the type of the proposal. This infers what topic it belongs to (e.g., #NodeAdmin), the system function that will process the proposal if it is adopted, and the type and structure of the parameters that will be passed to that function.
-* **Parameters:** the parameters that will be passed to the system function that will be invoked if the proposal is adopted, as determined by its type. When a proposal is submitted, the NNS checks the parameters.
+Neuron owners are economically incentivized to vote in the best interests of the network by locking their ICP tokens for longer periods.  
 
-The NNS assigns a unique identity to each proposal that it receives.
+The number of rewards disbursed to a neuron derives from such factors as the size of the locked balance, the minimum lockup period remaining (the “dissolve delay”), the neuron’s “age,” the proportion of possible votes it has participated in, and the sum of voting activity across all neurons, since the overall total rewards disbursed is capped and must be divided between voters.
```
LaraAS, line 40:
```suggestion
The number of rewards that neuron receives derive from factors as the amount of staked tokens, the lockup period remaining (the “dissolve delay”), the neuron’s “age”, the proportion of possible votes it has participated in, and the sum of voting activity across all neurons, since the overall total rewards are capped and must be divided between voters.
```

docs/tokenomics/nns/nns-intro.md
```diff
@@ -1,218 +1,53 @@
...
 
-### Topics
-The topic of a proposal, which is inferred from its type, determines how it will be processed. For example, the NNS may require voters to have a greater degree of agreement, or to try to process proposals faster, for some topics. Also, neurons follow other neurons on a per-topic basis. Initial topics include:
+It is expected that a large proportion of the overall supply of ICP will be locked to earn rewards to secure the Internet Computer’s network governance from attacker to acquire a sufficiently large stake to gain significant influence.
```
LaraAS, line 42: I am not sure if I would make this claim.
First, I think we shouldn't imply that by just staking you get rewards (you actually have to vote to get rewards). Second, I am not sure we want to say this is an expectation.
We could maybe just say that at the moment it is a reality (if this is true)? Or maybe omit this?

docs/tokenomics/nns/nns-intro.md
```diff
@@ -1,218 +1,53 @@
...
+Neuron owners face challenges in manually voting on every NNS proposal due to high submission volumes and a potential lack of expertise. The NNS allows automatic voting through delegation where neurons automatically vote by following others for all proposals or a specific proposal topic. This ensures efficient decision-making without requiring constant manual input.
```
LaraAS, line 49:
```suggestion
Neurons can manually vote on all proposals. 
If they cannot manually vote on all proposals due to high submission volumes or a lack of expertise, neurons can delegate their vote, called _following_, for specific proposal topic. This ensures efficient decision-making without requiring constant manual input.
```

docs/tokenomics/nns/nns-intro.md
```diff
@@ -1,218 +1,53 @@
...
 
-* **ManageNetworkEconomics (#NetworkEconomics)**: this is a single proposal type which can update one or several economic parameters:
+It is expected that a large proportion of the overall supply of ICP will be locked to earn rewards. This secures the Internet Computer’s network governance, by making it both difficult and exorbitantly expensive for an attacker to acquire a sufficiently large stake to gain significant influence. Since neuron owners may wish to maximize their rewards by voting on all proposals, most neurons will either be actively managed or configured to follow other neurons so that they can vote automatically.
```
LaraAS, line 51: I think this is a repetition of above. As already mentioned there I am not sure if we want to mention this.
Do you think this is important? What is the main message we want to deliver here?

docs/tokenomics/nns/proposal-requirements.md
```diff
@@ -0,0 +1,206 @@
...
+
+Only neurons with at least 10 ICP and 6 months of dissolve delay can submit proposals to the NNS. To learn more on how to submit a proposal, refer to the [How to Create an NNS motion proposal](https://wiki.internetcomputer.org/wiki/How-To:_Create_an_NNS_motion_proposal). 
+
+### Proposal Format
+
+Each proposal submitted to the NNS has the following fields:
+
+* **Summary:** text providing a short description of the proposal, composed using a maximum of 280 bytes.
+* **URL:** the web address of additional content required to evaluate the proposal, specified using HTTPS. For example, the address might describe content supporting the assignment of a DCID (data center id) to a new data center.
+* **Proposer:** the ID of the neuron that submitted the proposal. When a proposal is submitted, a “charge” is placed on its balance in case it is rejected. So the balance needs to be big enough to pay the charge on (all) rejection(s). We require a neuron to have a dissolve delay of ≥ 6 months to vote, and this applies to submitting proposals too.
```
LaraAS, line 15:
```suggestion
* **Proposer:** the ID of the neuron that submitted the proposal. When a proposal is submitted, a “charge” is placed on its balance in case it is rejected. So the balance needs to be big enough to pay the charge on (all) rejection(s). The NNS requires a neuron to have a dissolve delay of ≥ 6 months to vote, and this applies to submitting proposals too.
```

docs/tokenomics/nns/proposal-requirements.md
```diff
@@ -0,0 +1,206 @@
...
+* **Proposal Type:** the type of the proposal. This infers what topic it belongs to (e.g., #NodeAdmin), the system function that will process the proposal if it is adopted, and the type and structure of the parameters that will be passed to that function.
+* **Parameters:** the parameters that will be passed to the system function that will be invoked if the proposal is adopted, as determined by its type. When a proposal is submitted, the NNS checks the parameters.
```
LaraAS, line 17:
```suggestion
* **Parameters:** the parameters that will be passed to the system function that will be invoked if the proposal is adopted, as determined by its type. 
```
(not sure if this is true for all proposals)

docs/tokenomics/nns/proposal-requirements.md
```diff
@@ -0,0 +1,206 @@
...
+
+The NNS assigns a unique identity to each proposal that it receives.
+
+### Topics
+
+Each type of proposal belongs to a specific “proposal topic,” such as “#NodeAdmin” or “#NetworkEconomics,” which determines details about how it will be processed. To prevent users (neurons) from spamming the NNS with proposals, a fee is levied on the neuron that submitted a proposal if it is rejected.
```
LaraAS, line 23:
```suggestion
Each type of proposal belongs to a specific “proposal topic,” such as “#NodeAdmin” or “#NetworkEconomics,” which determines details about how it will be processed. 
```
This was already mentioned and I think is not relevant for the "topic"-section

docs/tokenomics/nns/proposal-requirements.md
```diff
@@ -0,0 +1,206 @@
...
+Proposals are adopted or rejected based on voting activity by “neurons” that network participants have created.
```
LaraAS, line 24: This was also already mentioned, so wondering if it is needed here again. I would separate the explanation how proposals work and then just waht the topics are here

docs/tokenomics/nns/proposal-requirements.md
```diff
@@ -0,0 +1,206 @@
...
+
+For example, the NNS may require voters to have a greater degree of agreement or to try to process proposals faster, for some topics. Also, neurons follow other neurons on a per-topic basis. Initial topics include:
```
LaraAS, line 26: while this is true in theory, right now all proposals require the same thresholds (except for the very weird manage neuron proposals that standard users never interact with). So wondering if it  is needed to say this here

docs/tokenomics/nns/proposal-requirements.md
```diff
@@ -0,0 +1,206 @@
...
+
+* **#NeuronManagement:** a special topic using which a neuron can be managed by the followees for this topic (in this case, there is no fallback to default). Votes on this topic are not included in the voting history of the neuron. For proposals on this topic, only the neuron’s followers on the topic that the proposals pertain to are allowed to vote. Because the set of eligible voters for proposals on this topic is restricted, proposals on this topic have a shorter than normal voting period.
```
LaraAS, line 28:
```suggestion
* **#NeuronManagement:** a special topic with which a neuron can be managed by the followees for this topic. Votes on this topic are not included in the voting history of the neuron. For proposals on this topic, only the neuron’s followers on the topic that the proposals pertain to are allowed to vote. Because the set of eligible voters for proposals on this topic is restricted, proposals on this topic have a shorter than normal voting period.
```
How about putting this to the very end as this is irrelevant to most users? 
I didn't understand "(in this case, there is no fallback to default)" - so suggest to remove it

docs/tokenomics/nns/proposal-requirements.md
```diff
@@ -0,0 +1,206 @@
...
+* **#ExchangeRate:** all proposals provide information in “real-time” about the market value of ICP, as measured by an International Monetary Fund (IMF) Special Drawing Right (SDR), which allows the NNS to convert ICP to cycles (which power computation) at a rate that keeps their real-world cost constant. Because proposals on this topic are very frequent, they have a shorter voting period, and votes on this topic are not included in the voting history of the neuron.
```
LaraAS, line 29: This is not used anymore. The exchange rate is now on the CMC. I would remove this (Even though I think we have not yet cleaned up all code these proposals should not exist anymore)

docs/tokenomics/nns/proposal-requirements.md
```diff
@@ -0,0 +1,206 @@
...
+* **#NetworkEconomics:** proposals that administer network economics — for example, determining what rewards should be paid to node operators.
+* **#Governance:** all proposals that administer governance — for example, motions and the configuration of certain parameters.
+* **#NodeAdmin:** all proposals that administer node machines somehow, including but not limited to upgrading or configuring the OS, upgrading or configuring the virtual machine framework, and upgrading or configuring the node replica software.
```
LaraAS, line 32:
```suggestion
* **#NodeAdmin:** all proposals that administer node machines, including but not limited to upgrading or configuring the OS, upgrading or configuring the virtual machine framework, and upgrading or configuring the node replica software.
```

docs/tokenomics/nns/proposal-requirements.md
```diff
@@ -0,0 +1,206 @@
...
+* **#ParticipantManagement:** all proposals that administer network participants — for example, granting and revoking DCIDs (data center identities) or NPIDs (node provider identities).
+* **#SubnetManagement:** all proposals that administer network subnets — for example, creating new subnets, adding and removing subnet nodes, and splitting subnets.
+* **#NetworkCanisterManagement:** installing and upgrading “system” [[NNS Canisters|canisters]] that belong to the network — for example, upgrading the NNS.
+* **#KYC:** proposals that update KYC information for regulatory purposes — for example, during the initial Genesis distribution of ICP in the form of neurons.
```
LaraAS, line 36: if we don't list all topics I would leave this out. It is not relevant anymore

docs/tokenomics/nns/proposal-requirements.md
```diff
@@ -0,0 +1,206 @@
...
+* **#NodeProviderRewards:** topic for proposals to reward node providers.
```
LaraAS, line 37: not sure if this is still used. another candidate to remove.
Maybe instead one coud mention SNS proposals that are used more frequently?

docs/tokenomics/nns/proposal-requirements.md
```diff
@@ -0,0 +1,206 @@
...
...ex and time-consuming). To address this using this proposal type, the important neuron can be configured to follow the neurons controlled by individual members of a team. Now they can submit proposals to make the important neuron perform actions, which are adopted if and only if a majority of them vote to adopt. (Submitting such a proposal costs a small fee, to prevent denial-of-service attacks.) Nearly any command on the target neuron can be executed, including commands that change the following rules, allowing the set of team members to be dynamic. Only the final step of dissolving the neuron once its dissolve delay reaches zero cannot be performed using this type of proposal, since this would allow control/“ownership” over the locked balances to be transferred. (The only exception to this rule applies to not-for-profit organizations, which may be allowed to dissolve their neurons without using the initial private key.) To prevent a neuron from falling under the malign control of the principal’s private key by accident, the private key can be destroyed so that the neuron can only be controlled by its followers, although this makes it impossible to subsequently unlock the balance.
```
LaraAS, line 43:
```suggestion
* **ManageNeuron (#NeuronManagement, Restricted Voting)**: this type of proposal calls a function on a specified target neuron. Only the followers of the target neuron may vote on these proposals, which effectively provides the followees with control over the target neuron. This can provide convenient means for a team of individuals to manage an important neuron. For example, a neuron might hold a large stake, or belong to an organization of high repute, and be publicized so that many other neurons can follow its vote. In both cases, managing the private key of the principal securely could be problematic. (Either a single copy is held, which is very insecure and provides for a single party to take control, or a group of individuals must divide responsibility — for example, using threshold cryptography, which is complex and time-consuming). To address this using this proposal type, the important neuron can be configured to follow the neurons controlled by individual members of a team. Now they can submit proposals to make the important neuron perform actions, which are adopted if and only if a majority of them vote to adopt. (Submitting such a proposal costs a small fee, to prevent denial-of-service attacks.) Nearly any command on the target neuron can be executed, including commands that change the following rules, allowing the set of team members to be dynamic. Only the final step of dissolving the neuron once its dissolve delay reaches zero cannot be performed using this type of proposal, since this would allow control/“ownership” over the locked balances to be transferred. (The only exception to this rule applies to not-for-profit organizations, which may be allowed to dissolve their neurons without using the initial private key.) To prevent a neuron from falling under the malign control of the principal’s private key by accident, the private key can be destroyed so that the neuron can only be controlled by its followers, although this makes it impossible to subsequently unlock the balance.
```

I think this is all true but I wonder how relevant this is. If we think this is needed for develeopers I wouuls still put this to the very end. 
I think there is no way to create a neuron that is not-for-profit. Only some neurons at genesis had this field. If we want devs to use this they might need to know this. But I wonder if we should recommend that

docs/tokenomics/nns/proposal-requirements.md
```diff
@@ -0,0 +1,206 @@
...
+
+* **Facilitating network governance**: ICP tokens can be locked to create neurons that participate in network governance by voting, through which they can earn economic rewards.
+
+* **Production of cycles for compute**: ICP provides a source store of value that can be converted into “cycles,” which power computation in the role of fuel that is burned when it is used. The NNS converts ICP to cycles at a variable rate, so chosen to ensure users of the network can always create new cycles at approximately constant cost concerning IMF SDR, such that the cost of acquiring fuel is predictable.
+
+* **Rewarding participants**: the network provides rewards to incentivize those playing important roles that enable the network to function, including a) the provision of “voting rewards” to those participating in governance, and b) the provision of “node provider rewards” to those operating the node machines that are hosting the network.
+
+## Ledger
+The ICP ledger is hosted within the NNS and records all balances of ICP in the manner of a spreadsheet. Each row is called an “account,” which has two fields (i.e., there are two “columns”):
```
LaraAS, line 120:
```suggestion
The ICP ledger is hosted within the NNS and records all balances of ICP in the manner similar to a spreadsheet. Each row is called an “account,” which has two fields (i.e., there are two “columns”):
```

docs/tokenomics/nns/proposal-requirements.md
```diff
@@ -0,0 +1,206 @@
...
+
+* **Account identifier (bytes)**: a unique value that is derived from the identity of the “principal” that “controls” the account. Currently, the principal must either be (i) the owner of a public key pair, or (ii) a canister smart contract that is part of the NNS. Account identifiers are derived by hashing the concatenation of a domain separator, the principal ID, and the subaccount (or zeros if no subaccount is given).
+
+* **Balance** (positive integer, representing one-hundredth of a millionth of an ICP): the quantity of ICP assigned to the principal of the account.
+
+When the principal is a public key or Canister, they can apply the following operation to an account:
```
LaraAS, line 126:
```suggestion
When the principal is a public key or Canister, they can apply the following operation to an account:
```
Is "Canister" capitalized? Deferring to @jessiemongeon1 here :)

docs/tokenomics/nns/proposal-requirements.md
```diff
@@ -0,0 +1,206 @@
...
+
+* **Send**: send a portion of the ICP balance to another account. If all the ICP is sent to another account, then the sending account ceases to exist (i.e., is deleted from the ledger).
+* **Notify**: when the destination of the funds sent is the account of an NNS canister (e.g., an account of the governance canister), the sender can ask the ledger to notify the recipient canister of the incoming transfer. The recipient canister can then act on this notification. Two examples where this ability is used are creating a neuron and refreshing the stake of a neuron. These are detailed below.
```
LaraAS, line 129: Is this still usable? Let's make sure with @MarioDfinity . This is not anymore how neurons are created

docs/tokenomics/nns/proposal-requirements.md
```diff
@@ -0,0 +1,206 @@
...
+
+Operations that require interaction between the ledger and the governance system (Neurons):
+
+* **Create neuron**: when the principal is a public key holder, they may lock a portion of their balance inside a new neuron. Technically, creating the neuron is done in two stages. First, transfer the ICP to be staked to an account of the governance canister (which corresponds to a new neuron — the details of the association are omitted here). Then notify the governance canister of the incoming transfer which updates its internal neuron bookkeeping. If the entire balance is locked inside a new neuron, the account ceases to exist (i.e., is deleted from the ledger). To move these ICP to a different account, such as back to the original account, where they can once again be controlled like a normal balance, the associated neuron must be fully dissolved and disbursed (destroyed). The new neuron that has been created is controlled by the private key of the principal that created it.
```
LaraAS, line 133:
```suggestion
* **Create neuron**: when the principal is a public key holder, they may lock a portion of their balance inside a new neuron. Technically, creating the neuron is done in two stages. First, transfer the ICP to be staked to an account of the governance canister (which corresponds to a new neuron — the details of the association are omitted here). Then notify the governance canister of the incoming transfer which updates its internal neuron bookkeeping. To move these ICP to a different account, such as back to the original account, where they can once again be controlled like a normal balance, the associated neuron must be fully dissolved and disbursed (destroyed). 
```

" If the entire balance is locked inside a new neuron, the account ceases to exist (i.e., is deleted from the ledger)." is not fully correct: the tokens will still be on the ledger, they are just in an account that is owned by governance and thus cannot be touched by the user

"The new neuron that has been created is controlled by the private key of the principal that created it." is not fully true: I can claim a neuron for someone else. It just depends which principal is used to compute the subaccount. Maybe this is too technical to exampl here?

docs/tokenomics/nns/proposal-requirements.md
```diff
@@ -0,0 +1,206 @@
...
+* **Refresh stake**: the stake of a neuron may be increased by transferring to its address/account in the ledger and notifying the governance canister of the incoming transfer. Refreshing the stake will change the maturity and age of the neuron prorated. For example, if the stake is doubled, the maturity and age will be halved, so spawning will yield the same amount and the age bonus will be the same as before (in absolute terms).
```
LaraAS, line 134: Is it true that the maturity is also affected? @bjoernek maybe you know but I would find that surprising

docs/tokenomics/nns/proposal-requirements.md
MarioDfinity: No. `notify` is deprecated and may be removed in future. It should not be mentioned in the documentation.

docs/tokenomics/nns/proposal-requirements.md
```diff
@@ -0,0 +1,206 @@
...
+
+## Neurons
+
+A neuron locks a balance of ICP and enables its owner to participate in network governance, through which they can earn rewards.
+
+### Attributes
+Neurons have the following attributes:
+
+* **Identity (uint64)**: the general identity of the neuron object. When a neuron is configured to follow another neuron, this is the value that is used. This is a random 64-bit value selected at neuron creation.
```
LaraAS, line 143:
```suggestion
* **Identity (uint64)**: the general identity of the neuron object. When a neuron is configured to follow another neuron, this is the value that is used. This is a random 64-bit value selected by NNS governance at neuron creation.
```

docs/tokenomics/nns/proposal-requirements.md
```diff
@@ -0,0 +1,206 @@
...
+
+* **Maturity (positive number – percent)**: the maturity of a neuron, which determines its ability to spawn a new neuron and corresponding locked balance of newly minted ICP in expectation equal to this value in percent of the stake of the spawning neuron. When new neurons are created, their maturity is zero. When neurons vote, over time the NNS increases their maturity to reward them.
+
+* **Follow relationships (mapping from topic to list of followers, private)**: a neuron can be configured to vote automatically by following other neurons on a topic-by-topic basis. For any valid topic, a list of followers can be specified, and the neuron will follow the vote of a majority of the followers on a proposal with a type belonging to that topic. If a null topic is specified, this acts as a catch-all that enables the neuron to follow the vote of followees where a rule has not been specified.
```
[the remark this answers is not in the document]
LaraAS: We could mention here that a few topics are not included in teh "catch-all"

docs/tokenomics/nns/proposal-requirements.md
```diff
@@ -0,0 +1,206 @@
...
+
+* **Rewarding participants**: the network provides rewards to incentivize those playing important roles that enable the network to function, including a) the provision of “voting rewards” to those participating in governance, and b) the provision of “node provider rewards” to those operating the node machines that are hosting the network.
+
+## Ledger
+The ICP ledger is hosted within the NNS and records all balances of ICP in the manner of a spreadsheet. Each row is called an “account,” which has two fields (i.e., there are two “columns”):
+
+* **Account identifier (bytes)**: a unique value that is derived from the identity of the “principal” that “controls” the account. Currently, the principal must either be (i) the owner of a public key pair, or (ii) a canister smart contract that is part of the NNS. Account identifiers are derived by hashing the concatenation of a domain separator, the principal ID, and the subaccount (or zeros if no subaccount is given).
```
MarioDfinity, line 122: 1. The controller (or owner) of the account is any principal. A canister that is not part of the NNS is still a valid owner.
2. Is there a reason why we do the distinction between principals associated with a key pair and the principal of a canister? Both are principals from the Ledger point of view.
3. Do we explain what the subaccount is? It is used but not introduced before.

docs/tokenomics/nns/proposal-requirements.md
```diff
@@ -0,0 +1,206 @@
...
+
+* **NotForProfit (boolean)**: whether this neuron is “not for profit,” making it dissolvable by voting.
+
+The following attributes can be computed:
+
+* **Age (seconds)** (computed from AgingSince and current time): the period that has elapsed since the neuron was created or last stopped dissolving. Conceptually, whenever a neuron starts dissolving, then its age is reset to zero and remains zero while it is dissolving. If a dissolving neuron has been turned off, the current time becomes the effective neuron creation date to calculate the age.
+* **State (LOCKED or DISSOLVING or DISSOLVED)** (computed from DissolveState and the current time):
+
+    - LOCKED: In this state, the neuron is locked with a specific DissolveDelay. It accrues age by the passage of time and it can vote if DissolveDelay is at least six months. The method `start_dissolving` can be called to transfer the neuron to the DISSOLVING state. The method `increase_dissolve_delay` can be used to increase the dissolve delay without affecting the state or the age of the neuron.
```
LaraAS, line 174:
```suggestion
    - 
    NOT_DISSOLVING: In this state, the neuron is locked with a specific DissolveDelay. It accrues age by the passage of time and it can vote if DissolveDelay is at least six months. The method `start_dissolving` can be called to transfer the neuron to the DISSOLVING state. The method `increase_dissolve_delay` can be used to increase the dissolve delay without affecting the state or the age of the neuron.
```
This should be applied everywhere: this is now called NOT_DISSOLVING in the code and "in English" we call this non-dissolving sometimes, but let's try to not use "locked" as this might be confusing

docs/tokenomics/nns/proposal-requirements.md
```diff
@@ -0,0 +1,206 @@
...
+
+* **Rewarding participants**: the network provides rewards to incentivize those playing important roles that enable the network to function, including a) the provision of “voting rewards” to those participating in governance, and b) the provision of “node provider rewards” to those operating the node machines that are hosting the network.
+
+## Ledger
+The ICP ledger is hosted within the NNS and records all balances of ICP in the manner of a spreadsheet. Each row is called an “account,” which has two fields (i.e., there are two “columns”):
+
+* **Account identifier (bytes)**: a unique value that is derived from the identity of the “principal” that “controls” the account. Currently, the principal must either be (i) the owner of a public key pair, or (ii) a canister smart contract that is part of the NNS. Account identifiers are derived by hashing the concatenation of a domain separator, the principal ID, and the subaccount (or zeros if no subaccount is given).
+
+* **Balance** (positive integer, representing one-hundredth of a millionth of an ICP): the quantity of ICP assigned to the principal of the account.
```
MarioDfinity, line 124: The balance is the quantity of ICP assigned to an account, not to a principal of an account. The quantity of ICP assigned to the principal of the account is the sum of the balances of all accounts owned by that principal.

docs/tokenomics/nns/proposal-requirements.md
```diff
@@ -0,0 +1,206 @@
...
...ay only be voted upon by the neuron’s followers. This provides a foundation for the management of highly security-sensitive neurons since it allows them to be maintained without hot keys or the secret key of the principal, which can be kept in cold storage or even destroyed (so long as the associated balance of ICP needs never be unlocked). For example, the Internet Computer Association or other known neurons might publicize the address of special neurons that will be made to vote according to their wishes, so that others can configure their neurons to follow them and leverage their expertise and efforts in governance. One problem with such practices is that they introduce the risk that secret keys used in the management of the publicized neurons might be compromised, allowing hackers to take control and “trick” large numbers of following neurons into voting according to their wishes. If the publicized neurons have admin proposals enabled, however, then they can be administered by the neurons they follow (their followers), which are typically controlled by a large number of team members who cannot be simultaneously extorted, without any need for the usage of secret keys whatsoever.
```
LaraAS, line 178: I don't know this field. Would remove this. There are other neuron fields. Did you want to list all of them here? 
They are defined here https://sourcegraph.com/github.com/dfinity/ic/-/blob/rs/nns/governance/proto/ic_nns_governance/pb/v1/governance.proto?L226 but I wonder if it makes sense to just list the most important ones here and refer to the file which will always be up to date.

docs/tokenomics/nns/proposal-requirements.md
```diff
@@ -0,0 +1,206 @@
...
+## Ledger
+The ICP ledger is hosted within the NNS and records all balances of ICP in the manner of a spreadsheet. Each row is called an “account,” which has two fields (i.e., there are two “columns”):
+
+* **Account identifier (bytes)**: a unique value that is derived from the identity of the “principal” that “controls” the account. Currently, the principal must either be (i) the owner of a public key pair, or (ii) a canister smart contract that is part of the NNS. Account identifiers are derived by hashing the concatenation of a domain separator, the principal ID, and the subaccount (or zeros if no subaccount is given).
+
+* **Balance** (positive integer, representing one-hundredth of a millionth of an ICP): the quantity of ICP assigned to the principal of the account.
+
+When the principal is a public key or Canister, they can apply the following operation to an account:
+
+* **Send**: send a portion of the ICP balance to another account. If all the ICP is sent to another account, then the sending account ceases to exist (i.e., is deleted from the ledger).
```
MarioDfinity, line 128: Can we use transfer and not send? Transfer is the term used in [ICRC-1](https://github.com/dfinity/ICRC-1/blob/main/standards/ICRC-1/README.md#icrc1_transfer-) which itself is derived from the [same function on the Ethereum network](https://eips.ethereum.org/EIPS/eip-20#transfer).

`send` on the ICP Ledger is deprecated in favour of `transfer`, `icrc1_transfer` and `icrc2_transfer_from`. It should be used only by legacy code.

docs/tokenomics/nns/proposal-requirements.md
```diff
@@ -0,0 +1,206 @@
...
+
+### Commands
```
LaraAS, line 180: there are not all neuron commands. I think it makes sense to just mention the most important ones here, but maybe it would be good to explicitly say so? 
One could link to the github link corresponding to here https://sourcegraph.com/github.com/dfinity/ic/-/blob/rs/nns/governance/proto/ic_nns_governance/pb/v1/governance.proto?L829 where they are all defined.

docs/tokenomics/nns/nns-intro.md
jennifertrin: That's a good point. I will also note to provide more documentation on how to interact with the governance canister in the future :)

docs/tokenomics/nns/nns-intro.md
jennifertrin: Good catch! Yes, I think that your suggestion is more clear!

docs/tokenomics/nns/nns-intro.md
jennifertrin: I think we were aiming to convey this idea:

"Since the incentive to stake ICP in a neuron is high, it is difficult and exorbitantly expensive for an attacker to acquire a sufficiently large stake to gain significant influence."

Does this sound better?

docs/tokenomics/nns/proposal-requirements.md
jennifertrin: Hmm, maybe we can leave it out for now?

docs/tokenomics/nns/proposal-requirements.md
jennifertrin: Good point!

docs/tokenomics/nns/proposal-requirements.md
jennifertrin: I think transfer and send are pretty much used interchangeably on Ethereum. I'll update `transfer` as it aligns with the function name in the code. Thank you!

docs/tokenomics/nns/nns-intro.md
jennifertrin: I thnk a better flow is:

- Participation Rewards
- Proposals
- Participation

Thank you for the suggestion!

docs/tokenomics/nns/nns-intro.md
jennifertrin: Thank you, I moved it under Participation Rewards!

docs/tokenomics/nns/proposal-requirements.md
```diff
@@ -0,0 +1,206 @@
+# Submitting a Proposal
```
jessiemongeon1, line 1:
```suggestion
# Submitting a proposal
```

docs/tokenomics/nns/proposal-requirements.md
```diff
@@ -0,0 +1,206 @@
...
+
+## Overview 
+
+### Submit a Proposal
```
jessiemongeon1, line 5:
```suggestion

```

docs/tokenomics/nns/proposal-requirements.md
```diff
@@ -0,0 +1,206 @@
...
+
+Only neurons with at least 10 ICP and 6 months of dissolve delay can submit proposals to the NNS. To learn more on how to submit a proposal, refer to the [How to Create an NNS motion proposal](https://wiki.internetcomputer.org/wiki/How-To:_Create_an_NNS_motion_proposal).
```
jessiemongeon1, line 7:
```suggestion
Only neurons with at least 10 ICP and 6 months of dissolve delay can submit proposals to the NNS. To learn more on how to submit a proposal, refer to the [how to create an NNS motion proposal](https://wiki.internetcomputer.org/wiki/How-To:_Create_an_NNS_motion_proposal). 
```

docs/tokenomics/nns/proposal-requirements.md
```diff
@@ -0,0 +1,206 @@
+# Submitting a Proposal
+
+## Overview 
+
+### Submit a Proposal
+
+Only neurons with at least 10 ICP and 6 months of dissolve delay can submit proposals to the NNS. To learn more on how to submit a proposal, refer to the [How to Create an NNS motion proposal](https://wiki.internetcomputer.org/wiki/How-To:_Create_an_NNS_motion_proposal). 
+
+### Proposal Format
```
jessiemongeon1, line 9:
```suggestion
## Proposal format
```

docs/tokenomics/nns/proposal-requirements.md
```diff
@@ -0,0 +1,206 @@
...
+
+Each proposal submitted to the NNS has the following fields:
+
+* **Summary:** text providing a short description of the proposal, composed using a maximum of 280 bytes.
+* **URL:** the web address of additional content required to evaluate the proposal, specified using HTTPS. For example, the address might describe content supporting the assignment of a DCID (data center id) to a new data center.
+* **Proposer:** the ID of the neuron that submitted the proposal. When a proposal is submitted, a “charge” is placed on its balance in case it is rejected. So the balance needs to be big enough to pay the charge on (all) rejection(s). The NNS requires a neuron to have a dissolve delay of ≥ 6 months to vote, and this applies to submitting proposals too.
```
jessiemongeon1, line 15:
```suggestion
* **Proposer:** the ID of the neuron that submitted the proposal. When a proposal is submitted, a “charge” is placed on its balance in case it is rejected. The balance needs to be enough to pay the charge on (all) rejection(s). The NNS requires a neuron to have a dissolve delay of ≥ 6 months to vote, and which also applies to submitting proposals.
```

docs/tokenomics/nns/proposal-requirements.md
```diff
@@ -0,0 +1,206 @@
...
+* **Proposal Type:** the type of the proposal. This infers what topic it belongs to (e.g., #NodeAdmin), the system function that will process the proposal if it is adopted, and the type and structure of the parameters that will be passed to that function.
+* **Parameters:** the parameters that will be passed to the system function that will be invoked if the proposal is adopted, as determined by its type. When a proposal is submitted, the NNS checks the parameters.
```
jessiemongeon1, line 17:
```suggestion
* **Parameters:** the parameters that will be passed to the system function and invoked if the proposal is adopted, as determined by its type. When a proposal is submitted, the NNS checks these parameters.
```

docs/tokenomics/nns/proposal-requirements.md
```diff
@@ -0,0 +1,206 @@
...
+
+The NNS assigns a unique identity to each proposal that it receives.
+
+### Topics
```
jessiemongeon1, line 21:
```suggestion
## Proposal topics
```

docs/tokenomics/nns/proposal-requirements.md
```diff
@@ -0,0 +1,206 @@
...
+
+Each type of proposal belongs to a specific “proposal topic,” such as “#NodeAdmin” or “#NetworkEconomics,” which determines details about how it will be processed. To prevent users (neurons) from spamming the NNS with proposals, a fee is levied on the neuron that submitted a proposal if it is rejected.
+Proposals are adopted or rejected based on voting activity by “neurons” that network participants have created.
```
jessiemongeon1, line 24:
```suggestion
Proposals are adopted or rejected based on voting activity by neurons that network participants have created.
```

docs/tokenomics/nns/proposal-requirements.md
jessiemongeon1, line 7: Should we first include a definition of what a neuron is?

docs/tokenomics/nns/proposal-requirements.md
```diff
@@ -0,0 +1,206 @@
...
+
+For example, the NNS may require voters to have a greater degree of agreement or to try to process proposals faster, for some topics. Also, neurons follow other neurons on a per-topic basis. Initial topics include:
+
+* **#NeuronManagement:** a special topic using which a neuron can be managed by the followees for this topic (in this case, there is no fallback to default). Votes on this topic are not included in the voting history of the neuron. For proposals on this topic, only the neuron’s followers on the topic that the proposals pertain to are allowed to vote. Because the set of eligible voters for proposals on this topic is restricted, proposals on this topic have a shorter than normal voting period.
+* **#ExchangeRate:** all proposals provide information in “real-time” about the market value of ICP, as measured by an International Monetary Fund (IMF) Special Drawing Right (SDR), which allows the NNS to convert ICP to cycles (which power computation) at a rate that keeps their real-world cost constant. Because proposals on this topic are very frequent, they have a shorter voting period, and votes on this topic are not included in the voting history of the neuron.
```
jessiemongeon1, line 29:
```suggestion
* **#ExchangeRate:** all proposals provide information in real-time about the market value of ICP, as measured by an International Monetary Fund (IMF) Special Drawing Right (SDR), which allows the NNS to convert ICP to cycles (which power computation) at a rate that keeps their real-world cost constant. Because proposals on this topic are very frequent, they have a shorter voting period and votes on this topic are not included in the voting history of the neuron.
```

docs/tokenomics/nns/proposal-requirements.md
```diff
@@ -0,0 +1,206 @@
...
+* **#NetworkEconomics:** proposals that administer network economics — for example, determining what rewards should be paid to node operators.
```
jessiemongeon1, line 30:
```suggestion
* **#NetworkEconomics:** proposals that administer network economics; for example, determining what rewards should be paid to node operators.
```

docs/tokenomics/nns/proposal-requirements.md
```diff
@@ -0,0 +1,206 @@
...
+* **#Governance:** all proposals that administer governance — for example, motions and the configuration of certain parameters.
```
jessiemongeon1, line 31:
```suggestion
* **#Governance:** all proposals that administer governance; for example, motions and the configuration of certain parameters.
```

docs/tokenomics/nns/proposal-requirements.md
```diff
@@ -0,0 +1,206 @@
...
+* **#NodeAdmin:** all proposals that administer node machines somehow, including but not limited to upgrading or configuring the OS, upgrading or configuring the virtual machine framework, and upgrading or configuring the node replica software.
```
jessiemongeon1, line 32:
```suggestion
* **#NodeAdmin:** all proposals that administer node machines, including but not limited to upgrading or configuring the OS, the virtual machine framework, or the node replica software.
```

docs/tokenomics/nns/proposal-requirements.md
```diff
@@ -0,0 +1,206 @@
...
+* **#ParticipantManagement:** all proposals that administer network participants — for example, granting and revoking DCIDs (data center identities) or NPIDs (node provider identities).
```
jessiemongeon1, line 33:
```suggestion
* **#ParticipantManagement:** all proposals that administer network participants; for example, granting and revoking DCIDs (data center identities) or NPIDs (node provider identities).
```

docs/tokenomics/nns/proposal-requirements.md
```diff
@@ -0,0 +1,206 @@
...
+* **#SubnetManagement:** all proposals that administer network subnets — for example, creating new subnets, adding and removing subnet nodes, and splitting subnets.
```
jessiemongeon1, line 34:
```suggestion
* **#SubnetManagement:** all proposals that administer network subnets; for example, creating new subnets, adding and removing subnet nodes, and splitting subnets.
```

docs/tokenomics/nns/proposal-requirements.md
```diff
@@ -0,0 +1,206 @@
...
+* **#NetworkCanisterManagement:** installing and upgrading “system” [[NNS Canisters|canisters]] that belong to the network — for example, upgrading the NNS.
```
jessiemongeon1, line 35:
```suggestion
* **#NetworkCanisterManagement:** installing and upgrading system [[NNS Canisters|canisters]] that belong to the network; for example, upgrading the NNS.
```

docs/tokenomics/nns/proposal-requirements.md
```diff
@@ -0,0 +1,206 @@
...
+* **#KYC:** proposals that update KYC information for regulatory purposes — for example, during the initial Genesis distribution of ICP in the form of neurons.
```
jessiemongeon1, line 36:
```suggestion
* **#KYC:** proposals that update KYC information for regulatory purposes; for example, during the initial Genesis distribution of ICP in the form of neurons.
```

docs/tokenomics/nns/proposal-requirements.md
```diff
@@ -0,0 +1,206 @@
...
+* **#NodeProviderRewards:** topic for proposals to reward node providers.
+
+### Types
```
jessiemongeon1, line 39:
```suggestion
## Proposal types
```

docs/tokenomics/nns/proposal-requirements.md
jessiemongeon1: No, canister should be lowercase.

docs/tokenomics/nns/proposal-requirements.md
```diff
@@ -0,0 +1,206 @@
...
+
+* **ManageNetworkEconomics (#NetworkEconomics)**: this is a single proposal type that can update one or several economic parameters:
+
+    * **Reject cost:** the amount of ICP the proposer of a rejected proposal will be charged — to prevent the spamming of frivolous proposals.
```
jessiemongeon1, line 47:
```suggestion
    * **Reject cost:** the amount of ICP the proposer of a rejected proposal will be charged to prevent the spamming of frivolous proposals.
```

docs/tokenomics/nns/proposal-requirements.md
```diff
@@ -0,0 +1,206 @@
...
+* **Refresh stake**: the stake of a neuron may be increased by transferring to its address/account in the ledger and notifying the governance canister of the incoming transfer. Refreshing the stake will change the maturity and age of the neuron prorated. For example, if the stake is doubled, the maturity and age will be halved, so spawning will yield the same amount and the age bonus will be the same as before (in absolute terms).
+
+## Neurons
```
jessiemongeon1, line 136: This topic should be mentioned first to define Neurons and give context for their role in the proposals & voting

jessiemongeon1 (approved):

docs/tokenomics/nns/proposal-requirements.md
jennifertrin: We decided to remove KYC. Thank you!

docs/tokenomics/nns/proposal-requirements.md
jennifertrin: Removed it for now, will follow-up with Bjoern!

docs/tokenomics/nns/proposal-requirements.md
jennifertrin: I think it is noted that a canister can vote as well. I think it might be too much to introduce subaccount to start

PR #2464, opened 2024-02-06
Add landing pages for Azle/Kybra

jessiemongeon1:


jennifertrin (approved):

jennifertrin:
I assume you are getting the build error because the Rewards API does not return a `samples` section. I made a quick fix here: https://github.com/dfinity/portal/pull/2466 but we should probably follow up with @olaszakos

jessiemongeon1:
> I assume you are getting the build error because the Rewards API does not return a `samples` section. I made a quick fix here: #2466 but we should probably follow up with @olaszakos

This was fixed in this PR: https://github.com/dfinity/portal/pull/2467

PR #2473, opened 2024-02-07
Dev Tools Overview / Gitpod / CDKs

jessiemongeon1:


dfx-json (approved):
I have two pieces of feedback:

- I think we should swap the order of VSCode and Gitpod in the sidebars
- it might be helpful to mention that other IDEs can be used as well, although support is limited. I wouldn't want folks to think the only way they can develop a canister is to use gitpod, Motoko Playground, or VSCode.

other than that, looks good!

PR #2505, opened 2024-02-14
Update dfx new instances with new instructions for dfx 0.17.0

jessiemongeon1:


docs/developer-docs/backend/motoko/access-control.mdx
```diff
@@ -30,10 +33,61 @@ Open a terminal window on your local computer, if you don’t already have one o
 
 Create a new project by running the following command:
 
+<Tabs groupId="version">
```
dfx-json, line 36: suggestion: if you find yourself repeating the same snippet, it could be turned into a React component and imported into a `.mdx` file.

dfx-json:
@jessiemongeon1 I put together a small PR that shows how to extract some common code into a snippet, and then import that snippet into an existing file. Might help to reduce the footprint of this PR: https://github.com/dfinity/portal/pull/2507

jessiemongeon1:
> @jessiemongeon1 I put together a small PR that shows how to extract some common code into a snippet, and then import that snippet into an existing file. Might help to reduce the footprint of this PR: #2507

@dfx-json Thanks! This will help replace a few, but most of the examples use a different project name that is specific to the guide, so unfortunately this component can't be used in each instance, but it can definitely be used for all of the 'Hello world' instances!

docs/developer-docs/backend/motoko/access-control.mdx
jessiemongeon1: Updated to use the React component for all 'hello_world' examples

docs/developer-docs/backend/motoko/access-control.mdx
```diff
@@ -1,3 +1,8 @@
+import DfxNewSnippet from "/src/components/Snippets/DfxNewSnippet.mdx";
+
+import Tabs from '@theme/Tabs';
```
dfx-json, line 3: these imports can be removed I think

dfx-json (approved):
this is great Jessie. thanks for being on top of it!

PR #2509, opened 2024-02-15
Add a smart contracts call overview page

dsarlis:


jessiemongeon1 (approved):

docs/developer-docs/smart-contracts/call-overview.mdx
```diff
@@ -0,0 +1,161 @@
...
+</Tabs>
+
+## Comparison of update calls vs query calls
+
+| Update calls | Query Calls |
+| :----------------: | :------: |
+| Slow | Fast |
+| Goes through consensus | Does not go through consensus |
+| Asynchronous response | Synchronous response |
+| Executed on all nodes of a subnet | Executed on a single node |
```
domwoe, line 129:
```suggestion
| Update calls | Query Calls |
| :----------------: | :------: |
| Slow (1-2s) | Fast (200-400ms) |
| Can modify state | Can't modify state |
| Goes through consensus | Does not go through consensus |
| Asynchronous response | Synchronous response |
| Executed on all nodes of a subnet | Executed on a single node |
```

docs/developer-docs/smart-contracts/call-overview.mdx
```diff
@@ -0,0 +1,161 @@
...
+    global message
+    message = new_message
+```
+
+</TabItem>
+</Tabs>
+
+## Query calls
+
+Query calls are executed on a single node and return a synchronous response. Since they execute on a single node, they do not go through consensus and can be much faster than update calls.
```
domwoe, line 71: I'd recommend a warning here that a single malicious node could lie about a response, and an update call or a certified query (link) should be used for security-critical calls.

dsarlis: Good point, I added a warning.

domwoe (approved):
Thank you!

PR #2512, opened 2024-02-16
Add a message execution properties page

dsarlis:


docs/references/message-execution-properties.mdx
```diff
@@ -0,0 +1,52 @@
...
+  // block 2
+  ...
+}
+```
+
+
+In this example, there are two message executions involved to process a single call to the `example()` method of the canister. The first one involves “block 1” up to and including the point where the inter-canister call is made while the second message execution involves “block 2” until the end of the function declaration. The two message executions will always be scheduled sequentially.
+
+
+**Property 3**: Successfully delivered requests are received in the order in which they were sent. In particular, if canister A sends m1 and m2 in that order to canister B, then, if both are accepted, m1 will be executed before m2.
```
jessiemongeon1, line 25:
```suggestion
### Property 3

Successfully delivered requests are received in the order in which they were sent. In particular, if canister A sends m1 and m2 in that order to canister B, then, if both are accepted, m1 will be executed before m2.
```

docs/references/message-execution-properties.mdx
```diff
@@ -0,0 +1,52 @@
...
+
+:::info
+Note that this property only gives a guarantee on when the messages are executed, but there is no guarantee on the ordering of the responses received.
+:::
+
+
+**Property 4**: Messages from interleaving calls have no reliable execution ordering.
```
jessiemongeon1, line 32:
```suggestion
### Property 4

Messages from interleaving calls have no reliable execution ordering.
```

docs/references/message-execution-properties.mdx
```diff
@@ -0,0 +1,52 @@
...
+Property 3 provides a guarantee on the execution order of messages on a target canister. If multiple calls interleave, one cannot assume additional ordering guarantees for these interleaving calls. Your code should result in a correct state regardless of the message ordering.
+
+To illustrate this, let's consider the above example code again, and assume the method example is called twice in parallel, the resulting calls being Call 1 and Call 2. The following illustration shows two possible message orderings. On the left, both messages of the first call are scheduled first, and only then the second call's messages are executed (the dotted circles are used to denote the callback messages). On the right, you can see another possible message scheduling, where the first messages of each call are executed first and then their callback messages. 
+
+<div class="text--center">
+<img src="/img/docs/interleaving_calls.png" alt="Interleaving Calls" width="500"/>
+</div>
+
+
+**Property 5**: On a trap, modifications to the canister's state for the current message are not applied.
```
jessiemongeon1, line 43:
```suggestion
### Property 5

On a trap, modifications to the canister's state for the current message are not applied.
```

docs/references/message-execution-properties.mdx
```diff
@@ -0,0 +1,52 @@
...
+
+For example, if a trap in the second message (dotted circle) of the above example occurs, canister state changes resulting from that message are discarded. However, note that any state changes that happened in the first message (solid circle) have been applied if that message executed successfully.
+
+
+**Property 6**: Inter-canister calls are not guaranteed to make it to the destination canister. If the call does reach the destination canister, the destination canister can trap or return a **reject** response while processing the call.
```
jessiemongeon1, line 48:
```suggestion
### Property 6

Inter-canister calls are not guaranteed to make it to the destination canister. If the call does reach the destination canister, the destination canister can trap or return a **reject** response while processing the call.
```

jessiemongeon1 (approved):

PR #2563, opened 2024-02-26
Update ic-admin doc

jessiemongeon1:


dprats (approved):

dprats:
I approved the PR, but I would hesitate from mergeing until the preview build actually builds (in case there is some subtle reason the build fails like a character off).

PR #2586, opened 2024-03-04
feat: add tooltip component

dfx-json:
Adds a new tooltip component. This component can be used to embellish certain phrases with a popover that provides additional context / clarification.

The tooltip is powered by the [floating UI library](https://floating-ui.com/docs/tooltip).

I have also added a new section for UI tests in the documentation that only shows up in development mode. The purpose of this section is to test out features such as the Tooltip extensively during development before rolling out big changes to production. This section can be accessed in the main subnav under the heading "UI Tests"

At first, I attempted to use [Storybook](https://storybook.js.org/) for this, but I found it difficult to align storybook's webpack configuration with the one used by docusaurus. Both are internal to each respective system. I got close, but the added complexity wasn't worth it in my opinion. I think it's more valuable to use the same build system that builds and runs the docs pages to test new UI changes as well - this way we can be sure there will be few differences between what you see in development vs production.

Figma mockups: https://www.figma.com/file/4QNgQZciSxhF227kNnH9q3/Dev-Docs?type=design&node-id=198-5243&mode=design&t=LkIDO6pPnBqR8dXx-0

Screenshots:
Screenshot 2024-03-06 at 09 55 45


plugins/utils/redirects.js
```diff
@@ -432,8 +424,6 @@ const redirects = `
   /docs/current/developer-docs/integrations/https-outcalls/https-outcalls-how-to-use /docs/current/developer-docs/smart-contracts/advanced-features/https-outcalls/https-outcalls-how-to-use
   /docs/current/developer-docs/integrations/https-outcalls/https-outcalls-overview /docs/current/developer-docs/smart-contracts/advanced-features/https-outcalls/https-outcalls-overview
   /docs/current/developer-docs/integrations/https-outcalls/https-outcalls-post /docs/current/developer-docs/smart-contracts/advanced-features/https-outcalls/https-outcalls-post
-  /docs/current/developer-docs/integrations/bitcoin/ckBTC-reference /docs/current/references/ckbtc-reference
```
jessiemongeon1, line 435: These ckBTC redirects are necessary since they redirect case-sensitive URLs to the correct lowercase URLs.

jessiemongeon1 (approved):

plugins/utils/redirects.js
dfx-json: I should have only deleted a duplicate. I'll double check

dfx-json: yep, I only deleted a duplicate. `/docs/current/developer-docs/integrations/bitcoin/ckbtc-reference /docs/current/references/ckbtc-reference` is still there

jessiemongeon1: We still need the redirects for `/docs/current/developer-docs/integrations/bitcoin/ckBTC-reference` and `/docs/current/tutorials/developer-journey/level-4/4.3-ckBTC-and-bitcoin`  since this URL is case-sensitive. Slack thread about this here: https://dfinity.slack.com/archives/C03856W4LRG/p1708731304240029

jessiemongeon1: 🤔 That's weird. I guess let's remove them for now so the build passes, and I can take a closer look at re-adding them in the future.

PR #2517, opened 2024-02-16
add reference to comparison with Ethereum to smart contracts overview

domwoe:
Thank you for your contribution to the IC Developer Portal.
Before submitting your Pull Request, please make sure that:

- [ ] You have updated the [`.github/CODEOWNERS`](https://github.com/dfinity/portal/blob/master/.github/CODEOWNERS) file according to your changes.
- [ ] You have registered new documents to [`/sidebars.js`](https://github.com/dfinity/portal/blob/master/sidebars.js) if any.

PR #2620, opened 2024-03-07
Implement tooltips

jessiemongeon1:


docs/developer-docs/getting-started/overview-of-icp.mdx
```diff
@@ -244,7 +245,7 @@ Currently, subnets contain between 13 and 40 nodes, and most of them are geograp
 </tr>
 
 <tr>
-  <td> Upgradability </td>
+  <td> Upgradeability </td>
```
jennifertrin, line 248: I believe that it should be Upgradability because we use "Upgradable" in some places: https://github.com/dfinity/portal/pull/2517/files

jennifertrin (commented):
Woof! Great work on adding all of the tooltips! 

I suggest that long term, we find a solution to make it easier to add and update tooltips. For example, we could programmatically add the same tooltip to the first mention of a specific word.

dfx-json (approved):
Looks good! I think these will be super helpful. Two thoughts come to mind:

1. we ought to darken the background on the tooltip a bit more - maybe get the opacity up to 95%
2. I see the tooltip definitions being repeated for some common phrases, such as `canister`. We can probably create a separate component (e.g. `GlossaryTooltip`) that will automatically look up  the word in some internal dictionary we put together and render the tooltip. Maybe something like
```mdx
This is a <GlossaryTooltip>canister</GlossaryTooltip> deployed on ICP.
```

PR #2634, opened 2024-03-11
fix: move dfx deps II instructions into dfx 0.16.0 + older tab

jessiemongeon1:

### DIFF
--- a/docs/developer-docs/web-apps/user-login/internet-identity/integrate-internet-identity.mdx
+++ b/docs/developer-docs/web-apps/user-login/internet-identity/integrate-internet-identity.mdx
@@ -28,6 +28,12 @@ import TabItem from '@theme/TabItem';
 <Tabs groupId="version">
 <TabItem value="0-17-0" label="dfx v0.17.0 or newer" default>
 
+Start `dfx`:
+
+```
+dfx start --clean --background
+```
+
 Use `dfx new <project_name>` to create a new project:
 
 ```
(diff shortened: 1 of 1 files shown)


docs/developer-docs/web-apps/user-login/internet-identity/integrate-internet-identity.mdx
```diff
@@ -112,17 +130,24 @@ Open the `dfx.json` file and replace the existing content with the following:
...
+```
+dfx deps pull
+```
 
-### Step 2: Start `dfx`:
+Initialize the canister. You can use the `'(null)'` value passed to the init command to use the default values. To do so, run the command:
 
 ```
-dfx start --clean --background
+dfx deps init internet_identity --argument '(null)'
```
frederikrothenberger, line 142: `dfx 0.18.0` actually automatically intializes the argument to `(null)`, so no need to specify this explicitly. Although `dfx deps init` still has to be invoked, see [here](https://dfinity.slack.com/archives/CGA566TPV/p1710148272653119).

frederikrothenberger (approved):
Thanks @jessiemongeon1

docs/developer-docs/web-apps/user-login/internet-identity/integrate-internet-identity.mdx
jessiemongeon1: The instructions in this are for dfx 0.16.0 and older, since in 0.17.0 and newer, II can be added to the project as part of the `dfx new` command instead.

jennifertrin (approved):

PR #2642, opened 2024-03-13
Improve inspect_message best practice

tmu0:


docs/developer-docs/security/rust-canister-development-security-best-practices.mdx
```diff
@@ -718,12 +718,12 @@ Canisters pay for their cycles which makes them inherently vulnerable to attacks
 
 The correct execution of [canister_inspect_message](../../references/ic-interface-spec#system-api-inspect-message) is not guaranteed in the presence of a malicious node, because it is executed as a [query call](../../references/ic-interface-spec#http-query).
```
robin-kunzler, line 719: Would the following make the malicious node issue a bit clearer? 
```suggestion
The correct execution of [canister_inspect_message](../../references/ic-interface-spec#system-api-inspect-message) is not guaranteed because it is executed by a single node and if that node is malicious, it can e.g. simply skip this check. In that case the update call would be executed without any message inspection checks. 
```

docs/tutorials/developer-journey/level-4/4.6-motoko-lvl4.mdx
```diff
@@ -317,7 +317,7 @@ Notice that now the value of `c` is set to `init` rather than set to `0`, to ref
...
+When a canister receives an ingress message submitted through HTTP, the canister can inspect and either accept or decline the message before executing it. For example, when a canister receives an update call, it uses the canister method `canister_inspect_method` to determine if the message will be accepted. If a canister does not contain a Wasm module, the message will be rejected automatically. If the canister is not empty, but it does not implement the `canister_inspect_message` method, then the ingress method will be accepted. `canister_inspect_method` is not used for HTTP query calls, inter-canister calls, for calls made to the management canister. Please make sure to get familiar with the [canister development security best practices](/docs/current/developer-docs/security/rust-canister-development-security-best-practices#do-not-rely-on-ingress-message-inspection) before using this feature in production.
```
robin-kunzler, line 320: What do you think about the following? - Fixed a few typos in the existing text and extended your suggested sentence to hint at the security issue.
```suggestion
When a canister receives an ingress message submitted through HTTP, the canister can inspect and either accept or decline the message before executing it. For example, when a canister receives an update call, it uses the canister method `canister_inspect_message` to determine if the message will be accepted. If a canister does not contain a Wasm module, the message will be rejected automatically. If the canister is not empty, but it does not implement the `canister_inspect_message` method, then the ingress method will be accepted. `canister_inspect_message` is not used for HTTP query calls, inter-canister calls, and for calls made to the management canister. Please make sure to get familiar with the [canister development security best practices](/docs/current/developer-docs/security/rust-canister-development-security-best-practices#do-not-rely-on-ingress-message-inspection) before using this feature in production, because `canister_inspect_message` is not a secure way of implementing e.g. authorization checks. 
```

robin-kunzler (approved):
Thanks a lot @tmu0 ! Looks great. Left two minor improvement suggestions. PTAL but leave those up to you. Approving the PR.

jessiemongeon1 (approved):

PR #2672, opened 2024-03-21
chore: fix syntax highlighting for motoko and candid

dfx-json:
Syntax highlighting wasn't working for Motoko or Candid. This PR simplifies how Motoko and Candid code is rendered, using as much built-in capability provided by Docusaurus as possible.

This also introduces two new grammar files for [Prism](https://github.com/PrismJS/prism/tree/master/components) - `prism-motoko.js` and `prism-candid.js`.

Both of these are based on the `prism-go.js` file and extend the "c-like" grammar. @rvanasa or @chenyan-dfinity I could use your help updating `prism-motoko.js` and `prism-candid.js` with the most accurate grammar definition.

image

### DIFF
--- a/src/theme/prism/prism-candid.js
+++ b/src/theme/prism/prism-candid.js
@@ -1,0 +1,19 @@
+Prism.languages.candid = Prism.languages.extend('clike', {
+  'string': {
+    pattern: /(^|[^\\])"(?:\\.|[^"\\\r\n])*"|`[^`]*`/,
+    lookbehind: true,
+    greedy: true
+  },
+  'keyword': /\b(?:break|case|chan|const|continue|default|defer|else|fallthrough|for|actor|func|go(?:to)?|if|import|interface|map|package|range|return|select|struct|switch|type|var)\b/,
+  'boolean': /\b(?:_|false|iota|nil|true)\b/,
+  'number': [
+    // binary and octal integers
+    /\b0(?:b[01_]+|o[0-7_]+)i?\b/i,
+    // hexadecimal integers and floats
+    /\b0x(?:[a-f\d_]+(?:\.[a-f\d_]*)?|\.[a-f\d_]+)(?:p[+-]?\d+(?:_\d+)*)?i?(?!\w)/i,
+    // decimal integers and floats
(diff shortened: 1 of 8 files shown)


jessiemongeon1 (approved):

src/theme/prism/prism-motoko.js
```diff
@@ -0,0 +1,98 @@
...
+      lookbehind: true,
+      greedy: true
+    },
+    {
+      pattern: /b?'\\?(x\w{2}|u\w{4}|U\w{8}|.)'/,
+      lookbehind: true,
+      greedy: true
+    }
+  ],
+  keyword: new RegExp(keywords.filter((k) => k !== "async").join("|")),
```
chenyan-dfinity, line 87: Why excluding `async`?

src/theme/prism/prism-candid.js
```diff
@@ -0,0 +1,26 @@
...
+      greedy: true
+    },
+    {
+      pattern: /b?'\\?(x\w{2}|u\w{4}|U\w{8}|.)'/,
+      lookbehind: true,
+      greedy: true
+    }
+  ],
+  keyword:
+    /\b(?:opt|vec|record|variant|func|blob|principal|nat|nat8|nat16|nat32|nat64|int|int8|int16|int32|int64|float32|float64|bool|text|null|reserved|empty|oneway|query')\b/,
```
chenyan-dfinity, line 15:
```suggestion
    /\b(?:opt|vec|record|variant|func|blob|principal|nat|nat8|nat16|nat32|nat64|int|int8|int16|int32|int64|float32|float64|bool|text|null|reserved|empty|oneway|query|composite_query')\b/,
```

src/theme/prism/prism-motoko.js
dfx-json: just following https://github.com/rvanasa/highlightjs-motoko/blob/main/src/languages/motoko.js as closely as I can

rvanasa: The original reasoning was to render `async` in the same way as type identifiers.

rvanasa (approved):

PR #2701, opened 2024-03-25
Add reference for tSchnorr Ed25519 key derivation

fspreiss:
Adds a reference page for Ed25519 key derivation for the upcoming threshold Schnorr signatures features.

This reference will be required by an upcoming PR that proposes a respective management canister API for the threshold Schnorr signatures feature.

### DIFF
--- a/docs/references/t-schnorr-how-it-works.mdx
+++ b/docs/references/t-schnorr-how-it-works.mdx
@@ -49,7 +49,7 @@ We assume the following functions:
 
 Public keys are extended with an extra 256 bits of entropy, which extension is called chain code and consists of 32 bytes.
 An extended public key is represented as (K, c), with K = point(k) and c being the chain code, for some private key k.
-Each extended key has 2<sub>32</sub> child keys using indices 0 through 2<sub>32</sub>-1.
+Each extended key can have an arbitrary number of child keys.
 We do not distinguish between hardened and non-hardenend (aka normal) child keys.
 
 ### Child key derivation (CKD) function
--- a/docs/references/t-schnorr-how-it-works.mdx
+++ b/docs/references/t-schnorr-how-it-works.mdx
@@ -60,6 +60,6 @@ The function CKDpub((K<sub>par</sub>, c<sub>par</sub>), i) → (K<sub>i</sub>, c
 
 - let IKM = ser<sub>P</sub>(K<sub>par</sub>) || i.
 - let OKM = HKDF(c<sub>par</sub>, IKM, utf8("Ed25519"), 96).
-- Split OKM into a 64-byte and a 32-byte sequence, tweak and c<sub>i</sub>.
+- Split OKM into a 64-byte and a 32-byte sequence, tweak and c<sub>i</sub>, and reduce tweak modulo n.
 - let K<sub>i</sub> = K<sub>par</sub> + point(parse<sub>512</sub>(tweak)).
 - return (K<sub>i</sub>, c<sub>i</sub>).
--- a/docs/references/t-schnorr-how-it-works.mdx
+++ b/docs/references/t-schnorr-how-it-works.mdx
@@ -23,7 +23,7 @@ For this reason, a new derivation scheme is specified here.
 The specified scheme does not make use of _clamping_ (see [RFC8032, Section 5.1.5, Item 2](https://datatracker.ietf.org/doc/html/rfc8032#section-5.1.5)), because it is unnecessary in the given setting, and satisfies the following requirements:
 
 - Off-chain availability: New public keys can be computed off-chain from a master public key without requiring interaction with the IC.
-- Hierarchical derivation: Derived keys are organized in a tree such that from any public key it is possible to derive new child keys. This tree can have arbitrary depth. The first level is used to derive unique canister-specific keys from the master key.
+- Hierarchical derivation: Derived keys are organized in a tree such that from any public key it is possible to derive new child keys. The first level is used to derive unique canister-specific keys from the master key.
 - Compatibility: The scheme is compatible with existing standards, and it is simple to implement using existing libraries.
 
 ### Conventions
--- a/docs/references/t-schnorr-how-it-works.mdx
+++ b/docs/references/t-schnorr-how-it-works.mdx
@@ -30,7 +30,7 @@ The specified scheme does not make use of _clamping_ (see [RFC8032, Section 5.1.
 
 We will assume the elliptic curve cryptography using the field and curve parameters defined by Ed25519 (see [RFC8032, Section 5.1](https://datatracker.ietf.org/doc/html/rfc8032#section-5.1)). Variables below are either:
 
-- Integers modulo the order of the curve (referred to as L).
+- Integers modulo the order of the curve's prime order subgroup (referred to as L).
 - Coordinates of points on the curve.
 - Byte sequences.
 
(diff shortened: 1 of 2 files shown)


docs/references/t-schnorr-how-it-works.mdx
```diff
@@ -0,0 +1,65 @@
...
+The section is inspired by [BIP32](https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki) and uses similar wording and structure.
+
+### Motivation
+
+To support the Ed25519 variant of threshold Schnorr signatures on the Internet Computer, a key derivation scheme compatible with Ed25519 signatures is required.
+For a respective signing service on the Internet Computer to be efficient, the signing subnet maintains only a single master key pair and _derives_ signing child keys for each canister.
+Although there exist various hierarchical key derivation schemes (e.g., [BIP32](https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki), [SLIP10](https://github.com/satoshilabs/slips/blob/master/slip-0010.md), [BIP32-Ed25519](https://input-output-hk.github.io/adrestia/static/Ed25519_BIP.pdf), [Schnorrkel](https://github.com/w3f/schnorrkel)), none of the analyzed schemes is suitable for the Internet Computer's (threshold) setting.
+For this reason, a new derivation scheme is specified here.
+The specified scheme does not make use of _clamping_, because it is unnecessary in the given setting, and satisfies the following requirements:
```
randombit, line 23: It might be worth a reference here to explain what clamping means in this context, perhaps something like https://datatracker.ietf.org/doc/html/rfc8032#section-5.1.5 (which explains clamping but unfortunately does not name it as such, even though that is what it calls it in the reference Python code in the appendix...)

docs/references/t-schnorr-how-it-works.mdx
```diff
@@ -0,0 +1,65 @@
...
+
+- Off-chain availability: New public keys can be computed off-chain from a master public key without requiring interaction with the IC.
+- Hierarchical derivation: Derived keys are organized in a tree such that from any public key it is possible to derive new child keys. This tree can have arbitrary depth. The first level is used to derive unique canister-specific keys from the master key.
```
randombit, line 26: Actually the tree is limited to 255 levels, matching BIP32

docs/references/t-schnorr-how-it-works.mdx
```diff
@@ -0,0 +1,65 @@
...
+- Compatibility: The scheme is compatible with existing standards, and it is simple to implement using existing libraries.
+
+### Conventions
+
+We will assume the elliptic curve cryptography using the field and curve parameters defined by Ed25519 (see [RFC8032, Section 5.1](https://datatracker.ietf.org/doc/html/rfc8032#section-5.1)). Variables below are either:
+
+- Integers modulo the order of the curve (referred to as n).
```
randombit, line 33: While in most literature `n` is the typical variable used to refer to the group order, Ed25519 notation usually uses `L` instead (for example RFC 8032 calls it `L`, using `n` to refer to the bitlength of the order)

Given the existence of the Ed25519 cofactor this should probably also specify that we are referring to the prime order subgroup, rather than the overall elliptic curve which has order `8*L`

docs/references/t-schnorr-how-it-works.mdx
```diff
@@ -0,0 +1,65 @@
...
+- ser<sub>P</sub>(P): serializes the coordinate pair P = (x,y) as a byte sequence using standard 32-byte compressed form (see [RFC8032, 5.1.2 Encoding](https://datatracker.ietf.org/doc/html/rfc8032#section-5.1.2)).
+- utf8(s): returns the UTF-8 encoding of string s.
+- parse<sub>512</sub>(p): interprets a 64-byte sequence as a 512-bit number, most significant byte first.
+- HKDF(salt,IKM,info,L) -> OKM: HMAC-based key derivation function (see [RFC5869](https://datatracker.ietf.org/doc/html/rfc5869)) using HMAC-SHA512 (see [RFC4231](https://datatracker.ietf.org/doc/html/rfc4231)) calculating L-bytes long output key material (OKM) from (byte sequences) salt, input key material (IKM), and application specific information *info*.
+
+### Extended keys
+
+Public keys are extended with an extra 256 bits of entropy, which extension is called chain code and consists of 32 bytes.
+An extended public key is represented as (K, c), with K = point(k) and c being the chain code, for some private key k.
+Each extended key has 2<sub>32</sub> child keys using indices 0 through 2<sub>32</sub>-1.
```
randombit, line 52: This comment seems BIP32 specific - in our extended case the key has effectively unlimited child keys since any arbitrary bytestring can be used.

docs/references/t-schnorr-how-it-works.mdx
fspreiss: Good point. I added the reference you suggested.

docs/references/t-schnorr-how-it-works.mdx
fspreiss: > While in most literature n is the typical variable used to refer to the group order, Ed25519 notation usually uses L instead (for example RFC 8032 calls it L, using n to refer to the bitlength of the order)

n is now renamed to L. We also had an L for HKDF: this L was renamed to N.

> Given the existence of the Ed25519 cofactor this should probably also specify that we are referring to the prime order subgroup, rather than the overall elliptic curve which has order 8*L

I adapted the sentence to "Integers modulo the order of the curve's prime order subgroup (referred to as L)."

docs/references/t-schnorr-how-it-works.mdx
fspreiss: True! I adapted the sentence to "Each extended key can have an arbitrary number of child keys."

docs/references/t-schnorr-how-it-works.mdx
fspreiss: There is indeed a restriction to 255 levels. However, this is not a restriction of the key derivation scheme specified in this PR (which scheme stands somewhat by itself), but rather an artificial restriction we impose as per the IC interface spec to protect against various kinds of attacks. In any case, at least for now, I removed the sentence "This tree can have arbitrary depth." to avoid confusion. Let me know if you think I should add back the sentence.

randombit: Fair - unlike BIP32 there is no external constraint that limits us to depth 255 vs some other depth.

randombit (approved):

docs/references/t-schnorr-how-it-works.mdx
```diff
@@ -0,0 +1,80 @@
+import { MarkdownChipRow } from "/src/components/Chip/MarkdownChipRow";
```
jessiemongeon1, line 1:
```suggestion
---
keywords: [reference, tschnorr, threshold schnorr signatures, signatures]
---


import { MarkdownChipRow } from "/src/components/Chip/MarkdownChipRow";
```

docs/references/t-schnorr-how-it-works.mdx
```diff
@@ -0,0 +1,80 @@
...
+
+# Threshold Schnorr signatures reference
+
+<MarkdownChipRow labels={["Reference"]} />
+
+## Overview
+Currently, this page (only) contains a technical specification for performing Ed25519 hierarchical key derivation.
+This specification is relevant when using the Ed25519 variant of threshold Schnorr signatures on ICP.
+It is, for example, of interest as reference for developers who implement libraries that perform this type of key derivation.
```
jessiemongeon1, line 10:
```suggestion
It is, for of interest as reference for developers who implement libraries that perform this type of key derivation.
```

jessiemongeon1 (approved):

docs/references/t-schnorr-how-it-works.mdx
fspreiss: 👍

fspreiss:
The tSchnorr Ed25519 key derivation scheme is not directly included in the interface spec PR (see https://github.com/dfinity/interface-spec/pull/288). With this, I'm closing this PR.

PR #2719, opened 2024-03-28
Make the featured event use the actual picture from airtable

meodai:
Makes sure the featured event also uses the airtable picture


marc0olo:
I assume this is not relevant anymore, closing this (cc @StepankaHru)

PR #2728, opened 2024-04-02
chore: fix permission in docs

saikatdas0790:


jessiemongeon1:
Hi @saikatdas0790 , 

`Commit` includes `Prepare`, so this edit is not necessary. You can learn more here: https://forum.dfinity.org/t/failed-to-create-batch-the-replica-returned-a-replica-error-reject-code-canisterreject-reject-message-caller-does-not-have-prepare-permission-error-code-none/27319

PR #2762, opened 2024-04-09
Add Non-Replicated and Replicated Query Terminology to Smart Contract Call Page

jennifertrin:

### DIFF
--- a/docs/developer-docs/smart-contracts/call/overview.mdx
+++ b/docs/developer-docs/smart-contracts/call/overview.mdx
@@ -167,7 +167,7 @@ Certified queries use certified variables to prove the authenticity of a piece o
 
 ### Replicated queries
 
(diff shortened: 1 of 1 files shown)


docs/developer-docs/smart-contracts/call/overview.mdx
```diff
@@ -165,6 +165,10 @@ Composite queries are query calls that can call other queries (on the same <Glos
 
 Certified queries use certified variables to prove the authenticity of a piece of data that comes along with the query's response. Certified variables can be set via an update call and can then be read via a query call.
 
+### Replicated queries
+
+Replicated queries, also referred to as "query-as-update" execution mode, are query calls executed as updates. The queries still discard the state changes, but execution happens on all nodes and the execution results go through consensus.
```
jessiemongeon1, line 170:
```suggestion
Replicated queries, also referred to as the "query-as-update" execution model, are query calls executed as updates. The query still discards the state changes, but execution happens on all subnet nodes and the results go through consensus.
```

jessiemongeon1 (approved):

PR #2809, opened 2024-04-22
Add performance limits to the resource limits page

ulan:
This also updates the limit for update calls from 20B to 40B.

### DIFF
--- a/docs/developer-docs/smart-contracts/maintain/resource-limits.mdx
+++ b/docs/developer-docs/smart-contracts/maintain/resource-limits.mdx
@@ -30,7 +30,7 @@ The limits depend on the message type as shown in the following table.
 | Instruction resource limits                                                          | Constraint  |
 | ------------------------------------------------------------------------------------ | ----------- |
 | Instruction limit, instructions per update call/heartbeat/timer                      | 40 Billion  |
-| Instruction limit, instructions per query calls                                      | 5 Billion   |
+| Instruction limit, instructions per query call                                      | 5 Billion   |
 | Instruction limit, instructions per canister install/upgrade                         | 200 Billion |
 | Instruction limit, instructions per inspect_message                                  | 200 Million |
 


docs/developer-docs/smart-contracts/maintain/resource-limits.mdx
```diff
@@ -29,7 +29,7 @@ The limits depend on the message type as shown in the following table.
 
 | Instruction resource limits                                                          | Constraint  |
 | ------------------------------------------------------------------------------------ | ----------- |
-| Instruction limit, instructions per update call/heartbeat/timer                      | 20 Billion  |
+| Instruction limit, instructions per update call/heartbeat/timer                      | 40 Billion  |
 | Instruction limit, instructions per query calls                                      | 5 Billion   |
```
yvonneanne, line 33:
```suggestion
| Instruction limit, instructions per query call                                      | 5 Billion   |
```

docs/developer-docs/smart-contracts/maintain/resource-limits.mdx
```diff
@@ -57,6 +57,17 @@ The limits depend on the message type as shown in the following table.
 | Update calls execution threads, per subnet                                           | 4           |
 | Update calls execution threads, per canister                                         | 1           |
 
+## Performance limits
+
+The block rate varies from 0.75 blocks per second to 1.5 blocks per second depending on the subnet and the number of nodes in the subnet.
+
+Up to 1000 messages can be included in a block.
```
yvonneanne, line 64: What we could add

1. The messages are selected round-robin per canister from a pool of up to 10_000 messages. 
2. Nodes do not admit messages to their pool when its full.
3. Messages are deleted from the pool after they have been received by execution or after they expired.

I am not sure how much it helps the developers though... What do you think?

What I'd definitely omit is the fact that for the NNS subnet the limit is 400 messages per block, since ordinary developers cannot deploy there

ulan: That seems too low-level for most developers.

jessiemongeon1: I agree with Ulan, this might be too low-level for this page. Maybe it would fit in [this reference page](https://internetcomputer.org/docs/current/references/message-execution-properties) though?

jessiemongeon1 (approved):

docs/developer-docs/smart-contracts/maintain/resource-limits.mdx
jessiemongeon1: I agree with Ulan, I think this it too low-level for this page. Maybe it would fit in [this reference page](https://internetcomputer.org/docs/current/references/message-execution-properties) though?

yvonneanne: Thanks, Ulan and Jessie! Let's keep it out then :)

I wouldn't add it to the other page, since it has nothing to do with execution.

yvonneanne (approved):

Issue #2821, opened 2024-04-24
Change all internal references to be relative


andrew-lee-work:
I believe all references to IC documentation need to be made relatively. So all https://internetcomputer.org/docs/current/... need to be replaced. If had someone change my PR to fix this but maybe we can try to do that ourselves?

_Originally posted by @roelstorms in https://github.com/dfinity/portal/issues/2356#issuecomment-2037231944_

jessiemongeon1:
Addressed in https://github.com/dfinity/portal/pull/2850

PR #2850, opened 2024-04-30
Update all Dev Docs to use relative links

jessiemongeon1:
Addresses issue https://github.com/dfinity/portal/issues/2821


dfx-json (approved):
One thing I'm curious about is if the use of relative urls is triggering the problem of failed redirects. i.e. when we have a redirect defined, click on a link, and get a 404, only to navigate to that url directly and it works. I'm still trying to trace down the cause of that issue

PR #2878, opened 2024-05-06
Add details on more execution errors

adambratschikaye:
Added errors:
Instruction limit exceeded
Trapped
Trapped Explicitly
Wasm module not found
Out of memory
Reserved pages for old Motoko
Slice overrun
Memory access limit exceeded
Insufficient cycles in memory grow
Reserved cycles limit exceeded in memory grow
Insufficient cycles in message memory grow
Wasm memory limit exceeded

### DIFF
--- a/docs/developer-docs/smart-contracts/maintain/resource-limits.mdx
+++ b/docs/developer-docs/smart-contracts/maintain/resource-limits.mdx
@@ -46,6 +46,8 @@ The limits depend on the message type as shown in the following table.
 | Wasm custom sections, per canister                                                   | 1MiB        |
 | Wasm custom sections, sections per canister                                          | 16          |
 | Wasm code section, per canister                                                      | 10MiB       |
+| Wasm stable memory, data access per message                                          | 8GiB        |
+| Wasm stable memory, data written per message                                         | 8GiB        |
 
 | Query call resource limits                                                           | Constraint  |
 | ------------------------------------------------------------------------------------ | ----------- |
--- a/docs/references/execution-errors.mdx
+++ b/docs/references/execution-errors.mdx
@@ -9,8 +9,9 @@ A list of possible errors returned when executing canisters.
 
 ## Errors
 
+
 ### Method not found
-A canister was called with a method name not exported by that canister.
+The canister was called with a method name not exported by that canister.
 
 An example of this error is:
 ```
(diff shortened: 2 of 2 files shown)


docs/references/execution-errors.mdx
```diff
@@ -23,4 +24,190 @@ To fix this error, consider:
...
+The canister reached the maximum number of allowed instructions before completing execution.
+
+An example of this error is:
+```
+  Error from Canister xxx-xxx: Canister exceeded the instruction limit for single message execution.
+```
+
+Instruction limits differ depending on the message type and can be found in the [documentation on resource limits][limits].
+
+To fix this error, consider using tools such as the [performance counter API](https://internetcomputer.org/docs/current/references/samples/rust/performance_counters/) or [canbench][canbench] to determine which sections of code are using significant instructions and optimize them.
```
jessiemongeon1, line 40:
```suggestion
To fix this error, consider using tools such as the [performance counter API](/docs/current/references/samples/rust/performance_counters/) or [canbench](/docs/current/developer-docs/smart-contracts/test/benchmarking) to determine which sections of code are using significant instructions and optimize them.
```

docs/references/execution-errors.mdx
```diff
@@ -23,4 +24,190 @@ To fix this error, consider:
...
+The canister exists, but has no Wasm module installed.
+
+An example of this error is:
+```
+  Error from Canister xxx-xxx: Attempted to execute a message, but the canister contains no Wasm module.
+```
+
+Canisters can [exist without having Wasm code installed](https://internetcomputer.org/docs/current/references/ic-interface-spec#canister-lifecycle). A canister will have no Wasm code if it has never been installed, or if it has been [uninstalled](https://internetcomputer.org/docs/current/references/ic-interface-spec#ic-uninstall_code).
+
+To better diagnose this error use the [`dfx canister status`][dfx-canister-status] command or the [`canister_status` API][canister-status-api] to check if the canister has a module installed. If there is a module the "Module hash" field will be non-null.
```
jessiemongeon1, line 79:
```suggestion
To better diagnose this error use the [`dfx canister status`](/docs/current/developer-docs/developer-tools/cli-tools/cli-reference/dfx-canister#dfx-canister-status) command or the [`canister_status` API](/docs/current/references/ic-interface-spec/#ic-canister_status) to check if the canister has a module installed. If there is a module the "Module hash" field will be non-null.
```

docs/references/execution-errors.mdx
```diff
@@ -23,4 +24,190 @@ To fix this error, consider:
...
+
+### Wasm module not found
+The canister exists, but has no Wasm module installed.
+
+An example of this error is:
+```
+  Error from Canister xxx-xxx: Attempted to execute a message, but the canister contains no Wasm module.
+```
+
+Canisters can [exist without having Wasm code installed](https://internetcomputer.org/docs/current/references/ic-interface-spec#canister-lifecycle). A canister will have no Wasm code if it has never been installed, or if it has been [uninstalled](https://internetcomputer.org/docs/current/references/ic-interface-spec#ic-uninstall_code).
```
jessiemongeon1, line 77:
```suggestion
Canisters can [exist without having Wasm code installed](/docs/current/references/ic-interface-spec#canister-lifecycle). A canister will have no Wasm code if it has never been installed, or if it has been [uninstalled](/docs/current/references/ic-interface-spec#ic-uninstall_code).
```

docs/references/execution-errors.mdx
```diff
@@ -23,4 +24,190 @@ To fix this error, consider:
...
+
+### Trapped explicitly
+The canister aborted execution by calling the `ic0.trap` API.
+
+An example of this error is:
+```
+  Error from Canister xxx-xxx: Canister called `ic0.trap` with message: <Canister error message>
+```
+
+When encountering an error, canister's may choose to fail with an error message by calling the [`ic0.trap` API](https://internetcomputer.org/docs/current/references/ic-interface-spec#debugging-aids). The Rust and Motoko CDK's will insert calls to this API when panicking.
```
jessiemongeon1, line 64:
```suggestion
When encountering an error, canister's may choose to fail with an error message by calling the [`ic0.trap` API](/docs/current/references/ic-interface-spec#debugging-aids). The Rust and Motoko CDK's will insert calls to this API when panicking.
```

docs/references/execution-errors.mdx
```diff
@@ -23,4 +24,190 @@ To fix this error, consider:
...
+An example of this error is:
+```
+  Error from Canister xxx-xxx: Attempted to execute a message, but the canister contains no Wasm module.
+```
+
+Canisters can [exist without having Wasm code installed](https://internetcomputer.org/docs/current/references/ic-interface-spec#canister-lifecycle). A canister will have no Wasm code if it has never been installed, or if it has been [uninstalled](https://internetcomputer.org/docs/current/references/ic-interface-spec#ic-uninstall_code).
+
+To better diagnose this error use the [`dfx canister status`][dfx-canister-status] command or the [`canister_status` API][canister-status-api] to check if the canister has a module installed. If there is a module the "Module hash" field will be non-null.
+
+To fix this error, consitder installing code using [`dfx deploy`](https://internetcomputer.org/docs/current/developer-docs/developer-tools/cli-tools/cli-reference/dfx-deploy), or the APIs for [`install_code`](https://internetcomputer.org/docs/current/references/ic-interface-spec#ic-install_code) or [`install_chunked_code`](https://internetcomputer.org/docs/current/references/ic-interface-spec#ic-install_chunked_code).
```
jessiemongeon1, line 81:
```suggestion
To fix this error, consitder installing code using [`dfx deploy`](/docs/current/developer-docs/developer-tools/cli-tools/cli-reference/dfx-deploy), or the APIs for [`install_code`](/docs/current/references/ic-interface-spec#ic-install_code) or [`install_chunked_code`](/docs/current/references/ic-interface-spec#ic-install_chunked_code).
```

docs/references/execution-errors.mdx
```diff
@@ -23,4 +24,190 @@ To fix this error, consider:
...
+
+### Out of memory
+The canister was tried to request more memory than its allocation allowed during execution, causing the execution to fail.
+
+An example of this error is:
+```
+  Error from Canister xxx-xxx: Canister exceeded its allowed memory allocation.
+```
+
+There are [system wide limits][limits] on the main and stable memory of each canister, as well as limits on the total memory of a subnet. This error could be triggered by any one of those limits being reached.
```
jessiemongeon1, line 92:
```suggestion
There are [system wide limits](/docs/current/developer-docs/smart-contracts/maintain/resource-limits/) on the main and stable memory of each canister, as well as limits on the total memory of a subnet. This error could be triggered by any one of those limits being reached.
```

docs/references/execution-errors.mdx
```diff
@@ -23,4 +24,190 @@ To fix this error, consider:
...
+
+In addition, canisters may reserve memory using the [`memory-allocation` setting](https://internetcomputer.org/docs/current/developer-docs/developer-tools/cli-tools/cli-reference/dfx-canister/#options-8). In this case the canister is guaranteed to be able to use up to the allocated memory, but will receive an "Out of memory" error when trying to use more than the allocated amount.
```
jessiemongeon1, line 94:
```suggestion
In addition, canisters may reserve memory using the [`memory-allocation` setting](/docs/current/developer-docs/developer-tools/cli-tools/cli-reference/dfx-canister/#options-8). In this case the canister is guaranteed to be able to use up to the allocated memory, but will receive an "Out of memory" error when trying to use more than the allocated amount.
```

docs/references/execution-errors.mdx
```diff
@@ -23,4 +24,190 @@ To fix this error, consider:
...
+
+To diagnose this error use the [`dfx canister status`][dfx-canister-status] command or the [`canister_status` API][canister-status-api] to check the canister's current memory usage and memory allocation. The subnet memory usage can also be seen on the [ICP dashboard](https://dashboard.internetcomputer.org/subnets).
```
jessiemongeon1, line 96:
```suggestion
To better diagnose this error use the [`dfx canister status`](/docs/current/developer-docs/developer-tools/cli-tools/cli-reference/dfx-canister#dfx-canister-status) command or the [`canister_status` API](/docs/current/references/ic-interface-spec/#ic-canister_status) to check the canister's current memory usage and memory allocation. The subnet memory usage can also be seen on the [ICP dashboard](https://dashboard.internetcomputer.org/subnets).
```

docs/references/execution-errors.mdx
```diff
@@ -23,4 +24,190 @@ To fix this error, consider:
...
+
+To fix this error, consider:
+
+  - If the canister has reached its current memory allocation, try allocating more memory.
+  - If the canister has reached the system-wide limits for memory usage and it seems reasonable for the canister to have used 100s of GiBs, try sharding the data across multiple canisters.
+  - If the canister has unexpectedly reached the system-wide limits, try debugging it to see if there could be a memory leak.[canbench][canbench] can help with profiling memory usage.
```
jessiemongeon1, line 102:
```suggestion
  - If the canister has unexpectedly reached the system-wide limits, try debugging it to see if there could be a memory leak. [Canbench](/docs/current/developer-docs/smart-contracts/test/benchmarking) can help with profiling memory usage.
```

docs/references/execution-errors.mdx
```diff
@@ -23,4 +24,190 @@ To fix this error, consider:
...
+
+### Reserved pages for old Motoko
+The canister is running an older version of Motoko which requires reserving some additional pages on the Wasm heap for upgrading and tried to allocate those reserved pages.
+
+An example of this error is:
+```
+  Canister tried to allocate pages reserved for upgrading older versions of Motoko.
+```
+
+Newer versions of Motoko don't require these reserved pages, so upgrading the version of Motoko used in the canister will fix the issue. `dfx` stores the Motoko compiler (`moc`) in the directory returned by [`dfx cache show`](https://internetcomputer.org/docs/current/developer-docs/developer-tools/cli-tools/cli-reference/dfx-cache#dfx-cache-show). So the version of Motoko can be shown by running `$(dfx cache show)/moc --version`. This issue only occurs for versions `0.6.20` and older.
```
jessiemongeon1, line 114:
```suggestion
Newer versions of Motoko don't require these reserved pages, so upgrading the version of Motoko used in the canister will fix the issue. `dfx` stores the Motoko compiler (`moc`) in the directory returned by [`dfx cache show`](/docs/current/developer-docs/developer-tools/cli-tools/cli-reference/dfx-cache#dfx-cache-show). So the version of Motoko can be shown by running `$(dfx cache show)/moc --version`. This issue only occurs for versions `0.6.20` and older.
```

docs/references/execution-errors.mdx
```diff
@@ -23,4 +24,190 @@ To fix this error, consider:
...
+Growing the canister's memory would require reserving more cycles than allowed by the canister's reserved cycles limit.
+
+An example of this error is:
+```
+  Canister cannot grow memory by 65536 bytes due to its reserved cycles limit. The current limit (5000000000) would be exceeded by 1000000.
+```
+
+When subnets start to run low on memory, using up more memory requires reserving cycles to pay for the future use of that memory. Canisters have a setting that limits the number of cycles that they will reserve and this error indicates that a canister failed to grow its memory because doing so would cause the number of reserved cycles to exceed that limit.
+
+To diagnose this error use the [`dfx canister status`][dfx-canister-status] command or the [`canister_status` API][canister-status-api] to check the canister's current reserved cycles limit. The subnet memory usage can also be seen on the [ICP dashboard](https://dashboard.internetcomputer.org/subnets).
```
jessiemongeon1, line 168:
```suggestion
To diagnose this error use the [`dfx canister status`](/docs/current/developer-docs/developer-tools/cli-tools/cli-reference/dfx-canister#dfx-canister-status) command or the [`canister_status` API](/docs/current/references/ic-interface-spec/#ic-canister_status) to check the canister's current reserved cycles limit. The subnet memory usage can also be seen on the [ICP dashboard](https://dashboard.internetcomputer.org/subnets).
```

docs/references/execution-errors.mdx
```diff
@@ -23,4 +24,190 @@ To fix this error, consider:
...
+The canister does not have enough cycles to grow its memory.
+
+An exmaple of this error is:
+```
+  Canister cannot grow memory by 65536 bytes due to insufficient cycles.
+```
+
+Since canister's need to pay for their memory each round, growing the memory requires that the canister have enough cycles to pay for the increased usage.
+
+To fix this error, top up the canister with more cycles on decrease its [freezing threshold][setting-freezing-threshold].
```
jessiemongeon1, line 155:
```suggestion
To fix this error, top up the canister with more cycles on decrease its [freezing threshold](/docs/current/developer-docs/smart-contracts/topping-up/topping-up-canister/#managing-cycle-depletion-with-freezing-threshold).
```

docs/references/execution-errors.mdx
```diff
@@ -23,4 +24,190 @@ To fix this error, consider:
...
+
+### Memory access limit exceeded
+The amount of data that the canister tried to read or write from stable memory exceeded the limits for a single message.
+
+An example of this error is:
+```
+  Error from Canister xxx-xxx: Canister exceeded memory access limits: Exceeded the limit for the number of modified pages in the stable memory in a single message execution: limit: 8388608 KB.
+```
+
+Although the stable memory of a canister can hold 100s of GiBs of data, each individual message needs to execute within a round and so is limited to reading and writing only a portion of that data. Current limits can be found in the page on [ICP limits][limits].
```
jessiemongeon1, line 140:
```suggestion
Although the stable memory of a canister can hold 100s of GiBs of data, each individual message needs to execute within a round and so is limited to reading and writing only a portion of that data. Current limits can be found in the page on [ICP limits](/docs/current/developer-docs/smart-contracts/maintain/resource-limits/).
```

docs/references/execution-errors.mdx
```diff
@@ -23,4 +24,190 @@ To fix this error, consider:
...
+
+An example of this error is:
+```
+  Canister exceeded its current Wasm memory limit of 2147483648 bytes. The peak Wasm memory usage was 2147485000 bytes. If the canister reaches 4GiB, then it may stop functioning and may become unrecoverable. Please reach out to the canister owner to investigate the reason for the increased memory usage. It might be necessary to move data from the Wasm memory to the stable memory. If such high Wasm memory usage is expected and safe, then the developer can increase the Wasm memory limit in the canister settings.
+```
+
+Canisters may impose limits on the amount of Wasm heap memory they are allowed to use in order to prevent them from using the full 4 GiB. This is desirable because a canister that uses the full 4 GiB may to be able to upgrade if the pre-upgrade hook requires allocating more heap memory.
+
+
+To diagnose this error use the [`dfx canister status`][dfx-canister-status] command or the [`canister_status` API][canister-status-api] to check the canister's Wasm memory limit. If the memory usage is unexpected, try using [canbench][canbench] to determine if there is a memory leak.
```
jessiemongeon1, line 198:
```suggestion
To diagnose this error use the [`dfx canister status`](/docs/current/developer-docs/developer-tools/cli-tools/cli-reference/dfx-canister#dfx-canister-status) command or the [`canister_status` API](/docs/current/references/ic-interface-spec/#ic-canister_status) to check the canister's Wasm memory limit. If the memory usage is unexpected, try using [canbench](/docs/current/developer-docs/smart-contracts/test/benchmarking/) to determine if there is a memory leak.
```

docs/references/execution-errors.mdx
```diff
@@ -23,4 +24,190 @@ To fix this error, consider:
...
+
+[limits]: https://internetcomputer.org/docs/current/developer-docs/smart-contracts/maintain/resource-limits#resource-constraints-and-limits
+
+[canbench]: https://github.com/dfinity/dfxvm?tab=readme-ov-file#installation
+
+[dfx-canister-status]: https://internetcomputer.org/docs/current/developer-docs/developer-tools/cli-tools/cli-reference/dfx-canister#dfx-canister-status
+
+[canister-status-api]: https://internetcomputer.org/docs/current/references/ic-interface-spec#ic-canister_status
+
+[setting-freezing-threshold]: https://internetcomputer.org/docs/current/tutorials/hackathon-prep-course/managing-canisters/#setting-the-canisters-freezing-threshold
```
jessiemongeon1, line 213:
```suggestion

```

jessiemongeon1 (changes requested):

docs/references/execution-errors.mdx
adambratschikaye: Is there a reason we don't want to have the commonly used references at the end? I was thinking it's easier to just change them in one spot if they need to be updated later.

ielashi:
High-level questions:

1. Discovery: how will developers know that this page exists?
2. Maintainability: how do we plan on keeping this page up to date? I can already see that the 8GiB limit being added here will change in a currently open PR. AFAICT, these errors are at the level of the interface spec, as they are practically part of the protocol and changing them would be a breaking change. Would it make sense to at least think about whether these errors should be defined as part of the interface, maintained in a similar rigor as the interface spec, etc?

cc @dsarlis

adambratschikaye:
For discoverability, `Hypervisor` errors will have an attached `doc_link` field introduced [here](https://gitlab.com/dfinity-lab/public/ic/-/merge_requests/18762/diffs#b784be22cdc243cdef9551d74d1c7acbaeec5e22_3_6) which will provide a link to this page.

Hopefully also the errors are unique enough that if a user copies the error into google they get a hit here.

adambratschikaye:
For maintainability, I agree that it's not easy. Maybe some of the limits could be generated by a library coming from the replica code? I could see some of them going into the spec, but I don't know if they all make sense given that, at least in theory, the spec should be for a general protocol and "other implementations" could have other limits.

Also I'm not sure about changes to the errors being breaking changes since they're not returned in a structured way. If someone is parsing the rejections to get details about the errors, it could be argued that they're relying on an unstable interface.

docs/references/execution-errors.mdx
jessiemongeon1: This isn't typically the format/style we use across the docs, so for consistency I recommend we format the links in the same in-line style that the rest of the docs use, but you can leave it as-is if you prefer that.

adambratschikaye:
Oh, and as for including the specific errors in the spec, I wonder if it's realistic to fully specify when each type of error occurs. For example, with the errors about growing memory I guess we would then need to add the notion of the subnet total memory to the spec and then the scheduling algorithm would also need to be included in order to specify which messages succeed and which ones fail when there's a race to allocate the last memory on the subnet.

docs/references/execution-errors.mdx
adambratschikaye: Well if it's up to me then I'd definitely keep it like this. I also like that it's just easier to read and write the raw md with fewer long urls everywhere.

docs/references/execution-errors.mdx
```diff
@@ -23,4 +24,190 @@ To fix this error, consider:
...
+To diagnose this error use the [`dfx canister status`][dfx-canister-status] command or the [`canister_status` API][canister-status-api] to check the canister's Wasm memory limit. If the memory usage is unexpected, try using [canbench][canbench] to determine if there is a memory leak.
+
+To fix this error, consider re-architecting the canister to use stable memory instead of heap memory or sharding the data across multiple canisters. Raising the memory limit may be another option if you are certain the canister can use more memory and still be upgradeable.
+
+
+
+
+[limits]: https://internetcomputer.org/docs/current/developer-docs/smart-contracts/maintain/resource-limits#resource-constraints-and-limits
+
+[canbench]: https://github.com/dfinity/dfxvm?tab=readme-ov-file#installation
```
jessiemongeon1, line 207:
```suggestion
[canbench]: https://internetcomputer.org/docs/current/developer-docs/smart-contracts/test/benchmarking
```

jessiemongeon1 (changes requested):

docs/references/execution-errors.mdx
adambratschikaye: Switched it to a relative link.

jessiemongeon1 (approved):

ulan (approved):
Thanks a lot!

ielashi:
> Some of the errors are implementation specific which means that we probably don't want to include them in the spec. And as Adam also mentioned explaining some of them might require going into implementation details that don't fit in the spec anyway.

I don't disagree with documenting these errors as done in this PR, and I agree that maintaining errors is hard. I don't have a clear answer myself, but I wouldn't necessarily agree that these error messages are implementation details. The north star of the interface spec is that external developers can build their own version of the replica, deploy that implementation, even on the same subnet as our own replica, and they would work. Clearly, they'd need to know exactly what error messages to return in what circumstances to avoid the states from diverging.

I know we're very far away from this goal, and I'm not sure we'll ever reach it, but if a change we make can break canisters (which can certainly happen by changing these messages), then we shouldn't treat these as purely implementation details.

docs/developer-docs/smart-contracts/maintain/resource-limits.mdx
```diff
@@ -46,6 +46,8 @@ The limits depend on the message type as shown in the following table.
 | Wasm custom sections, per canister                                                   | 1MiB        |
 | Wasm custom sections, sections per canister                                          | 16          |
 | Wasm code section, per canister                                                      | 10MiB       |
+| Wasm stable memory, data access per message                                          | 8GiB        |
+| Wasm stable memory, data written per message                                         | 8GiB        |
```
ielashi, line 50: I believe this will change soon if this PR is merged, FYI https://gitlab.com/dfinity-lab/public/ic/-/merge_requests/19014

adambratschikaye:
> The north star of the interface spec is that external developers can build their own version of the replica, deploy that implementation, even on the same subnet as our own replica, and they would work.

Is it really correct that the goal is they should be able to run on the same subnet? In that case we'd need to to specify the exact cycle costs of everything. I thought it was supposed to be more that they could build their own subnet.

ielashi:
> Is it really correct that the goal is they should be able to run on the same subnet? In that case we'd need to to specify the exact cycle costs of everything. I thought it was supposed to be more that they could build their own subnet.

Perhaps that was just my interpretation and I misinterpreted it, but with that interpretation, yes, cycle costs would need to be specified, I agree.

In any case, perhaps this is a bigger discussion than this PR. The documentation you're adding helps for sure, thanks for putting the effort into this.

PR #2890, opened 2024-05-08
add MSQ showcase

seniorjoinu:
[🎉 MSQ Hits The MetaMask Snap Store!](https://forum.dfinity.org/t/msq-hits-the-metamask-snap-store/30499)


seniorjoinu:
@jennifertrin 

How do we make it merge?
It says only people with write access can do that.

PR #2913, opened 2024-05-15
 feat(RUN-972): Add Wasm module validation errors

adambratschikaye:
Add documentation for execution errors that occur when a Wasm module
fails validation.

### DIFF
--- a/docs/references/execution-errors.mdx
+++ b/docs/references/execution-errors.mdx
@@ -309,7 +309,7 @@ An example of this error is:
   Canister's Wasm module is not valid: Wasm module has an invalid export section. The number of exported functions called `canister_update <name>`, `canister_query <name>`, or `canister_composite_query <name>` exceeds 1000.
 ```
 
-There are some [limits][limits] on how canister's Wasm modules can declare methods. To fix this error, consider consolidating the logic of multiple methods into a single method with an additional argument or a enum argument to distinguish which backend code to run. Or consider separating the logic across multiple canisters.
+There are some [limits][limits] on how a canister's Wasm modules can declare methods. To fix this error, consider consolidating the logic of multiple methods into a single method with an additional argument or an enum argument to distinguish which backend code to run. Or, consider separating the logic across multiple canisters.
 
 
 ### Wasm module sum of exported name lengths too large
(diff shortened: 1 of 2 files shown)


ulan (approved):
Thanks!

docs/references/execution-errors.mdx
```diff
@@ -279,13 +279,121 @@ An example of this error is:
...
+
+### Wasm module exports too many methods
+The canister exports more methods than the maximum allowed by ICP.
+
+An example of this error is:
+```
+  Canister's Wasm module is not valid: Wasm module has an invalid export section. The number of exported functions called `canister_update <name>`, `canister_query <name>`, or `canister_composite_query <name>` exceeds 1000.
+```
+
+There are some [limits][limits] on how canister's Wasm modules can declare methods. To fix this error, consider consolidating the logic of multiple methods into a single method with an additional argument or a enum argument to distinguish which backend code to run. Or consider separating the logic across multiple canisters.
```
jessiemongeon1, line 312:
```suggestion
There are some [limits][limits] on how a canister's Wasm modules can declare methods. To fix this error, consider consolidating the logic of multiple methods into a single method with an additional argument or an enum argument to distinguish which backend code to run. Or, consider separating the logic across multiple canisters.
```

docs/references/execution-errors.mdx
```diff
@@ -279,13 +279,121 @@ An example of this error is:
...
+
+### Wasm module function complexity too high
+The canister's Wasm module contains a function which ICP rejects because it may take too long to compile.
+
+An example of this error is:
+```
+  Canister's Wasm module is not valid: Wasm module contains a function at index 7 with complexity 1300000 which exceeds the maximum complexity allowed 1000000.
+```
+
+Certain Wasm instructions (e.g. those involving branching or indirection) may take a long time to compile, so each functions in only allowed to use them a limited number of times. This error indicates that there is a large function exceeding these limits.
```
jessiemongeon1, line 358:
```suggestion
Certain Wasm instructions (e.g. those involving branching or indirection) may take a long time to compile, so each function is only allowed to use them a limited number of times. This error indicates that there is a large function exceeding these limits.
```

jessiemongeon1 (approved):

PR #2943, opened 2024-05-21
use cases overview page

reigj1:


src/pages/use-cases.tsx
```diff
@@ -0,0 +1,241 @@
...
+            variants={transitions.container}
+          >
+            <div className="flex flex-col gap-16 md:gap-40 relative">
+              <TranslatedLayout
+                imageUrl="/img/use-cases/1a.webp"
+                reverse={true}
+              >
+                <div className="blob blob-md md:blob-lg blob-white md:blob-white-dense translate-x-[5%] -translate-y-2/3 z-0 md:opacity-35 " />
+                <h5 className="tw-heading-6 md:tw-heading-5 mb-2">
+                  Chainfusion
```
dprats, line 76: i think this should be "Chain fusion" (two words)

dprats (approved):
I only had one comment (which does not block). I will let you look at it and decide. All else looks good.

PR #3028, opened 2024-06-05
Add info about EVM RPC raw transaction error messages

jessiemongeon1:

  side navigation bar.

### DIFF
--- a/docs/developer-docs/multi-chain/ethereum/evm-rpc/evm-rpc-canister.mdx
+++ b/docs/developer-docs/multi-chain/ethereum/evm-rpc/evm-rpc-canister.mdx
@@ -757,6 +757,8 @@ dfx canister call evm_rpc eth_feeHistory "(variant {$RPC_SOURCE}, $RPC_CONFIG, r
 
 ## Send a raw transaction
 
+The EVM RPC canister can also be used to send raw transactions to the Ethereum and other EVM-compatible chains. Examples for using the EVM RPC canister can be found below, or you can view the [documentation for sending a raw ETH transaction](/docs/current/developer-docs/multi-chain/ethereum/using-eth/submit-transactions).
+
 <AdornedTabs groupId="languages">
 <TabItem value="motoko" label="Motoko" default>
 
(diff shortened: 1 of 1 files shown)


dprats (approved):

docs/developer-docs/multi-chain/ethereum/evm-rpc/evm-rpc-canister.mdx
```diff
@@ -835,6 +837,33 @@ dfx canister call evm_rpc eth_sendRawTransaction "(variant {$RPC_SOURCE}, $RPC_C
...
+
+#### Error "TooFewCycles"
+
+```
+"ProviderError(TooFewCycles { expected: 798336000, received: 307392000 })"
+```
+
+You may receive this error if multiple JSON-RPC providers were queried and one did not receive enough cycles for the request. You can safely attach more cycles to your query, since unused cycles will be refunded.
+
+#### Error "Your app has exceeded its compute units per second capacity"
```
gregorydemay, line 859: I'm not sure I would put this error yet since we are internally discussing whether we could upgrade the current Alchemy tier we are at. WDYT @jasonaibrahim ?

docs/developer-docs/multi-chain/ethereum/evm-rpc/evm-rpc-canister.mdx
```diff
@@ -835,6 +837,33 @@ dfx canister call evm_rpc eth_sendRawTransaction "(variant {$RPC_SOURCE}, $RPC_C
...
+
+### Error messages
+
+#### Error "already known"
+
+```
+"{ code: -32603, message: \"already known\" }"
+```
+
+If you receive this error, but your transaction can be seen on Ethereum, then this error is expected because HTTPS outcalls are replicated meaning only one of the contacted nodes from the JSON-RPC provide should return the success results, while others return an error indicating the transaction is already known.
```
gregorydemay, line 849: I think this error needs a bit more context because it's specific to `eth_sendRawTransaction`, so I'm not sure putting it in a general `Error messages` section is the best place. Generally I think it should mention the following points:

Error when sending a transaction:

1. Sending a transaction to the Ethereum network is a state changing operation and since the EVM-RPC canister sends a transaction to some JSON-RPC providers via HTTPs outcalls , each contacted provider will receive multiple times the same transaction.
2. It is therefore possible that the contacted provider will return an error such as `{ code: -32603, message: "already known" }`, since assuming the sent transaction is valid, one node of the contacted provider should accept the transaction, while the other ones should return some error. Note that the exact error returned is EVM-client specific, since unfortunately the Ethereum JSON-RPC API doesn't specify errors, meaning that other errors could occur or even an error indicating that consensus on the IC could not be reached.
3. Note that even in case an error is returned, you SHOULD assume that your transaction is known to the Ethereum network. You can control whether your transaction was indeed included in a block by querying the transaction count (`eth_getTransactionCount`) at `Latest` block height.

@rvanasa : feel free to expand if you think I forgot something.

jessiemongeon1: Follow up PR here: https://github.com/dfinity/portal/pull/3038

PR #3075, opened 2024-06-17
fix https references

nathanosdev:
HTTPS doesn't sound right before words like request, response, server etc... I've updated the following to use HTTP instead of HTTPS:
- HTTPS GET
- HTTPS PUT
- HTTPS response
- HTTPS header
- HTTPS server
- HTTPS call
- HTTPS API
- HTTPS principal
- HTTPS 2.0
- HTTPS services
- HTTPS method
- HTTPS proxy
- HTTPS gateway

The main references that have not changed are:
- HTTPS outcall
- HTTPS interface

They refer to IC concepts that are specifically named with HTTPS.

  side navigation bar.


jessiemongeon1 (approved):

jessiemongeon1:
There are some instances where HTTP and HTTPS are used in the same sentence, do we assume that developers understand that most HTTP traffic is served over HTTPS, and we are using HTTP as a style choice? 

I recently updated all of these docs to use HTTPS instead of HTTP, since most traffic is served over HTTPS instead of plain HTTP, so HTTPS is more accurate to say. I think it is fine to revert those changes I made, but want to confirm we are being accurate and understood by developers.

nathanosdev:
> I recently updated all of these docs to use HTTPS instead of HTTP, since most traffic is served over HTTPS instead of plain HTTP, so HTTPS is more accurate to say.

I see this as mixing a few different concepts together.

There is the base version of HTTP, which defines a lot of things, including the shape/format of requests and responses. Normally developers are not discussing encrypted / unencrypted connections when discussing format, since that part of the HTTPS protocol is fairly opaque to most developers, unless they're doing very low level networking stuff.

The same goes for the version of the HTTP standard. Developers normally don't say "HTTP 2.0 request", they  just say "HTTP request". An HTTP request is always an HTTP request regardless of whether it's served over an encrypted connection or not, or whether it's a v1.0, v1.1, or v2.0 request. Otherwise there'd be a huge mix of HTTP v1.0 / HTTPS v1.0 / HTTP v1.1, etc...

HTTPS is an _extension_ of HTTP that introduces encryption on top of the existing HTTP standard. Likewise HTTP 2.0 and 1.1 add new things on top of HTTP, without fundamentally changing the HTTP protocol.

In the end, saying HTTPS or HTTP request doesn't affect how accurate it is, but HTTPS request feels awkward to me as a developer.

PR #3091, opened 2024-06-19
feat: Add security best practices for certified variables

venkkatesh-sekar:

  side navigation bar.


docs/developer-docs/security/security-best-practices/data-integrity-and-authenticity.mdx
roelstorms, line 1: The overview of the query and update calls is really useful but it does reiterate [this info](https://f7i7o-aaaaa-aaaam-abtba-cai.icp0.io/docs/current/developer-docs/smart-contracts/call/overview/#update-calls). Should we just reference the other page to avoid duplication and possibly outdated info?

venkkatesh-sekar: Good point, I didn't know we had this section already. I will reference it back to the page.

docs/developer-docs/security/security-best-practices/data-integrity-and-authenticity.mdx
```diff
@@ -8,19 +8,334 @@ import { MarkdownChipRow } from "/src/components/Chip/MarkdownChipRow";
...
+
+:::tip
+ICP also provides Replica signed queries, where query responses are signed by the answering replica node but however it doesn't have the same security guarantees as `update` calls. Replica signed queries are enabled by default on both `agent-rs` and `agent-js`. You can read more about the feature [here](https://internetcomputer.org/blog/features/replica-signed-queries)
```
roelstorms, line 16:
```suggestion
ICP also provides Replica signed queries, where query responses are signed by the answering replica node but it doesn't have the same security guarantees as `update` calls. Replica signed queries are enabled by default on both `agent-rs` and `agent-js`. You can read more about the feature [here](https://internetcomputer.org/blog/features/replica-signed-queries)
```

docs/developer-docs/security/security-best-practices/data-integrity-and-authenticity.mdx
```diff
@@ -8,19 +8,334 @@ import { MarkdownChipRow } from "/src/components/Chip/MarkdownChipRow";
...
+:::
+
+### What's Certified data?
+* A canister can store a small amount of data (32 bytes) during an update call in the state tree under `certified_data` which will be certified by the subnet.
+* During query call, the canister can obtain a certificate about that data, which can be verified by the IC root public key.
```
roelstorms, line 21:
```suggestion
* During a query call, the canister can obtain a certificate about that data, which can be verified by the IC root public key.
```

docs/developer-docs/security/security-best-practices/data-integrity-and-authenticity.mdx
```diff
@@ -8,19 +8,334 @@ import { MarkdownChipRow } from "/src/components/Chip/MarkdownChipRow";
...
+
+In the rest of the section, we will see an example of how to design a canister, which can serve a certified response for a `query` using `certified_data` and how to verify the response. The examples are written in Rust but can be imported to other languages following the same recommendations.
```
roelstorms, line 23:
```suggestion
The rest of the section shows an example canister, which can serve a certified response for a `query` using `certified_data` which is verified in the frond-end. The examples are written in Rust but the overall design can be implemented in other languages.
```
Imported implies this is automatic.

docs/developer-docs/security/security-best-practices/data-integrity-and-authenticity.mdx
```diff
@@ -8,19 +8,334 @@ import { MarkdownChipRow } from "/src/components/Chip/MarkdownChipRow";
...
+
+### Building a canister with Certified Variables
+Let's consider the following canister,
```
roelstorms, line 26:
```suggestion
Let's consider the following canister interface:
```

docs/developer-docs/security/security-best-practices/data-integrity-and-authenticity.mdx
```diff
@@ -8,19 +8,334 @@ import { MarkdownChipRow } from "/src/components/Chip/MarkdownChipRow";
...
+    witness : blob;
+};
+
+service : {
+    "set_user": (User) -> (nat64);
+    "get_user": (nat64) -> (CertifiedUser) query;
+}
+```
+
+The canister exposes the following service,
```
roelstorms, line 46:
```suggestion
The canister exposes the following service:
```

docs/developer-docs/security/security-best-practices/data-integrity-and-authenticity.mdx
```diff
@@ -8,19 +8,334 @@ import { MarkdownChipRow } from "/src/components/Chip/MarkdownChipRow";
...
+service : {
+    "set_user": (User) -> (nat64);
+    "get_user": (nat64) -> (CertifiedUser) query;
+}
+```
+
+The canister exposes the following service, 
+* **set_user** - The caller provides a `User` object to the canister. The canister records it and serves a corresponding `index` for the entry as the response. Since `certified_data` can only store 32 bytes of data, we use a specialized data strucutre from `ic_certified_map` to store the `User` data. 
+  * The data strucutre internally stores the data in a `HashTree` and we record the `root_hash` of the data structure in the `certified_data`, which is 32 bytes. 
+  * The `root_hash` cryptographically guarantees that only one tree can correspond to that hash. You can read more about it [here](https://docs.rs/ic-certified-map/latest/ic_certified_map/struct.RbTree.html).
```
roelstorms, line 49: The referenced documentation doesn't really explain HashTrees. Is there a better source for people to understand how a 32 bytes hash can certify an endless datastructure?

docs/developer-docs/security/security-best-practices/data-integrity-and-authenticity.mdx
```diff
@@ -8,19 +8,334 @@ import { MarkdownChipRow } from "/src/components/Chip/MarkdownChipRow";
...
+* **get_user** - The caller provides a `index: nat64` to the canister and gets a Certified response for the corresponding `User`. The CertifiedUser response must have the following strucutre for verifying the response.
```
roelstorms, line 50:
```suggestion
* **get_user** - The caller provides an `index: nat64` to the canister and gets a certified response for the corresponding `User`. The `CertifiedUser` response must have the following structure to allow verifying the response:
```

docs/developer-docs/security/security-best-practices/data-integrity-and-authenticity.mdx
venkkatesh-sekar: I don't think we have anything explaining Merkle Trees.  I can reword it a bit and point to Merkle trees Wiki page.

docs/developer-docs/security/security-best-practices/data-integrity-and-authenticity.mdx
```diff
@@ -8,19 +8,334 @@ import { MarkdownChipRow } from "/src/components/Chip/MarkdownChipRow";
...
+            certificate,
+            witness,
+        }
+    })
+}
+```
+
+### Verifying the Certified Variables
+
+Once you have the response `CertifiedUser`, for the secureness guarantee, one must verify the certification in the response. This is broken down into several steps with an example below implementing these steps in Rust.
```
roelstorms, line 153:
```suggestion
Once you have the response `CertifiedUser`, for the integrity guarantee, the front-end must verify the certification in the response. This is broken down into several steps implemented in the Rust example below.
```

docs/developer-docs/security/security-best-practices/data-integrity-and-authenticity.mdx
```diff
@@ -8,19 +8,334 @@ import { MarkdownChipRow } from "/src/components/Chip/MarkdownChipRow";
...
+
+:::note
+The example has some extra steps to setup the canister with some `User` data before verification. You can ignore the section marked between `// ==== START of canister data setup` and `// ==== END of canister data setup`
+:::
+
+1. Recompute the `root_hash` of `certificate.tree` (pruned state tree with the canister's `certified_data`) and verify the `certificate.signature` with `root_hash` as message and `certificate.delegation` and IC root_key as the public key. This confirms that the signature is valid for the current state tree. 
+2. Validate that the response is not stale by verifying the time at `/time` in `certificate.tree` is less than a certain delta of currrent time. The recommended delta is 5 minutes.
```
roelstorms, line 160:
```suggestion
2. Validate that the response is not stale by verifying the time at `/time` in `certificate.tree` is less than a certain delta of current time. The recommended delta is 5 minutes but should be adapted to the use case.
```
I'm wondering if we have a concrete example where the /time would have to be more recent than 5 minutes. Maybe some sort of DeFi app where price information determines the default order price which could trick users into using a 5 min old value that would make them overpay on a market order.

docs/developer-docs/security/security-best-practices/data-integrity-and-authenticity.mdx
```diff
@@ -8,19 +8,334 @@ import { MarkdownChipRow } from "/src/components/Chip/MarkdownChipRow";
...
+                result
+                    .expect("Query call get_user failed")
+                    .as_slice(),
+                u64
+            )
+            .unwrap()
+        })
+        .collect();
+
+    // From response indexes, chose a random index for get_user
```
roelstorms, line 251:
```suggestion
    // From response indexes, choose a random index for get_user
```

docs/developer-docs/security/security-best-practices/data-integrity-and-authenticity.mdx
```diff
@@ -8,19 +8,334 @@ import { MarkdownChipRow } from "/src/components/Chip/MarkdownChipRow";
...
+    let mut deserializer = Deserializer::from_slice(&certified_user.certificate);
+    let certificate: Certificate = serde::de::Deserialize::deserialize(&mut deserializer).unwrap();
+
+    let start = SystemTime::now();
+    let current_time = start
+        .duration_since(UNIX_EPOCH)
+        .expect("Time went backwards")
+        .as_nanos();
+
+    // Step 1: Check if signature in the certificate can be validated with the root_hash of the tree in certificate as message and root_key as public_key
```
roelstorms, line 277:
```suggestion
    // Step 1: Check if signature in the certificate can be validated with the 
    // root_hash of the tree in certificate as message and root_key as public_key
```
Split into two lines for readability.

docs/developer-docs/security/security-best-practices/data-integrity-and-authenticity.mdx
```diff
@@ -8,19 +8,334 @@ import { MarkdownChipRow } from "/src/components/Chip/MarkdownChipRow";
...
+        _ => panic!("Certified data not found"),
+    };
+
+    let mut deserializer = Deserializer::from_slice(&certified_user.witness);
+    let witness_decoded: HashTree<Vec<u8>> =
+        serde::de::Deserialize::deserialize(&mut deserializer).unwrap();
+    let witness_digest = witness_decoded.digest();
+
+    println!(
+        "Step 3: Witness digest mataches certified data: {:?} ",
```
roelstorms, line 308:
```suggestion
        "Step 3: Witness digest matches certified data: {:?} ",
```

docs/developer-docs/security/security-best-practices/data-integrity-and-authenticity.mdx
```diff
@@ -8,19 +8,334 @@ import { MarkdownChipRow } from "/src/components/Chip/MarkdownChipRow";
 
 <MarkdownChipRow labels={["Intermediate", "Security", "Concept"]} />
```
roelstorms, line 9: Not sure if this is still intermediate.

docs/developer-docs/security/security-best-practices/data-integrity-and-authenticity.mdx
venkkatesh-sekar: We could add one when we add a best practice about system time. Right now, even agents enforce 5m and not sure if it's adaptable.

docs/developer-docs/security/security-best-practices/data-integrity-and-authenticity.mdx
venkkatesh-sekar: I haven't use agent-js much to know if all of this can be done in a similar way. But I agree, having it in JS/TS would be best useful for the community. I can do a follow up PR sometime later to add the JS variant.

docs/developer-docs/security/security-best-practices/data-integrity-and-authenticity.mdx
```diff
@@ -8,19 +8,334 @@ import { MarkdownChipRow } from "/src/components/Chip/MarkdownChipRow";
...
+
+:::tip
+ICP also provides Replica signed queries, where query responses are signed by the answering replica node but however it doesn't have the same security guarantees as `update` calls. Replica signed queries are enabled by default on both `agent-rs` and `agent-js`. You can read more about the feature [here](https://internetcomputer.org/blog/features/replica-signed-queries)
+:::
+
+### What's Certified data?
```
roelstorms, line 19: I would elaborate on this a bit more. Maybe using a diagram that includes the data struct, the hash, the signature/certificate and where it is updated/created (update), retrieved (query) and validated (client).

I'm also missing the drawbacks/caveats of certified data. For example, this doesn't work if your query call returns a calculation result since only update calls can update the certificate. The workaround it to give all the certified source data to the client which can check the certificate and make the same calculations or to have certain pre calculated values certified in update calls.

docs/developer-docs/security/security-best-practices/data-integrity-and-authenticity.mdx
roelstorms: We could add a short summary along the lines of:

> As explain in REF, update calls are slow and expensive but query call results can be manipulated by a replica or boundary node. In certain use cases, there is a third option whereby query results can return data that has been certified by the subnet in an earlier update call. This is the concept of certified data and it requires changed to the update call to create the certification, query call to return the certificate and fron-end to verify the certificate.

roelstorms:
@venkkatesh-sekar do you have the example code in a repo and did you test it?
It's always nice to have example code also in Motoko and the FE in JS/TS. It would be great to have both examples in all our best practices using [this](https://docusaurus.io/docs/markdown-features/code-blocks#multi-language-support-code-blocks) Docusaurus block.

docs/developer-docs/security/security-best-practices/data-integrity-and-authenticity.mdx
jessiemongeon1: I would classify this as 'Advanced'

docs/developer-docs/security/security-best-practices/data-integrity-and-authenticity.mdx
```diff
@@ -8,19 +8,324 @@ import { MarkdownChipRow } from "/src/components/Chip/MarkdownChipRow";
 
 <MarkdownChipRow labels={["Intermediate", "Security", "Concept"]} />
 
-## Certify query responses if they are relevant for security
+## Using Certified Variables for secure queries
```
jessiemongeon1, line 11:
```suggestion
## Using certified variables for secure queries 
```

docs/developer-docs/security/security-best-practices/data-integrity-and-authenticity.mdx
```diff
@@ -8,19 +8,324 @@ import { MarkdownChipRow } from "/src/components/Chip/MarkdownChipRow";
...
...site_query` under queries for the rest of the section. A detailed overview between `update` and `query` calls can be found [here](https://internetcomputer.org/docs/current/developer-docs/smart-contracts/call/overview/#introduction). In general, queries have fast response time but lack integrity guarantees, which makes them unfit for security critical applications. To obtain query like response times and update like certified responses, developers could make use of `certified_data` in the state tree to certify the response served in a query. Alternatively, these calls could be issued as update calls by the caller (e.g. in agent-js), but that impacts performance by taking a few seconds. Some examples of Certified Variables are asset certification in [Internet Identity](https://github.com/dfinity/internet-identity/blob/b29a6f68bbe5a49d048e12bc7a3263a9f43d080b/src/internet_identity/src/main.rs#L775-L808), [NNS dapp](https://github.com/dfinity/nns-dapp/blob/372c3562127d70c2fde059bc9c268e8ae858583e/rs/src/assets.rs#L121-L145), or the [canister signature implementation in Internet Identity](https://github.com/dfinity/internet-identity/blob/main/src/internet_identity/src/signature_map.rs).
```
jessiemongeon1, line 13:
```suggestion
ICP offers three modes of operations for canisters, `update`, `query` and `composite_query`. For the sake of simplicity, we will club `composite_query` under queries for the rest of this section. For more information, view the  [detailed overview between `update` and `query` calls](https://internetcomputer.org/docs/current/developer-docs/smart-contracts/call/overview/#introduction). 

In general, queries have fast response time but lack integrity guarantees, which makes them unfit for security critical applications. To obtain query-like response times and update-like certified responses, developers can make use of `certified_data` in the state tree to certify the response served in a query. Alternatively, these calls could be issued as update calls by the caller (e.g. in agent-js), but that impacts performance by taking a few seconds. 

Some examples of certified variables are asset certification in [Internet Identity](https://github.com/dfinity/internet-identity/blob/b29a6f68bbe5a49d048e12bc7a3263a9f43d080b/src/internet_identity/src/main.rs#L775-L808), [NNS dapp](https://github.com/dfinity/nns-dapp/blob/372c3562127d70c2fde059bc9c268e8ae858583e/rs/src/assets.rs#L121-L145), or the [canister signature implementation in Internet Identity](https://github.com/dfinity/internet-identity/blob/main/src/internet_identity/src/signature_map.rs).
```

docs/developer-docs/security/security-best-practices/data-integrity-and-authenticity.mdx
```diff
@@ -8,19 +8,324 @@ import { MarkdownChipRow } from "/src/components/Chip/MarkdownChipRow";
...
 
-The responses to [query calls](/docs/current/references/ic-interface-spec#http-interface) are not threshold-signed by the canister and subnet. Thus, a single malicious replica or boundary node may change the data, violating its authenticity. This is especially risky if update calls, which do returned signed responses, depend on the response to query calls.
+:::tip
+ICP also provides Replica signed queries, where query responses are signed by the answering replica node but however it doesn't have the same security guarantees as `update` calls. Replica signed queries are enabled by default on both `agent-rs` and `agent-js`. You can read more about the feature [here](https://internetcomputer.org/blog/features/replica-signed-queries)
```
jessiemongeon1, line 16:
```suggestion
ICP also provides replica signed queries, where query responses are signed by the answering replica node, however it doesn't have the same security guarantees an `update` call. Replica signed queries are enabled by default on both `agent-rs` and `agent-js`. [Read more about the feature](https://internetcomputer.org/blog/features/replica-signed-queries).
```

docs/developer-docs/security/security-best-practices/data-integrity-and-authenticity.mdx
```diff
@@ -8,19 +8,324 @@ import { MarkdownChipRow } from "/src/components/Chip/MarkdownChipRow";
...
+:::
 
-### Recommendation
+### What's Certified data?
```
jessiemongeon1, line 19:
```suggestion
### What is certified data?
```

docs/developer-docs/security/security-best-practices/data-integrity-and-authenticity.mdx
```diff
@@ -8,19 +8,324 @@ import { MarkdownChipRow } from "/src/components/Chip/MarkdownChipRow";
...
+* A canister can store a small amount of data (32 bytes) during an update call in the state tree under `certified_data` which will be certified by the subnet.
```
jessiemongeon1, line 20: Why is this info formatted as bullet points? Could this be reformatted as a normal paragraph?

docs/developer-docs/security/security-best-practices/data-integrity-and-authenticity.mdx
```diff
@@ -8,19 +8,324 @@ import { MarkdownChipRow } from "/src/components/Chip/MarkdownChipRow";
...
+* During query call, the canister can obtain a certificate about that data, which can be verified by the IC root public key.
+
+In the rest of the section, we will see an example of how to design a canister, which can serve a certified response for a `query` using `certified_data` and how to verify the response. The examples are written in Rust but can be imported to other languages following the same recommendations.
+
+### Building a canister with Certified Variables
```
jessiemongeon1, line 25:
```suggestion
### Building a canister with certified variables
```

docs/developer-docs/security/security-best-practices/data-integrity-and-authenticity.mdx
```diff
@@ -8,19 +8,324 @@ import { MarkdownChipRow } from "/src/components/Chip/MarkdownChipRow";
...
+};
+
+service : {
+    "set_user": (User) -> (nat64);
+    "get_user": (nat64) -> (CertifiedUser) query;
+}
+```
+
+The canister exposes the following service, 
+* **set_user** - The caller provides a `User` object to the canister. The canister records it and serves a corresponding `index` for the entry as the response. Since `certified_data` can only store 32 bytes of data, we use a specialized data strucutre from `ic_certified_map` to store the `User` data.
```
jessiemongeon1, line 47:
```suggestion
* **set_user**: The caller provides a `User` object to the canister. The canister records it and serves a corresponding `index` for the entry as the response. Since `certified_data` can only store 32 bytes of data, it uses a specialized data structure from `ic_certified_map` to store the `User` data. 
```

docs/developer-docs/security/security-best-practices/data-integrity-and-authenticity.mdx
```diff
@@ -8,19 +8,324 @@ import { MarkdownChipRow } from "/src/components/Chip/MarkdownChipRow";
...
+
+service : {
+    "set_user": (User) -> (nat64);
+    "get_user": (nat64) -> (CertifiedUser) query;
+}
+```
+
+The canister exposes the following service, 
+* **set_user** - The caller provides a `User` object to the canister. The canister records it and serves a corresponding `index` for the entry as the response. Since `certified_data` can only store 32 bytes of data, we use a specialized data strucutre from `ic_certified_map` to store the `User` data. 
+  * The data strucutre internally stores the data in a `HashTree` and we record the `root_hash` of the data structure in the `certified_data`, which is 32 bytes.
```
jessiemongeon1, line 48:
```suggestion
  * The data structure internally stores the data in a `HashTree` and records the `root_hash` of the data structure in the `certified_data`, which is 32 bytes. 
```

docs/developer-docs/security/security-best-practices/data-integrity-and-authenticity.mdx
```diff
@@ -8,19 +8,324 @@ import { MarkdownChipRow } from "/src/components/Chip/MarkdownChipRow";
...
+service : {
+    "set_user": (User) -> (nat64);
+    "get_user": (nat64) -> (CertifiedUser) query;
+}
+```
+
+The canister exposes the following service, 
+* **set_user** - The caller provides a `User` object to the canister. The canister records it and serves a corresponding `index` for the entry as the response. Since `certified_data` can only store 32 bytes of data, we use a specialized data strucutre from `ic_certified_map` to store the `User` data. 
+  * The data strucutre internally stores the data in a `HashTree` and we record the `root_hash` of the data structure in the `certified_data`, which is 32 bytes. 
+  * The `root_hash` cryptographically guarantees that only one tree can correspond to that hash. You can read more about it [here](https://docs.rs/ic-certified-map/latest/ic_certified_map/struct.RbTree.html).
```
jessiemongeon1, line 49:
```suggestion
  * The `root_hash` cryptographically guarantees that only one [tree can correspond to that hash](https://docs.rs/ic-certified-map/latest/ic_certified_map/struct.RbTree.html).
```

docs/developer-docs/security/security-best-practices/data-integrity-and-authenticity.mdx
```diff
@@ -8,19 +8,324 @@ import { MarkdownChipRow } from "/src/components/Chip/MarkdownChipRow";
...
+* **get_user** - The caller provides a `index: nat64` to the canister and gets a Certified response for the corresponding `User`. The CertifiedUser response must have the following strucutre for verifying the response.
```
jessiemongeon1, line 50:
```suggestion
* **get_user**: The caller provides a `index: nat64` to the canister and gets a certified response for the corresponding `User`. The `CertifiedUser` response must have the following structure for verifying the response:
```

docs/developer-docs/security/security-best-practices/data-integrity-and-authenticity.mdx
```diff
@@ -8,19 +8,324 @@ import { MarkdownChipRow } from "/src/components/Chip/MarkdownChipRow";
...
+  * **user** - The actual response
+  * **certificate** - The payload for verifying the signature on the `certified_data`. ICP provides the system API `data_certificate()` for this 
+  * **witness** - The witness allows us to complete the final verification of the response with the requested input and `certified_data`.
```
jessiemongeon1, line 53:
```suggestion
  * **user**: The actual response.
  * **certificate**: The payload for verifying the signature on the `certified_data`. ICP provides the system API `data_certificate()` for this.
  * **witness**: Allows for the final verification of the response to be completed with the requested input and `certified_data`.
```

docs/developer-docs/security/security-best-practices/data-integrity-and-authenticity.mdx
```diff
@@ -8,19 +8,324 @@ import { MarkdownChipRow } from "/src/components/Chip/MarkdownChipRow";
...
+        CertifiedUser {
+            user,
+            certificate,
+            witness,
+        }
+    })
+}
+```
+
+### Verifying the Certified Variables
```
jessiemongeon1, line 151:
```suggestion
### Verifying certified variables
```

docs/developer-docs/security/security-best-practices/data-integrity-and-authenticity.mdx
```diff
@@ -8,19 +8,324 @@ import { MarkdownChipRow } from "/src/components/Chip/MarkdownChipRow";
...
+
+Once you have the response `CertifiedUser`, for the secureness guarantee, one must verify the certification in the response. This is broken down into several steps with an example below implementing these steps in Rust.
+
+:::note
+The example has some extra steps to setup the canister with some `User` data before verification. You can ignore the section marked between `// ==== START of canister data setup` and `// ==== END of canister data setup`
+:::
+
+1. Recompute the `root_hash` of `certificate.tree` (pruned state tree with the canister's `certified_data`) and verify the `certificate.signature` with `root_hash` as message and `certificate.delegation` and IC root_key as the public key. This confirms that the signature is valid for the current state tree.
```
jessiemongeon1, line 159:
```suggestion
1. Recompute the `root_hash` of `certificate.tree` (pruned state tree with the canister's `certified_data`) and verify the `certificate.signature` with `root_hash` as the message,`certificate.delegation`, and the IC `root_key` as the public key. This confirms that the signature is valid for the current state tree. 
```

docs/developer-docs/security/security-best-practices/data-integrity-and-authenticity.mdx
```diff
@@ -8,19 +8,324 @@ import { MarkdownChipRow } from "/src/components/Chip/MarkdownChipRow";
...
+:::
+
+1. Recompute the `root_hash` of `certificate.tree` (pruned state tree with the canister's `certified_data`) and verify the `certificate.signature` with `root_hash` as message and `certificate.delegation` and IC root_key as the public key. This confirms that the signature is valid for the current state tree. 
+2. Validate that the response is not stale by verifying the time at `/time` in `certificate.tree` is less than a certain delta of currrent time. The recommended delta is 5 minutes.
+3. Recompute the `root_hash` of witness and verify equality with the `certified_data`. The `certified_data` can be obtained from `certificate.tree` under the path `/canister/<canister_id>/certified_data`
+4. Check if query parameters are in the witness. In our example, the lookup path is `/user/<index>` and it should be present in the witness. 
+5. Validate if the value found in `/user/<index>` matches `user` from the response. 
+6. If all of previous steps succceed, return `user` as the valid response.
```
jessiemongeon1, line 164:
```suggestion
3. Recompute the `root_hash` of the witness and verify equality with the `certified_data`. The `certified_data` can be obtained from `certificate.tree` under the path `/canister/<canister_id>/certified_data`.
4. Check if query parameters are in the witness. In this example, the lookup path is `/user/<index>` and should be present in the witness. 
5. Validate if the value found in `/user/<index>` matches `user` from the response. 
6. If all of the previous steps succeed, return `user` as the valid response.
```

docs/developer-docs/security/security-best-practices/data-integrity-and-authenticity.mdx
venkkatesh-sekar: I think this is already explained in Certified Variables section and I have referenced the link.

venkkatesh-sekar:
> @venkkatesh-sekar do you have the example code in a repo and did you test it? It's always nice to have example code also in Motoko and the FE in JS/TS. It would be great to have both examples in all our best practices using [this](https://docusaurus.io/docs/markdown-features/code-blocks#multi-language-support-code-blocks) Docusaurus block.
> 
> @robin-kunzler WDYT? Is it worth the time investment?

@roelstorms I have added the JS and Motoko example as well. The code is here (https://github.com/Spockuto/certified_variables). It's on my personal github, so not sure where we want to move to reference it.

docs/developer-docs/security/security-best-practices/data-integrity-and-authenticity.mdx
roelstorms: Maybe [this](https://f7i7o-aaaaa-aaaam-abtba-cai.icp0.io/how-it-works/response-certification/#certified-variables) is also a good resource. There is even a 30min [YouTube video](https://www.youtube.com/watch?v=3mZHEfICi_U) we could link.

venkkatesh-sekar: Added a section with a high level explanation of certified variables.

robin-kunzler:
Thanks a lot @venkkatesh-sekar for explaining the concept of certified variables in so much detail!

I'm adding a high level comment here, leaving the detailed review to @roelstorms . 

Would it make sense to stress the following points more explicitly in the first section? 
* You now say "but query call results can be manipulated by a replica or boundary node". I'm worried this does not sufficiently point out the security concern that we see over and over again: People just use query calls without thinking about authenticity of data. I think we should really stress that it can be a significant security issue if people just use data returned from queries. We could even add an example (e.g. displaying wrong proposal data and make people vote for the wrong thing). Can we advise for what sort of data they should be careful about the authenticity? I think this is the core issue here and right now to me it seems this got a bit lost. WDYT? 
* Explain (as you are doing already) that update calls solve the problem, but are slow. If that is too slow, one can consider certified variables. 
* I'm also a bit worried that this best practice might now read like: "Use certified variables whenever you can, they are great, everybody should do it". However, I'm not sure that is a good recommendation because this bears significant complexity and it is easy to get things wrong and introduce bugs. Can we make it clearer in what situations and for who we recommend using certified data? IMO people should use update calls and certified data is something for experts, only if needed if performance really requires it? 

Finally, I always used the "security concern" title to explain the security issue at hand. It looks like this could work for the first section here as well? I like highlighting the security concern very explicitly, but I don't have a super strong opinion, we can also discuss if we want to stick to that structure. There are already a few best practices that don't follow this and if you prefer a different title for the first section I don't object. Maybe @roelstorms also has an opinion or we should align on the structure for future best practices.

docs/developer-docs/security/security-best-practices/data-integrity-and-authenticity.mdx
```diff
@@ -12,16 +12,25 @@ import { MarkdownChipRow } from "/src/components/Chip/MarkdownChipRow";
 
 <MarkdownChipRow labels={["Advanced", "Security", "Concept"]} />
 
-## Using certified variables for secure queries 
+## Certified variables
 
-ICP offers three modes of operations for canisters, `update`, `query` and `composite_query`. For the sake of simplicity, we will club `composite_query` under queries for the rest of this section. For more information, view the  [detailed overview between `update` and `query` calls](https://internetcomputer.org/docs/current/developer-docs/smart-contracts/call/overview/#introduction).
+### Security concern
+
+ICP offers three modes of operations for canisters, `update`, `query` and `composite_query`. For the sake of simplicity, we will club `composite_query` under queries for the rest of this section. For more information, view the  [detailed overview between `update` and `query` calls](https://internetcomputer.org/docs/current/developer-docs/smart-contracts/call/overview/#introduction). As explained in the overview, update calls are slow and expensive but provide integrity guarantee as they are responded with a threshold signature signed by the subnet.
```
roelstorms, line 19:
```suggestion
ICP offers three modes of operations for canisters, `update`, `query` and `composite_query`. For the sake of simplicity, we will club `composite_query` under queries for the rest of this section. For more information, view the  [detailed overview between `update` and `query` calls](https://internetcomputer.org/docs/current/developer-docs/smart-contracts/call/overview/#introduction). As explained in the overview, update calls are slow and expensive but provide integrity guarantees as their response include a threshold signature signed by the subnet. 
```

docs/developer-docs/security/security-best-practices/data-integrity-and-authenticity.mdx
```diff
@@ -12,16 +12,25 @@ import { MarkdownChipRow } from "/src/components/Chip/MarkdownChipRow";
...
+
+On the other hand, query calls are fast since they are responded by a single replica but **there is no integrity guarantee, since the response can be manipulated if the responding replica or the boundary node is malicious.** For example, if the NNS dapp fetches proposal information from the governance canister via query calls and the responding node is malicious, it can mask an ill intentioned proposal that causes irrevocable damage as innocuous by modifying the proposal payload in the response and mislead voters into voting yes. Another consequence of query calls is that users can't rely on [canister_inspect_message](https://internetcomputer.org/docs/current/developer-docs/security/security-best-practices/iam#do-not-rely-on-ingress-message-inspection) as a guard. **This makes query calls, in it's raw form, unfit to serve data for security critical applications.**
```
roelstorms, line 21:
```suggestion
On the other hand, query calls are fast since a single replica formulates the response but **there is no integrity guarantee, since the response can be manipulated if the responding replica or the boundary node is malicious.** For example, if the NNS dapp fetches proposal information from the governance canister via query calls and the responding node is malicious, it can mask an ill intentioned proposal that causes irrevocable damage as innocuous by modifying the proposal payload in the response and mislead voters into voting yes. Another consequence of query calls is that users can't rely on [canister_inspect_message](https://internetcomputer.org/docs/current/developer-docs/security/security-best-practices/iam#do-not-rely-on-ingress-message-inspection) as a guard. **This makes query calls, in it's raw form, unfit to serve data for security critical applications.**
```

Inspect message should never be relied upon for anything but DoS protection. I'm not sure if your phrasing suggests otherwise. Do you mean inspect message only runs on update calls or that query calls have no other method for access control?

docs/developer-docs/security/security-best-practices/data-integrity-and-authenticity.mdx
```diff
@@ -2,25 +2,628 @@
...
+:::
+
+:::tip
+ICP also provides replica signed queries, where query responses are signed by the answering replica node, however it doesn't have the same security guarantees an `update` call and only protects from malicious boundary nodes. Replica signed queries are enabled by default on both `agent-rs` and `agent-js`. [Read more about the feature](https://internetcomputer.org/blog/features/replica-signed-queries).
+:::
+
+### What is certified data?
+A canister can store a small amount of data (32 bytes) during an update call in the state tree under `certified_data` which will be certified by the subnet. During a query call, the canister can obtain a certificate about that data, which can be verified against the IC root public key. Combining these features, a canister should be able to certify a variable in a query response with this [design](https://medium.com/dfinity/how-internet-computer-responses-are-certified-as-authentic-2ff1bb1ea659).
```
roelstorms, line 37: I've written an alternative introduction. It also explains the issue with calculations in query calls. WDYT?

> For a subnet to certify data, it has to create a signature over all certified data during an update call. Each update call that modifies the data needs to recreate this signature. However, a canister won't simply sign an arbitrarily large amount of data in complex data structure. It is limited to signing just 32 bytes which will end up in the subnet state tree under `certified_data`. Since this is rarely enough, it is up to the developer to reduce the data into this 32 bytes. Naturally, this can be done by computing a hash of the data structure. This is also why certified data is difficult to implement. Depending on your data structures, you will need to develop a different kind of hashing function.
> 
> Subsequent query calls can return the data as-is, including the signature which the front-end can verify. This means that data aggregation or other calculations can't be done in query calls as there would be no way to produce a signature over that newly created data. There are two workarounds, either this data is precomputed in the update call or all raw data is sent to the front-end which verifies it and does the calculations.

docs/developer-docs/security/security-best-practices/data-integrity-and-authenticity.mdx
```diff
@@ -2,25 +2,628 @@
...
 
-### Recommendation
+On the other hand, query calls are fast since they are responded by a single replica but **there is no integrity guarantee, since the response can be manipulated if the responding replica or the boundary node is malicious.** For example, if the NNS dapp fetches proposal information from the governance canister via query calls and the responding node is malicious, it can mask an ill intentioned proposal that causes irrevocable damage as innocuous by modifying the proposal payload in the response and mislead voters into voting yes. Another consequence of query calls is that users can't rely on [canister_inspect_message](https://internetcomputer.org/docs/current/developer-docs/security/security-best-practices/iam#do-not-rely-on-ingress-message-inspection) as a guard. **This makes query calls, in it's raw form, unfit to serve data for security critical applications.**
```
roelstorms:
```suggestion
On the other hand, query calls are fast since they are responded by a single replica but **there is no integrity guarantee, since the response can be manipulated by a single replica or the boundary node.** For example, if the NNS dapp fetches proposal information from the governance canister via query calls and the responding node is malicious, it can mask an ill intentioned proposal that causes irrevocable damage as innocuous by modifying the proposal payload in the response and mislead voters into voting yes. Another consequence of query calls is that users can't rely on [canister_inspect_message](https://internetcomputer.org/docs/current/developer-docs/security/security-best-practices/iam#do-not-rely-on-ingress-message-inspection) as a guard. **This makes query calls, in it's raw form, unfit to serve data for security critical applications.**
```

docs/developer-docs/security/security-best-practices/data-integrity-and-authenticity.mdx
```diff
@@ -2,25 +2,628 @@
...
+
+:::tip
+ICP also provides replica signed queries, where query responses are signed by the answering replica node, however it doesn't have the same security guarantees an `update` call and only protects from malicious boundary nodes. Replica signed queries are enabled by default on both `agent-rs` and `agent-js`. [Read more about the feature](https://internetcomputer.org/blog/features/replica-signed-queries).
+:::
+
+### What is certified data?
+A canister can store a small amount of data (32 bytes) during an update call in the state tree under `certified_data` which will be certified by the subnet. During a query call, the canister can obtain a certificate about that data, which can be verified against the IC root public key. Combining these features, a canister should be able to certify a variable in a query response with this [design](https://medium.com/dfinity/how-internet-computer-responses-are-certified-as-authentic-2ff1bb1ea659). 
+
+On a high level, in your canister,
+1. Choose an [authenticated data strucutre](https://cs.brown.edu/research/pubs/pdfs/2003/Tamassia-2003-ADS.pdf) like Merkle trees for storing the value in canister memory.
```
roelstorms, line 40:
```suggestion
1. Choose an [authenticated data strucutre](https://cs.brown.edu/research/pubs/pdfs/2003/Tamassia-2003-ADS.pdf) like Merkle trees to store a value in canister memory.
```

docs/developer-docs/security/security-best-practices/data-integrity-and-authenticity.mdx
```diff
@@ -2,25 +2,628 @@
...
+2. In the **update** call,
+  - Perform the computation and store the result in the Merkle store.
```
roelstorms, line 42: Do you mean Merkle tree?

docs/developer-docs/security/security-best-practices/data-integrity-and-authenticity.mdx
```diff
@@ -2,25 +2,628 @@
...
+  - The lookup path for the result must act as it's `key`. Ideally this `key` should be the parameters provided by the caller in the query method.
+  - Recompute the Merkle proof (`root_hash`)
+  - Store the `root_hash` as the canister's certified data.
+  - Return the `key` as response.
```
roelstorms, line 46: Why do you need to return the key?

docs/developer-docs/security/security-best-practices/data-integrity-and-authenticity.mdx
```diff
@@ -2,25 +2,628 @@
...
+3. In the **query** call, 
+  - Fetch the result from the Merkle strucutre using the query parameters as the lookup path.
+  - Fetch the current `certified_data` for the canister.
+  - Compute the witness for the result using the same lookup path. Merkle witness provide proof of inclusion that the requested result exists in the Merkle store under the given path.
```
roelstorms, line 50:
```suggestion
  - Compute the witness for the result using the same lookup path. The Merkle witness provides proof of inclusion that the requested result exists in the Merkle tree under the given path.
```

docs/developer-docs/security/security-best-practices/data-integrity-and-authenticity.mdx
```diff
@@ -2,25 +2,628 @@
...
+};
+
+service : {
+    "set_user": (User) -> (nat64);
+    "get_user": (nat64) -> (CertifiedUser) query;
+}
+```
+
+The canister exposes the following service:
+* **set_user**: The caller provides a `User` object to the canister. The canister records it and serves a corresponding `index` for the entry as the response. Since `certified_data` can only store 32 bytes of data, it uses a specialized data structure from `ic_certified_map` to store the `User` data.
```
roelstorms, line 77: The ic_certified_map is doing much of the heavy lifting here. Should we educate users on how to create their own certified data structures?

docs/developer-docs/security/security-best-practices/data-integrity-and-authenticity.mdx
```diff
@@ -2,25 +2,628 @@
...
+
+### Verifying certified variables
+
+Once you have the response `CertifiedUser`, for the integrity guarantee, the front-end must verify the certification in the response. This is broken down into several steps implemented in the Rust and JavaScript example below.
+
+:::note
+The example has some extra steps to setup the canister with some `User` data before verification. You can ignore the section marked between `// ==== START of canister data setup` and `// ==== END of canister data setup`
+:::
+
+1. Recompute the `root_hash` of `certificate.tree` (pruned state tree with the canister's `certified_data`) and verify the `certificate.signature` with `root_hash` as the message,`certificate.delegation`, and the IC `root_key` as the public key. This confirms that the signature is valid for the current state tree.
```
roelstorms, line 332:
```suggestion
1. Verify the IC certificate: Recompute the `root_hash` of `certificate.tree` (pruned state tree with the canister's `certified_data`) and verify the `certificate.signature` with `root_hash` as the message,`certificate.delegation`, and the IC `root_key` as the public key. This confirms that the signature is valid for the current state tree. 
```

docs/developer-docs/security/security-best-practices/data-integrity-and-authenticity.mdx
venkkatesh-sekar: The `key` would be used as the query parameter to fetch the data. If it's transparent to the user, the canister need not respond anything. But if you use an internal counter, I suppose you would need to return it as the response.

docs/developer-docs/security/security-best-practices/data-integrity-and-authenticity.mdx
venkkatesh-sekar: > Should we educate users on how to create their own certified data structures?

That would be it's own page since there's too much stuff. I believe here people shouldn't roll their own crypto as there are things you can go wrong while build the hash tree. 

> Such as hash collisions.
> the `root_hash` cryptographically guarantees that only one tree can correspond to that hash.

We already mention this. I don't think we should discuss about hash collisions and this would just make it more confusing.

docs/developer-docs/security/security-best-practices/data-integrity-and-authenticity.mdx
venkkatesh-sekar: I will try to adapt this.

roelstorms (approved):
Aside from those last typos, it looks great to me.

PR #3220, opened 2024-07-16
docs(ckerc20): Describe upgrades for ledger suite orchestrator

gregorydemay:
Describe how to check canister upgrades proposals for the ledger suite orchestrator

### DIFF
--- a/docs/developer-docs/daos/nns/proposal-advanced.mdx
+++ b/docs/developer-docs/daos/nns/proposal-advanced.mdx
@@ -76,15 +76,21 @@ If you are looking to verify NNS core canister upgrades apart from regularly che
 
 
 #### How to verify ledger canister upgrades
-If you are looking to verify the ledger suite, which includes the ICP ledger and all ICRC ledgers (used for ckBTC, ckETC, ckERC20) apart from regularly checking open proposals, you might want to consider the following information for proposals submitted by the DFINITY foundation:
+If you are looking to verify the ledger suite, which includes the ICP ledger and all ICRC ledgers (used for ckBTC, ckETH, ckERC20) apart from regularly checking open proposals, you might want to consider the following information for proposals submitted by the DFINITY foundation:
 
-* For updates to the NNS canisters, you can follow [this repository](https://github.com/dfinity/ic/rosetta-api).
+* For updates to the NNS canisters, you can follow [this repository](https://github.com/dfinity/ic).
 
 More specifically, the following sub-directories are relevant:
 - `rs/rosetta-api/icp_ledger` is the  directory for the ICP ledger
 - `rs/rosetta-api/icrc1/ledger` for the ckBTC, ckETH, and ckERC20 ledgers, where `rs/rosetta-api/icrc1/ledger:ledger_canister` is the Wasm relevant for the ckBTC ledger and `rs/rosetta-api/icrc1/ledger:ledger_canister_u256.wasm` the Wasm for the ckETH and ckERC20 ledgers.
 
 * You can find the instructions on how to verify that the Wasm in the proposal was indeed build from the provided `git commit` in the proposal.
+
+#### How to verify ledger suite orchestrator canister upgrades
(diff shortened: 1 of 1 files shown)


docs/developer-docs/daos/nns/proposal-advanced.mdx
```diff
@@ -76,16 +76,22 @@ If you are looking to verify NNS core canister upgrades apart from regularly che
...
-* For updates to the NNS canisters, you can follow [this repository](https://github.com/dfinity/ic/rosetta-api).
+* For updates to the NNS canisters, you can follow [this repository](https://github.com/dfinity/ic).
 
 More specifically, the following sub-directories are relevant:
 - `rs/rosetta-api/icp_ledger` is the  directory for the ICP ledger
 - `rs/rosetta-api/icrc1/ledger` for the ckBTC, ckETH, and ckERC20 ledgers, where `rs/rosetta-api/icrc1/ledger:ledger_canister` is the Wasm relevant for the ckBTC ledger and `rs/rosetta-api/icrc1/ledger:ledger_canister_u256.wasm` the Wasm for the ckETH and ckERC20 ledgers.
 
 * You can find the instructions on how to verify that the Wasm in the proposal was indeed build from the provided `git commit` in the proposal.
 
+#### How to verify ledger suite orchestrator canister upgrades
```
gregorydemay, line 89: @LaraAS : Does-it correspond to what you had in mind?

jessiemongeon1 (approved):

docs/developer-docs/daos/nns/proposal-advanced.mdx
LaraAS: Yes, this goes into the right direction, thanks!

Regarding the organization, and also because ICP ledger will soon be in another NNS topic, would it make sense to have the following sections? 
#### How to verify ICP ledger canister upgrades
(mention just the repo for the ledger)

#### How to verify ledger upgrades for ckBTC, ckETH, and ckERC20
(mention the repo where this ledger is and the orchestrator as you did)
One could also maybe add 1 sentence about what other canisters are there (archive, index) and what is the job of the orchestrator (i.e., I think there is one orchestrator for all ledgers but one archive per ledger - is this correct?). I know this is probably written in a README or so, but I think it would be useful to explain the high-level idea in 1 sentence and then say "you can find more information about this in the README [link to readme]".

WDYT?

gregorydemay: Thanks Lara for the feedback! Some follow-up questions:

1. You mean splitting the _existing_ section on "How to verify ledger canister upgrades" into 2, between ICP ledger and the other ones? Note that the ledgers are owned by FI team, so if that's the way you want to go to, I could make a suggestion in this MR, but it might meaningful if someone from FI team also reviews it.
2. I have the impression that explaining the architecture of a ledger suite (purpose of index/archive canisters) goes beyond the scope of this document. If you need contribution on that front, I think you should ask the FI team.
3. Regarding the role of the orchestrator is this sentence "which manages all ledger suites of all ckERC20 tokens" not enough? Good point regarding the README, I added a link it.

LaraAS: 1. Yes this is what I mean. If you would like to do the split and just focus on the ck* section? Happy to then have someone from FI look at the other part (I think they already reviewed this once, but cannot hurt to ask again)
2. OK
3. Maybe we can just add a part which gives an intuition of what "manage" means? Maybe something like "which manages all ledger suites of all ckERC20 tokens, e.g, it is responsible for upgrading all canisters to the same version and for adding new tokens". The "adding new tokens" part I am actually not sure myself, so if this works differently than maybe this can be mentioned in 1 sentence?
Thanks for adding the link to the README!

gregorydemay: Thanks for Lara for the clarifications!

1. The section "How to verify ledger canister upgrades"  was split as you suggested (between ICP ledger and the other ones, including ckERC20 ledger suites, since they could also be subject to direct NNS proposal)
2. I added examples of what manages mean as you suggested, but I'm not trying to be exhaustive here (the linked doc should be the reference)
PTAL

LaraAS (approved):
Changes since last time LGTM. Thanks Gregory!

PR #3245, opened 2024-07-19
Dd 511 tokens ledger ledger development workflow

NikolasHai:

  side navigation bar.

### DIFF
--- a/docs/developer-docs/defi/tokens/ledger/overview.mdx
+++ b/docs/developer-docs/defi/tokens/ledger/overview.mdx
@@ -6,5 +6,22 @@ import { MarkdownChipRow } from "/src/components/Chip/MarkdownChipRow";
 # Ledgers
 
 ## Overview
+In this section you will learn about the different types of ledgers that exist on the Internet Computer. This guide will go into detail what the different ledgers are, what their purpose is and how they differ from ledgers you may be familiar with from other blockchains. 
+
+## ICP ledger
+The ICP ledger is a canister which runs on the subnet 0, or NNS subnet. It is responsible for keeping track of ICP balances, process new transactions and providing clients with data about the history of transactions. The data format that is used to store this historic data is a blockchain. For every transaction a single block is created which points to its parent block by storing the parent block's hash.
+The ICP ledger is using `AccountIdentifiers` as its internal account representation. It is compatible with the ICRC-1 standard, however due to the fact that it is using `AccountIdentifiers` instead of `Accounts` it stores a different block format than ICRC-1 ledgers. The ICP ledger is part of the NNS and its controller is the governance canister. The same is true for the minting account of the ICP ledger. 
+Upgrades to the ICP ledger have to be approved by the NNS and the decision making process on whether to execute and upgrade is thus decentralized. 
+
+## ICRC-1 ledger
+ICRC-1 ledgers are a standardized version of ledgers on the Internet Computer. While ICRC-1 is the standard on how ledgers on the Internet Computer should behave, an ICRC-1 ledger is the implementation of the ICRC-1 standard. It differs to the ICP ledger in so far as it is using `Accounts` as its internal representation. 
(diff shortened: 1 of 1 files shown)


docs/developer-docs/defi/tokens/ledger/overview.mdx
```diff
@@ -6,5 +6,22 @@ import { MarkdownChipRow } from "/src/components/Chip/MarkdownChipRow";
...
+The ICP ledger is one smart contract which runs on the NNS subnet, which means it sits on the application layer of the internet computer rather than the protocol level. Consensus on the Internet Computer is independent of the existence of the ICP ledger. Nevertheless, the ICP ledger is fundamentally important to the way the Internet Computer functions, especially due to its importance as a part of the NNS. 
+Transfers of the ICP token do not happen on the base protocol level but are only confined to the canister running the ICP ledger. All ICP token transactions go through this canister. Any information gathered about the blockchain that stores historic ICP token transactions originate from the ICP ledger canister. This includes the ICP index canister which fetches blocks periodically from the ICP ledger canister. 
+
+ICRC-1 ledgers are very similar in this regard, however they resemble tokens that can often be found on other blockchains more closely. They are also run by a canister that runs the ICRC-1 ledger code and all transactions of the token that is represented by an ICRC-1 ledger go through the ICRC-1 ledger canister.
```
jessiemongeon1, line 25:
```suggestion
In this section, you will learn about the different types of ledgers that exist on the Internet Computer. This guide will go into detail about what the different ledgers are, what their purposes are, and how they differ from ledgers you may be familiar with from other blockchains. 

## ICP ledger
The ICP ledger is a canister that is responsible for keeping track of ICP balances, processing new transactions, and providing clients with data about the history of transactions. It stores all data on-chain. For every transaction, a single block is created, which points to its parent block by storing the parent block's hash.

The ICP ledger uses `AccountIdentifiers` as its internal account representation. It is compatible with the ICRC-1 standard; however, due to the fact that it uses `AccountIdentifiers` instead of 'Accounts', it stores a different block format than ICRC-1 ledgers. The ICP ledger. The same is true for the minting account of the ICP ledger. 

It runs on the NNS subnet, as it is part of the NNS, and its controller is the governance canister. Upgrades to the ICP ledger have to be approved by the NNS, and the decision-making process on whether to execute and upgrade is thus decentralized. 

## ICRC-1 ledgers
ICRC-1 ledgers are a standardized version of ledgers on the Internet Computer. While [ICRC-1](/docs/current/developer-docs/defi/tokens/token-standards) is the network standard defining how ledger canisters should behave, an ICRC-1 ledger is a single implementation of the ICRC-1 standard. It differs from the ICP ledger as it uses `Accounts` as its internal representation. 

ICRC-1 ledgers are stand-alone implementations, but in many cases, they are part of a larger software structure, such as an SNS or chain-key token implementation. 

## Differences from other blockchains
One of the biggest differences between ledgers deployed on ICP and ledgers deployed on other blockchains is that most account-based blockchains have the balances of accounts baked into their base protocol (e.g., Ethereum). This is not the case with either ICP or ICRC-1 ledgers. 

The ICP ledger canister sits on the application layer of the network rather than the protocol level. Consensus on the Internet Computer is independent of the existence of the ICP ledger. Nevertheless, the ICP ledger is fundamentally important to the way the network functions, especially due to its importance as a part of the NNS. 

Since the ICP ledger runs on the application layer, transfers of the ICP token do not happen at the protocol level as they do on other chains. All ICP token transactions go through this canister. Any information gathered about historic ICP token transactions originates from the ICP ledger canister. This includes the ICP index canister, which fetches blocks periodically from the ICP ledger canister. 

ICRC-1 ledgers are very similar in this regard; however, they resemble tokens that can often be found on other blockchains more closely. They are also run by a canister that runs the ICRC-1 ledger code, and all transactions of theICRC-1 token go through an ICRC-1 ledger canister. 
```

jessiemongeon1 (approved):

PR #3296, opened 2024-07-31
Add cards for Chain Fusion to Docs homepage

jessiemongeon1:
Applying user feedback and adding Chain Fusion cards/links directly to Docs Homepage

Screenshot 2024-07-30 at 7 14 28 PM


dfx-json (approved):
Looks super!

PR #3304, opened 2024-08-01
Add Motoko `finally` feature to `inter-canister-calls.mdx`

ggreif:
Describe how the new `finally` feature can help with resource deallocation, and teach users about the residual hazards.

_NOTE_: To be merged when `dfx` with the feature is GA.

---------

  side navigation bar.


jessiemongeon1 (approved):

docs/developer-docs/security/security-best-practices/inter-canister-calls.mdx
```diff
@@ -269,11 +279,73 @@ mod test {
 }
 ```
 
+</TabItem>
+
+<TabItem value="motoko" label="Motoko">
+
+```motoko
+import Result "mo:base/Result";
+import Text "mo:base/Text";
```
ggreif, line 288: I eliminated this import (hashing `Principal`s directly),
see https://play.motoko.org/?tag=2353996190.

Contains other stylistic changes too, feel free to take or ignore those.

docs/developer-docs/security/security-best-practices/inter-canister-calls.mdx
```diff
@@ -269,11 +279,73 @@ mod test {
...
+
+</TabItem>
+</AdornedTabs>
+
 This pattern can be extended to work for the following use cases:
 
-- A global lock that does not only lock per caller. For this, set a boolean flag in the canister state instead of using a `BTreeSet<Principal>`.
-- A guard that makes sure that only a limited number of principals are allowed to execute a method at the same time. For this, one can return an error in `CallerGuard::new()` in case `pending_requests.len() >= MAX_NUM_CONCURRENT_REQUESTS`.
-- A guard that limits the number of times a method can be called in parallel. For this, use a counter in the canister state that is checked and increased in `CallerGuard::new()` and decreased in `Drop`.
+- A global lock that does not only lock per caller. For this, set a boolean flag in the canister state instead of using a `BTreeSet<Principal>` (rust) or `Map.HashMap<Text, Bool>` (Motoko).
```
ggreif, line 342: `Map.HashMap<Principal, Bool>` (Motoko).

docs/developer-docs/security/security-best-practices/inter-canister-calls.mdx
```diff
@@ -269,11 +279,71 @@ mod test {
...
+
+</TabItem>
+</AdornedTabs>
+
 This pattern can be extended to work for the following use cases:
 
-- A global lock that does not only lock per caller. For this, set a boolean flag in the canister state instead of using a `BTreeSet<Principal>`.
-- A guard that makes sure that only a limited number of principals are allowed to execute a method at the same time. For this, one can return an error in `CallerGuard::new()` in case `pending_requests.len() >= MAX_NUM_CONCURRENT_REQUESTS`.
-- A guard that limits the number of times a method can be called in parallel. For this, use a counter in the canister state that is checked and increased in `CallerGuard::new()` and decreased in `Drop`.
+- A global lock that does not only lock per caller. For this, set a boolean flag in the canister state instead of using a `BTreeSet<Principal>` (rust) or `Map.HashMap<Principal, Bool>` (Motoko).
```
robin-kunzler, line 340:
```suggestion
- A global lock that does not only lock per caller. For this, set a boolean flag in the canister state instead of using a `BTreeSet<Principal>` (Rust) or `Map.HashMap<Principal, Bool>` (Motoko).
```

docs/developer-docs/security/security-best-practices/inter-canister-calls.mdx
```diff
@@ -269,11 +279,71 @@ mod test {
...
+
+```motoko
+import Result "mo:base/Result";
+import Map "mo:base/HashMap";
+import Error "mo:base/Error";
+import Principal "mo:base/Principal";
+
+actor {
+
+  let pending_requests = Map.HashMap<Principal, Bool>(10, Principal.equal, Principal.hash);
```
robin-kunzler, line 294: nit: IIUC you use a HashMap because there is no set right? Would it make sense to explain this e.g. by adding this comment?
```
  // We use HashMap (and set a dummy booleans as values) even though a set would be sufficient. However, there is currently no Set implementation in Motoko.
```

docs/developer-docs/security/security-best-practices/inter-canister-calls.mdx
```diff
@@ -269,11 +279,71 @@ mod test {
...
+      #ok principal;
+    };
+  };
+
+  private func drop(principal : Principal) : () {
+    ignore pending_requests.remove(principal);
+  };
+
+  public shared ({caller}) func example_call_with_locking_per_caller() : async Result.Result<(), (Error.ErrorCode, Text)> {
+    var success = false;
```
robin-kunzler, line 310: nit: maybe call this `guard_acquired` rather than `success` to be a bit more explicit also if more code is added? WDYT?

docs/developer-docs/security/security-best-practices/inter-canister-calls.mdx
```diff
@@ -269,11 +279,71 @@ mod test {
...
+    try {
+      // Try to create a lock for `caller`, return an error immediately if there is already a call in progress for `caller`
+      switch (guard(caller)) {
+        case (#err e) {
+          return #err(Error.code(e), Error.message(e));
+        };
+        case (#ok _) {
+          success := true;
+        };
+      };
```
robin-kunzler, line 320: I was wondering if we could make this shorter. IIUC this could be reduced to: 
```
guard(caller);
success := true; 
```
if we changed `guard` to `throw` instead of returning an error. Then the catch would return the error. But honestly I don't like that because it makes the error handling less explicit and relies on `catch`, making the whole thing more error prone IMO. 

Since I don't really see another way of making it shorter, I suggest we leave it as is. Agreed? Do you maybe see a nice way to shorten this @ggreif ?

docs/developer-docs/security/security-best-practices/inter-canister-calls.mdx
```diff
@@ -269,11 +279,71 @@ mod test {
...
+      // do anything, call other canisters
+      #ok;
+    } catch e {
+      #err(Error.code(e), Error.message(e));
+    } finally {
+      // Release the guard (only request's callbacks that have created a block)
+      if success {
+        drop caller;
```
robin-kunzler, line 328: nit: WDYT about changing the name to `drop_guard` to make even clearer what happens here?

docs/developer-docs/security/security-best-practices/inter-canister-calls.mdx
```diff
@@ -269,11 +279,71 @@ mod test {
...
+import Result "mo:base/Result";
+import Map "mo:base/HashMap";
+import Error "mo:base/Error";
+import Principal "mo:base/Principal";
+
+actor {
+
+  let pending_requests = Map.HashMap<Principal, Bool>(10, Principal.equal, Principal.hash);
+
+  private func guard(principal : Principal) : Result.Result<Principal, Error> {
```
robin-kunzler, line 296: Do we need the `Principal` as result or should we switch to `()`? IIUC the return value is simply ignored further down.

robin-kunzler (approved):
Looks great, thanks @eduarddfinity and @ggreif ! Left a few comments but these are mostly minor. Approving!

docs/developer-docs/security/security-best-practices/inter-canister-calls.mdx
```diff
@@ -189,7 +193,13 @@ See also: "Inter-canister calls" section in [how to audit an ICP canister](https
...
+The locks would usually be released in the callback. That bears the risk that the lock may never be released in case the callback traps, as we discussed in [securely handle traps in callbacks](#securely-handle-traps-in-callbacks). The code examples below shows how one can securely implement a lock per caller.
+- In Rust, one can use the drop pattern where each caller lock (`CallerGuard` struct) implementats the `Drop` trait to release the lock. From Rust CDK version `0.5.1`, any local variables still go out of scope if the callback traps, so the lock on the caller is released even in that case. Technically, the CDK calls into the `ic0.call_on_cleanup` API to release these resources. Recall that `ic0.call_on_cleanup` is executed if the `reply` or the `reject` callback executed and trapped.
+- In Motoko, one can use `try`/`finally` control flow construct. This construct guarantees lock release execution in the `finally` block regardless of any exceptions in the `try` block.
```
robin-kunzler, line 198: Can we link to the Motoko documentation of `try`/`catch`/`finally` here?

docs/developer-docs/security/security-best-practices/inter-canister-calls.mdx
eduarddfinity: AFAIK the Motoko base lib does not have a set structure. But I could be wrong. 
@ggreif?

docs/developer-docs/security/security-best-practices/inter-canister-calls.mdx
eduarddfinity: I return the `Result<Principal, Error>` value in case people want to use this as a template and extend it. But if you think it's not necessary I can remove it.

robin-kunzler: I'd suggest to use Result<(), Error> because the result is currently unused and for me it was slightly confusing as it doesn't seem needed.

docs/developer-docs/security/security-best-practices/inter-canister-calls.mdx
ggreif: Well a set is basically a map `<KeyType, ()>`, i.e. the value type is unit. So It should be easy to typedef it that way.

But I also think that a grant is being prepared to come up with modern implementations of associative containers. So we might have an efficient `Set` someday soon.

docs/developer-docs/security/security-best-practices/inter-canister-calls.mdx
ggreif: I think it is shorter if written as
``` Motoko
      switch (guard caller) {
        case (#ok _) success := true;
        case e e;
        };
      };
```

docs/developer-docs/security/security-best-practices/inter-canister-calls.mdx
```diff
@@ -293,43 +293,41 @@ actor {
...
+      errResult e;
     } finally {
       // Release the guard (only request's callbacks that have created a block)
       if guard_acquired {
         drop_guard caller;
       };
     };
   };
-};
+};
```
ggreif, line 330: maybe drop final space, sorry it was me who probably added it...

ggreif (commented):
`dfx` 0.24 beta is out, maybe we can merge this now...

PR #3406, opened 2024-08-29
docs: update AI roadmap

ielashi:


roadmap/roadmap.json
```diff
@@ -1176,6 +1150,32 @@
             "docs": "",
             "is_community": false,
             "in_beta": false
+          },
+          {
+            "title": "Public specification for GPU-enabled nodes",
+            "overview": "The ICP community needs to agree and adopt the hardware specification for replica nodes with GPUs.",
+            "description": "The ICP community needs to agree and adopt the hardware specification for replica nodes with GPUs.",
+            "status": "",
+            "milestone_id": "Gyrotron",
```
ulan, line 1159: Would it make sense to push this and the next feature to another milestone (that comes after Gyrotron)?

ulan (approved):

roadmap/roadmap.json
ielashi: Yes, this is the list of future features. Here's how it looks like:

image

Does this map to what you're thinking as well, or did I misunderstand?

ulan: Looks good!

jessiemongeon1 (approved):

PR #3462, opened 2024-09-12
First attempt at a document on idempotency

oggy-dfin:

  side navigation bar.


docs/developer-docs/smart-contracts/best-practices/idempotency.mdx
```diff
@@ -0,0 +1,90 @@
...
+request has been processed. For example, this can happen if the client loses its
+connection until the request status has been removed from the state tree (recall
+that ICP will remove the request from the system state tree some time after
+ingress expiry).
+
+Thus, it is important to design and/or use canister APIs such that it is possible to retry requests safely, even when the ICP doesn't provide any information as to previous request attempts. This page describes general approaches that both the canister authors and the clients can adopt to enable safe retries.
+
+## Idempotent Canister APIs
+
+We say that a canister endpoint is idempotent if it can be called multiple times without changing the result or canister state beyond the first call. Whenever an endpoint is idempotent or can be made idempotent by the developer, this provides an easy way to implement safe retries.
```
roelstorms, line 20:
```suggestion
An endpoint of a canister is considered idempotent if calling it multiple times does not alter the result or the state of the canister after the initial call. When an endpoint is idempotent or can be made idempotent by the developer, it offers a straightforward approach to implementing safe retries.
```

docs/developer-docs/smart-contracts/best-practices/idempotency.mdx
```diff
@@ -0,0 +1,90 @@
+---
+keywords: [beginner, concept, best practices, developer]
+---
+
+import { MarkdownChipRow } from "/src/components/Chip/MarkdownChipRow";
+
+# Retries and Idempotency
+
+In the case of network or other issues, ICP clients (such as agents) that issue
+ingress update calls may become unable to determine whether their ingress
```
roelstorms, line 10:
```suggestion
ingress update calls may be unable to determine whether their ingress
```

docs/developer-docs/smart-contracts/best-practices/idempotency.mdx
```diff
@@ -0,0 +1,90 @@
...
+request has been processed. For example, this can happen if the client loses its
+connection until the request status has been removed from the state tree (recall
+that ICP will remove the request from the system state tree some time after
+ingress expiry).
+
+Thus, it is important to design and/or use canister APIs such that it is possible to retry requests safely, even when the ICP doesn't provide any information as to previous request attempts. This page describes general approaches that both the canister authors and the clients can adopt to enable safe retries.
```
roelstorms, line 16:
```suggestion
Thus, it is important to design and/or use canister APIs such that it is possible to retry requests safely, even when the ICP provides no information about previous request attempts. This page describes general approaches that both the canister authors and the clients can adopt to enable safe retries.
```

docs/developer-docs/smart-contracts/best-practices/idempotency.mdx
```diff
@@ -0,0 +1,90 @@
...
+
+## Idempotent Canister APIs
+
+We say that a canister endpoint is idempotent if it can be called multiple times without changing the result or canister state beyond the first call. Whenever an endpoint is idempotent or can be made idempotent by the developer, this provides an easy way to implement safe retries.
+
+Given an idempotent endpoint, you can implement retries by simply retrying the call until you observe a certified response (either a replied or rejected status), see the illustration below. If such a response is ever observed, one can be sure that the transaction has been executed exactly once thanks to idempotency. However, the application may not be willing to wait for a response indefinitely and a timeout could be implemented. Upon timeout, an error should be displayed to the user instructing them to wait until the latest message that has been sent has expired (ingress_expiry) and then manually check the status of the transaction. Ideally, timeouts should be rare and not occur during normal operation.
```
roelstorms, line 22:
```suggestion
Given an idempotent endpoint, you can implement retries by retrying the call until you observe a certified response (either a replied or rejected status), see the illustration below. If such a response is ever observed, one can be sure that the transaction has been executed exactly once thanks to idempotency. However, the application may not be willing to wait for a response indefinitely and a timeout could be implemented. Upon timeout, an error should be displayed to the user instructing them to wait until the latest message that has been sent has expired (ingress_expiry) and then manually check the status of the transaction. Ideally, timeouts should be rare and not occur during normal operation.
```

docs/developer-docs/smart-contracts/best-practices/idempotency.mdx
```diff
@@ -0,0 +1,90 @@
...
+
+But even with this improvement used in the ledgers, the time window approach implicitly assumes that the client will be able to get a definite answer to their call within the time window. For example, after the 24 hours expire, the user cannot easily tell if their ledger transfer happened; their only option is to analyze the ledger blocks, which is somewhat tedious, and has to be done carefully to avoid asynchrony issues; see the section on [queryable call results](#queryable-call-results).
+
+An alternative is to do away with the time window, and store the deduplication data forever. This requires multiple canisters to prevent exhausting the canister memory, similar to how the ICP/ICRC ledgers store the transaction data in the archive canister. This effectively moves the tedious part of querying the deduplication data (e.g., ledger blocks) from the user to the canister.
```
roelstorms, line 52:
```suggestion
An alternative is to do away with the time window, and store the deduplication data forever. This requires multiple canisters to prevent exhausting the canister memory, similar to how the ICP/ICRC ledgers store the transaction data in the archive canister. This shifts the tedious part of querying the deduplication data (e.g., ledger blocks) from the user to the canister.
```

docs/developer-docs/smart-contracts/best-practices/idempotency.mdx
```diff
@@ -0,0 +1,90 @@
...
+
+## Other approaches to safe retries
+
+In absence of idempotent endpoints, or even as an addition to them, clients may be able to use other endpoints to make their retries safe.
+
+### Queryable call results
+
+If the canister, in addition to the update endpoint, also exposes a query that can inform the user of the result of the update, the client can also use this for safe retries as follows:
+
+1. Attempt to perform the update
```
jessiemongeon1, line 71:
```suggestion
1. Attempt to perform the update.
```

docs/developer-docs/smart-contracts/best-practices/idempotency.mdx
```diff
@@ -0,0 +1,90 @@
...
+1. If the result of the update is unknown (e.g., not present in the ingress history any more), query the call result endpoint to determine whether the call was applied or not. Moreover, one needs to ensure that the previously sent call call cannot be applied in the future. If both of these are true, the call might be retried, or safely reported as failed.
```
jessiemongeon1, line 72:
```suggestion
1. If the result of the update is unknown (e.g., not present in the ingress history any more), query the call result endpoint to determine whether the call was applied or not. Moreover, one needs to ensure that the previously sent call cannot be applied in the future. If both of these are true, the call might be retried or safely reported as failed.
```

docs/developer-docs/smart-contracts/best-practices/idempotency.mdx
```diff
@@ -0,0 +1,90 @@
...
+
+In practice, this pattern may be more complicated. For example, the ICP ledger exposes a `query_blocks` method that can be used to implement the above pattern for transfers initiated as ingress messages:
+
+1. Call the `query_blocks` method on the ledger to determine what the last block (as specified in the `chain_length` field of the response) currently is. Let's call this `last_block`.
+1. Attempt to perform a transfer. This ingress message includes an `ingress_expiry` field.
+1. If the result of the transfer is unknown, call the `read_state` endpoint on the ledger canister, to obtain the `/time` branch of the system state tree. Repeat this until the reported time exceeds the `ingress_expiry` time. This ensures that the transfer will not be applied at a later point.
```
jessiemongeon1, line 78:
```suggestion
1. If the result of the transfer is unknown, call the `read_state` endpoint on the ledger canister to obtain the `/time` branch of the system state tree. Repeat this until the reported time exceeds the `ingress_expiry` time. This ensures that the transfer will not be applied at a later point.
```

docs/developer-docs/smart-contracts/best-practices/idempotency.mdx
```diff
@@ -0,0 +1,90 @@
...
+1. Call the `query_blocks` method on the ledger again to retrieve all ledger blocks since `last_block`, and check that the `timestamp` also exceeds the `ingress_expiry` time. Then, scan through the returned blocks to determine whether the transaction has been included or not.
+
+### 2-step transfers
+
+Another approach applicable to ledgers (such as ICRC-1 or ICP) is to perform transfers in two steps:
+
+1. First, transfer the money to an intermediate subaccount of the sender that's specific to this transaction. For example, if the transaction has a unique ID, the client can hash the ID to obtain a subaccount. The transferred amount should be the desired amount increased by the ledger transaction fee.
```
jessiemongeon1, line 85:
```suggestion
1. First, transfer the tokens to an intermediate subaccount of the sender that's specific to this transaction. For example, if the transaction has a unique ID, the client can hash the ID to obtain a subaccount. The transferred amount should be the desired amount plus the ledger transaction fee.
```

docs/developer-docs/smart-contracts/best-practices/idempotency.mdx
```diff
@@ -0,0 +1,90 @@
...
+1. If the result of the above transfer is unknown, query the balance of the transaction-specific subaccount. Like in the ["queryable call result"](#queryable-call-results) approach, this should be repeated until the `timestamp` accompanying the response exceeds the `ingress_expiry`. If the balance is 0, the transaction can safely be reported as failed, or retried (starting from step 1). If the balance is at least the expected balance, one can proceed.
```
jessiemongeon1, line 86:
```suggestion
1. If the result of the above transfer is unknown, query the balance of the transaction-specific subaccount. Like in the ["queryable call result"](#queryable-call-results) approach, this should be repeated until the `timestamp` accompanying the response exceeds the `ingress_expiry`. If the balance is 0, the transaction can safely be reported as failed, or it can be retried (starting from step 1). If the balance is at least the expected balance, one can proceed.
```

docs/developer-docs/smart-contracts/best-practices/idempotency.mdx
```diff
@@ -0,0 +1,90 @@
...
+1. If the transfer to the transaction-specific subaccount succeeded (as determined either by the transfer result, or by the balance query above), the client sends another transfer from the transaction-specific subaccount to the desired target account. This can be repeated as many times as necessary until a result of the call is known. Once a result is known, the overall transfer can be declared as succeeded, even if this step fails with an error, as this signifies that some previous attempt to transfer the money to the target succeeded.
```
jessiemongeon1, line 87:
```suggestion
1. If the transfer to the transaction-specific subaccount succeeded (as determined either by the transfer result or by the balance query above), the client sends another transfer from the transaction-specific subaccount to the desired target account. This can be repeated as many times as necessary until a result of the call is known. Once a result is known, the overall transfer can be declared as succeeded, even if this step fails with an error, as this signifies that some previous attempt to transfer the money to the target succeeded.
```

jessiemongeon1 (approved):

docs/developer-docs/smart-contracts/best-practices/idempotency.mdx
alin-at-dfinity, line 72:
```suggestion
1. If the result of the update is unknown (e.g., not present in the ingress history anymore), query the call result endpoint to determine whether the update was applied or not. Moreover, one needs to ensure that the previously sent call cannot be applied in the future. If both of these are true, the call might be retried, or safely reported as failed.
```

That being said, the paragraph is not particularly clear. What exactly does it mean for a call to not be applied in the future? If both of what are true?

docs/developer-docs/smart-contracts/best-practices/idempotency.mdx
roelstorms: This advice for when a timeout happens is probably the best we can do but it's also not great. The user submitted a transactions, like a token transfer. The front-end now needs to keep that in the UI as a potentially failed transfer. The user can't make another transfer since that might be a double transfer.

I'm wondering if the front end should ever timeout and stop trying to reconcile the state.

If the user goes to another device, they might not even see the attempted transaction which is in an unknown state.

docs/developer-docs/smart-contracts/best-practices/idempotency.mdx
derlerd-dfinity: Is "without changing the result or canister state beyond the first call" accurate? It seems that it (1) assumes that the first call actually reaches the canister (otherwise a retry after an initial call that failed because it didn't even reach the canister would be required to fail as well, which is not the intention of idempotency, right?), and (2) ignores canister state that will change on every call like cycles balance.

Maybe this is fine for the introduction, but wanted to raise it nevertheless to see what you think.

derlerd-dfinity (commented):
Thank you. LGTM.

docs/developer-docs/smart-contracts/best-practices/idempotency.mdx
alin-at-dfinity: It could say "beyond the first executed call".

And leaving out "changing the state" and focusing only on the result. Although I guess that simply producing the same result does not imply idempotency; so maybe "without meaningfully altering the state".

roelstorms: Is it incorrect if we leave out "meaningfully"? Isn't any state change violating idempotency?

alin-at-dfinity: I guess David's point is that the canister state will change on any call, whether it's the cycle balance, incidentally changing the size of the heap or bumping a counter of "total requests handled".

Idempotency in this context is more about only applying the change _relevant to the specific call_ exactly once. So e.g. I would consider an account transfer idempotent as long as the transfer happened exactly once, even if some small fee was deducted from my account every time I retried said call. In that context I care primarily about making a payment first and foremost and not so much about how much I'm charged for it and when (assuming it's a reasonably small fee).

robin-kunzler (approved):
Looks great @oggy-dfin , thanks a lot for putting in the work!

Really like the explicit listing of advantages and disadvantages for the sequence numbers and deduplication approaches. This makes it much clearer. 

Approving, left a few mostly minor comments / questions below.

docs/developer-docs/smart-contracts/best-practices/idempotency.mdx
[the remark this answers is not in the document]
oggy-dfin: And that would be because I forgot to push some commits 🙈

docs/developer-docs/smart-contracts/best-practices/idempotency.mdx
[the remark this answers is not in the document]
oggy-dfin: "make it" wouldn't work with using the APIs in a particular way

docs/developer-docs/smart-contracts/best-practices/idempotency.mdx
oggy-dfin: Covered by other suggestions/changes

docs/developer-docs/smart-contracts/best-practices/idempotency.mdx
```diff
@@ -0,0 +1,90 @@
...
+
+We describe two approaches to making endpoints idempotent: sequence numbers and (time window) ID deduplication.
+
+### Update sequence numbers
+
+An endpoint can make use of sequence numbers to provide idempotency. The endpoint can take a sequence number parameter in the addition to other parameters. In the extreme case, the canister could keep a single sequence counter for every endpoint, and a call could only be accepted if it contains the next expected sequence number, causing the expected sequence number to be incremented upon call execution. This trivially implies that any call can only be executed once. More practically, an expected sequence number is kept for each caller principal, or, in case of ledger-like canisters, each ledger account. Note that Ethereum implements this mechanism.
```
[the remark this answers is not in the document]
oggy-dfin: Already applied the suggestion by someone else

docs/developer-docs/smart-contracts/best-practices/idempotency.mdx
oggy-dfin: I shortened this to "executing the endpoint multiple times is equivalent to executing it once" and added a footnote to comment on the fact that we take the "user intent" into the account when we say "equivalent".

docs/developer-docs/smart-contracts/best-practices/idempotency.mdx
```diff
@@ -0,0 +1,90 @@
...
+
+### ID deduplication
+
+Another approach to idempotency is to make the calls uniquely identifiable on the canister side (e.g. by using user-chosen IDs, sequence numbers, or a combination of several argument fields) to make sure a given call is executed at most once. The canister then deduplicates calls before executing them; if a call with the same ID has been executed previously, the new call is simply ignored (potentially returning the result of the previous call). Thus, the user can safely keep retrying the call, until they get a response.
+
+For example, the ICRC ledger standard provides deduplication in this way. Using identical values for all call parameters, including the `created_at_time` and `memo` parameters when issuing a transaction makes the transaction call idempotent, by deduplicating calls with the same parameters.
+
+However, a direct implementation of this approach can exhaust the canister memory, as all successfully executed IDs need to be kept around forever.
```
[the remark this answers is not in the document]
oggy-dfin: Changed

docs/developer-docs/smart-contracts/best-practices/idempotency.mdx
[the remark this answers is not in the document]
oggy-dfin: Addressed in other comments

docs/developer-docs/smart-contracts/best-practices/idempotency.mdx
```diff
@@ -0,0 +1,90 @@
...
+
+## Idempotent Canister APIs
+
+We say that a canister endpoint is idempotent if it can be called multiple times without changing the result or canister state beyond the first call. Whenever an endpoint is idempotent or can be made idempotent by the developer, this provides an easy way to implement safe retries.
+
+Given an idempotent endpoint, you can implement retries by simply retrying the call until you observe a certified response (either a replied or rejected status), see the illustration below. If such a response is ever observed, one can be sure that the transaction has been executed exactly once thanks to idempotency. However, the application may not be willing to wait for a response indefinitely and a timeout could be implemented. Upon timeout, an error should be displayed to the user instructing them to wait until the latest message that has been sent has expired (ingress_expiry) and then manually check the status of the transaction. Ideally, timeouts should be rare and not occur during normal operation.
+
+We describe two approaches to making endpoints idempotent: sequence numbers and (time window) ID deduplication.
+
+### Update sequence numbers
```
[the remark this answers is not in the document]
oggy-dfin: Discussed with Robin briefly, he's OK with this. I'll resolve to be able to merge, we can revisit it later if necessary.

docs/developer-docs/smart-contracts/best-practices/idempotency.mdx
oggy-dfin: I'll resolve this to be able to merge, hopefully everyone is happy with the current wording - shout if not.

PR #3526, opened 2024-09-24
A note and another strategy for ID-based deduplication

oggy-dfin:
Bogdan had a neat idea of how to extend the window-based ID deduplication approach. In the process of writing it up, I noticed that i didn't write anything about the backpressure when the capacity within the window is reached, so I added that.

- [X] Follows the [developer docs style guide](style-guide.md).
- [X] Follows the [best practices and guidelines](README.md#best-practices).
- [X] New documentation pages include [document tags](README.md#document-tags).
- [X] New documentation pages include [SEO keywords](README#seo-keywords).
- [X] New documents are in the `.mdx` file format to support the previous two components. 
- [X] New documents are registered in [`/sidebars.js`](https://github.com/dfinity/portal/blob/master/sidebars.js), otherwise, it will not appear in the
  side navigation bar.
- [X] Make sure that the [`.github/CODEOWNERS`](https://github.com/dfinity/portal/blob/master/.github/CODEOWNERS) file is


docs/developer-docs/smart-contracts/best-practices/idempotency.mdx
```diff
@@ -56,6 +56,8 @@ Moreover, the ICP/ICRC ledgers use the `created_at_time` parameter to limit the
...
 
+Furthermore, just using a time window for deduplication does not guarantee that the memory will stay bounded. In principle, arbitrarily many updates could happen within the time window, although this is bounded in practice by the scaling limits of the ICP. The ICP/ICRC ledgers thus also define a maximum capacity: the number of deduplicated transactions (i.e., deduplication IDs) that they will store in their deduplication store. Further transactions are simply rejected after the capacity is reached, until the window passes and old transactions can be removed from the deduplication store. Yet another extension of the approach is to make the time window simply the guaranteed lower bound, and just store deduplication IDs until the capacity is reached. This way, the clients obtain a hard deduplication guarantee for the time window, and a best-effort attempt to deduplicate transactions even past the window.
```
roelstorms, line 59:
```suggestion
Relying solely on a time window for deduplication does not guarantee bounded memory usage. In theory, an unlimited number of updates could occur within the time window, though in practice, this is constrained by the scaling limits of the ICP.  The ICP/ICRC ledgers thus also define a maximum capacity: a limit on the number of deduplicated transactions (i.e., deduplication IDs) that can be stored in their deduplication store. Once this capacity is reached, further transactions are rejected until older transactions expire from the deduplication store at the end of the time window. Yet another extension of the approach is to make the time window simply the guaranteed lower bound, and just store deduplication IDs until the capacity is reached. This way, the clients obtain a hard deduplication guarantee for the time window, and a best-effort attempt to deduplicate transactions even past the window.
```

robin-kunzler, line 59: @oggy-dfin Might be missing something but I don't fully understand this, can you explain?
> Yet another extension of the approach is to make the time window simply the guaranteed lower bound, and just store deduplication IDs until the capacity is reached.

What is the "guaranteed lower bound"?

roelstorms: You guarantee deduplication up to 24h. Afterwards it might still deduplicate but the transaction might have been evicted from storage if it exceeded the storage capacity. So long as the ledger isn't used much, the deduplication store will not be full and you might end up having deduplication for your transaction for days or weeks.

robin-kunzler: Ah now I get it, thanks!

Would the following wording make this clearer? 
> Yet another extension of the approach is to guarantee deduplication for a given time window, but keep storing deduplication IDs even beyond that window, until the storage capacity is reached.

robin-kunzler (approved):
Thanks a lot for improving this further @oggy-dfin and @bogwar for the nice idea! Approving - only had a minor suggestion, see above.

docs/developer-docs/smart-contracts/best-practices/idempotency.mdx
oggy-dfin: Agree that it's clearer that way, thanks! Adopted with a slight tweak.

PR #3529, opened 2024-09-24
Unify message execution properties

robin-kunzler:
We had duplicate information on message execution basics. This PR unifies the two places: 
* Move all the information to references "Properties of message executions on ICP". I think the original page here was originally duplicated from the security best practices on inter-canister calls. Since the latter evolved further in the meantime, I'm moving the content from security best practices to here, replacing the current content. 
* The security best practices on inter-canister now refer to the reference rather than providing the same content.
* Added a more detailed introduction to the "Properties of message executions on ICP" and improved a few details

  side navigation bar.


jessiemongeon1 (approved):

docs/references/message-execution-properties.mdx
```diff
@@ -2,67 +2,55 @@ import { MarkdownChipRow } from "/src/components/Chip/MarkdownChipRow";
...
 
-Message execution is sequential, i.e. only a single message per canister is processed at a time.
+The [community conversation on security best practices](https://www.youtube.com/watch?v=PneRzDmf_Xw&list=PLuhDt1vhGcrez-f3I0_hvbwGZHZzkZ7Ng&index=2&t=4s) also discusses the messaging properties.
 
+## Message execution properties
 
-### Property 2
+A **call** is a canister's implementation of either an [update](/references/ic-interface-spec.md#http-call) or [query call](/docs/current/references/ic-interface-spec#http-query) that it exposes. For example, if the Rust CDK is used, these are usually annotated with `#[query]` or `#[update]`, respectively. ICP also supports additional [entry points](/references/ic-interface-spec.md#entry-points) such as heartbeats, timers, initialization or upgrade hooks. A **message** is a set of consecutive instructions that a subnet executes for a canister. The code execution for such an entry point can be split into several messages if inter-canister calls are made. The following properties are essential:
```
roelstorms, line 15:
```suggestion
To interact with a canister's methods, there are two primary types of **calls** that can be used: [update](/references/ic-interface-spec.md#http-call) calls and [query](/docs/current/references/ic-interface-spec#http-query) calls.
If the Rust CDK is used, these are usually annotated with `#[update]` or `#[query]`, respectively. ICP also supports additional [entry points](/references/ic-interface-spec.md#entry-points) such as heartbeats, timers, initialization or upgrade hooks. 

A **message** is a set of consecutive instructions that a subnet executes for a canister. The code execution for such any entry point can be split into several messages if inter-canister calls are made. The following properties are essential:
```

This is the same explanation as https://internetcomputer.org/docs/current/developer-docs/smart-contracts/call/overview which I think is more accurate to describe what a call is.

roelstorms: Are we missing the property that message execution is atomic?

docs/references/message-execution-properties.mdx
```diff
@@ -2,67 +2,55 @@ import { MarkdownChipRow } from "/src/components/Chip/MarkdownChipRow";
...
+![example_orderings](./_attachments/example_orderings.png)
 
+- **Property 5**: On a trap or panic, modifications to the canister state for the current message are not applied.
 
-### Property 6
+For example, if a trap in the second message (blue box) of the above example occurs, canister state changes resulting from that message, even earlier in the blue box, are discarded. However, note that any state changes from earlier messages and in particular the first message (orange box) have been applied, as that message executed successfully.
 
-Inter-canister calls are not guaranteed to make it to the destination canister. If the call does reach the destination canister, the destination canister can trap or return a **reject** response while processing the call.
+- **Property 6**: Inter-canister calls are not guaranteed to make it to the destination canister, and if a call does reach the destination canister, the destination canister can trap or return a reject response while processing the call.
```
roelstorms, line 52: Should we add that the destination canister can simply NOT reply?

```suggestion
- **Property 6**: Inter-canister calls are not guaranteed to make it to the destination canister, and if a call does reach the destination canister, the destination canister can trap or return a reject response while processing the call. It could also choose to never produce a response.
```

docs/references/message-execution-properties.mdx
```diff
@@ -2,67 +2,55 @@ import { MarkdownChipRow } from "/src/components/Chip/MarkdownChipRow";
...
+Every inter-canister call is guaranteed to receive a response, either from the canister, or synthetically produced by the protocol. However, the response does not have to be successful, but can also be a reject response. The reject may come from the called canister, but it may also be generated by ICP. Such protocol-generated rejects can occur at any time before the call reaches the callee-canister, as well as once the call does reach the callee-canister if the callee-canister traps while processing the call. If the call reaches the callee-canister, the callee-canister can produce a reply or reject response and the protocol guarantees that the callee-canister's generated reply or reject response gets back to the caller-canister. Thus, it's important that the calling canister handles reject responses as well. A reject response means that the message hasn't been successfully processed by the receiver but doesn't guarantee that the receivers state wasn't changed.
```
roelstorms, line 54: > Every inter-canister call is guaranteed to receive a response

Is this true? 

I believe we contradict this in the security best pracices:

> When another canister is called with a callback being registered, and the receiver stalls the response indefinitely by not responding, the result would be a DoS.

roelstorms (approved):
Some minor comments but otherwise good to go. Thanks for cleaning this up!

docs/developer-docs/security/security-best-practices/inter-canister-calls.mdx
```diff
@@ -12,51 +12,9 @@ import { MarkdownChipRow } from "/src/components/Chip/MarkdownChipRow";
 
 <MarkdownChipRow labels={["Intermediate", "Security", "Concept"]} />
 
-## Message execution basics
+To understand the issues around async inter-canister calls, one needs to understand the [properties of message executions on ICP](/docs/current/references/message-execution-properties). Understanding these properties is a prerequisite for the security issues discussed below.
```
oggy-dfin, line 15:
```suggestion
To understand the issues around async inter-canister calls, one needs to understand the [properties of message execution on ICP](/docs/current/references/message-execution-properties). Understanding these properties is a prerequisite for understanding the security issues discussed below.
```

docs/developer-docs/security/security-best-practices/inter-canister-calls.mdx
```diff
@@ -177,7 +135,7 @@ GoldDAO's GLDT-swap has an implementation of journaling. In their case, the jour
...
+As described in the [properties of message executions on ICP](/docs/current/references/message-execution-properties), messages (but not entire calls) are processed atomically. In particular, as described in Property 4 above, messages from interleaving calls do not have a reliable execution ordering. Thus, the state of the canister (and other canisters) may change between the time an inter-canister call is started and the time when it returns, which may lead to issues if not handled correctly. These issues are generally called 'Reentrancy bugs' (see the [Ethereum best practices on reentrancy](https://consensys.github.io/smart-contract-best-practices/attacks/reentrancy/)). Note however that the messaging guarantees, and thus the bugs, on ICP are different from Ethereum.
```
oggy-dfin, line 138:
```suggestion
As described in the [properties of message executions on ICP](/docs/current/references/message-execution-properties), messages (but not entire calls) are processed atomically. In particular, as described in Property 4 in that document, messages from interleaving calls do not have a reliable execution ordering. Thus, the state of the canister (and other canisters) may change between the time an inter-canister call is started and the time when it returns, which may lead to issues if not handled correctly. These issues are generally called 'Reentrancy bugs' (see the [Ethereum best practices on reentrancy](https://consensys.github.io/smart-contract-best-practices/attacks/reentrancy/)). Note however that the messaging guarantees, and thus the bugs, on ICP are different from Ethereum.
```

docs/references/message-execution-properties.mdx
```diff
@@ -2,67 +2,55 @@ import { MarkdownChipRow } from "/src/components/Chip/MarkdownChipRow";
...
 # Properties of message executions on ICP
 
-<MarkdownChipRow labels={["Reference"]} />
+<MarkdownChipRow labels={["Intermediate", "Reference", "Security"]} />
 
-There are a few key properties that one needs to understand about the execution model on ICP. Let's walk through them.
+## Asynchronous messaging model
 
-### Property 1
+ICP relies on an asynchronous messaging model. Compared to synchronous messaging like on Ethereum, this provides performance advantages because individual messages can be flexibly scheduled. However, asynchrounous message execution can also lead to sometimes unexpected or unintuitive behavior. Therefore, it is important to understand the properties of message execution. Potential security issues that arise in this model, such as reentrancy bugs, are discussed in the [security best practices on inter-canister calls](/docs/current/developer-docs/security/security-best-practices/inter-canister-calls).
```
oggy-dfin, line 9:
```suggestion
ICP relies on an asynchronous messaging model. Compared to synchronous messaging like on Ethereum, this provides performance advantages because individual messages can be executed in parallel. However, asynchrounous message execution can also lead to sometimes unexpected or unintuitive behavior. Therefore, it is important to understand the properties of message execution. Potential security issues that arise in this model, such as reentrancy bugs, are discussed in the [security best practices on inter-canister calls](/docs/current/developer-docs/security/security-best-practices/inter-canister-calls).  
```

docs/references/message-execution-properties.mdx
oggy-dfin: Not to conflict with Roel's suggestion:
- "code execution for such any entry point" -> "any such"
- say how updates are represented in Motoko, since we're using a Motoko example below

docs/references/message-execution-properties.mdx
```diff
@@ -2,67 +2,55 @@ import { MarkdownChipRow } from "/src/components/Chip/MarkdownChipRow";
...
 
- Each call (update or query) triggers a message. If an inter-canister call is involved, the code after the call (i.e. the callback code) is executed as a separate message.
+- **Property 1**: Only a single message is processed at a time per canister. Message execution is sequential, and never parallel.
 
-Let's look at an example to further clarify this point:
+- **Property 2**: Each call, query or update, triggers a message. When an inter-canister call is made using `await`, the code after the call (the callback, highlighted in blue) is executed as a separate message.
```
oggy-dfin, line 19:
```suggestion
- **Property 2**: Each call, query or update, triggers a message. When using `await` on the response from an inter-canister call, the code after the `await` (the callback, highlighted in blue) is executed as a separate message.
```

docs/references/message-execution-properties.mdx
```diff
@@ -2,67 +2,55 @@ import { MarkdownChipRow } from "/src/components/Chip/MarkdownChipRow";
...
-```
-func example(): async Result {
-  // block 1
-  await some_inter_canister_call();
-  // block 2
-  ...
-}
-```
+:::info
+Note that if the code does not `await` the response, the code after the callback is executed in the same message, until the next inter-canister call is triggered using `await`.
```
oggy-dfin, line 22:
```suggestion
Note that if the code does not `await` the response, the code after the callback is executed in the same message, until the next inter-canister call is triggered using `await`.
Also, multiple outgoing calls can be triggered in parallel from the same message execution; see the [Rust](https://internetcomputer.org/docs/current/references/samples/rust/parallel_calls/) and [Motoko](https://internetcomputer.org/docs/current/references/samples/motoko/parallel_calls/) examples.
```

oggy-dfin: Actually I'd probably move this info box to the end of the discussion of this property, as this is "advanced" stuff.

docs/references/message-execution-properties.mdx
```diff
@@ -2,67 +2,55 @@ import { MarkdownChipRow } from "/src/components/Chip/MarkdownChipRow";
...
+:::
 
-Successfully delivered requests are received in the order in which they were sent. In particular, if canister A sends m1 and m2 in that order to canister B, then, if both are accepted, m1 will be executed before m2.
+- **Property 3**: Successfully delivered requests are received in the order in which they were sent. In particular, if a canister A sends `m1` and `m2` to canister B in that order, then, if both are accepted, `m1` is executed before `m2`.
 
 :::info
 Note that this property only gives a guarantee on when the messages are executed, but there is no guarantee on the ordering of the responses received.
```
oggy-dfin, line 38:
```suggestion
Note that this property only gives a guarantee on when the request messages are executed, but there is no guarantee on the ordering of the responses received.
```

docs/references/message-execution-properties.mdx
oggy-dfin: I wouldn't say that it can choose not to reply; if it does not reply, the system will generate a response for it (a canister error). But it can delay producing the response for arbitrarily long, potentially infinitely long if it's willing to put the cycles in. We can (and IMO should) say that.

oggy-dfin (approved):
LGTM, thank you for consolidating these! I've left a few suggestions, hopefully nothing controversial.

docs/references/message-execution-properties.mdx
robin-kunzler: Done, thanks!

docs/references/message-execution-properties.mdx
robin-kunzler: Done

docs/references/message-execution-properties.mdx
robin-kunzler: Agreed, I explain this in the paragraph below, using some of your wording. Intentionally keeping this vague though.

docs/references/message-execution-properties.mdx
robin-kunzler: True and agreed. Clarified this, see the previous comment, PTAL.

robin-kunzler (commented):
Thanks a lot for the reviews and good feedback @oggy-dfin and @roelstorms! I addressed everything but have one question remaining, PTAL!

docs/references/message-execution-properties.mdx
roelstorms: I'd add that it is also atomic which is not necessary for messages that are executed sequential. Messages could be executed in sequence but partially.

roelstorms: Maybe that is property 3?

robin-kunzler: Ok that works, adapted Property 1 as follows: 

> Message execution is atomic and sequential, and never parallel.

robin-kunzler:
Thanks again for the reviews and good input, proceeding to merge this.

PR #3535, opened 2024-09-25
Add ingress messages reference

robin-kunzler:
Adds a reference document explaining ingress messaging.

  side navigation bar.


docs/references/ingress-messages.mdx
```diff
@@ -0,0 +1,82 @@
...
+
+import { MarkdownChipRow } from "/src/components/Chip/MarkdownChipRow";
+
+# Ingress messages on ICP
+
+<MarkdownChipRow labels={["Intermediate","Reference", "Security"]} />
+
+## Overview
+
+Ingress messages are calls to ICP canisters that are sent by external entities, such as users, usually through an agent like [agent-js](https://github.com/dfinity/agent-js/tree/main) that runs within a web application, or [agent-rs](https://github.com/dfinity/agent-rs/tree/main) in a Rust application. See also the introductory [call overview](/docs/current/developer-docs/smart-contracts/call/overview).
```
oggy-dfin, line 13:
```suggestion
Ingress messages are calls to ICP canisters that are sent by external entities, such as users, usually through an [agent](https://internetcomputer.org/docs/current/developer-docs/developer-tools/dev-tools-overview/#agents) like [agent-js](https://github.com/dfinity/agent-js/tree/main) that runs within a web application, or [agent-rs](https://github.com/dfinity/agent-rs/tree/main) in a Rust application. See also the introductory [call overview](/docs/current/developer-docs/smart-contracts/call/overview). 
```
I linked to the agent definition elsewhere; @jessiemongeon1 , is there perhaps a more robust way to do this with mdx/docusaurus (that would detect if the link breaks)?

docs/references/ingress-messages.mdx
```diff
@@ -0,0 +1,82 @@
...
+
+In the following, we discuss ICP's ingress message APIs. While these APIs are defined in detail in the [HTTPS interface specification](/docs/current/references/ic-interface-spec#http-interface), we provide a more high level and intuitive overview. In addition to discussing the happy path, we lay a special focus on error handling. That aspect is particularly important as it can be tricky to determine if an ingress message has actually been successfully executed. In the worst case, misinterpreting errors could lead to bugs such as double spending.
+
+## Types of ingress messages
+
+ICP supports several types of calls, as defined in the [HTTPS interface specification](/docs/current/references/ic-interface-spec#http-interface). These include query and update calls. 
+
+## Query Calls
+Queries don't modify a canister's state and are answered and signed by a single replica. Responses are obtained within milliseconds, but they are typically not signed by the subnet, which implies that a malicious replica or boundary node could arbitrarily modify the response.
```
oggy-dfin, line 24:
```suggestion
Queries don't modify a canister's state and are answered and signed by a single replica. Responses can be obtained within milliseconds, but they are typically not signed by the subnet, which implies that a malicious replica or boundary node could arbitrarily modify the response. 
```
Maybe hint that we can have pretty slow queries, though maybe I'm overthinking it ;)

docs/references/ingress-messages.mdx
```diff
@@ -0,0 +1,82 @@
...
+
+Some applications may have higher authenticity requirements for query responses. For example, if a dapp queries a proposal description that the user then votes on, it may not be acceptable that a single replica or boundary node can tamper with that information, as they could trick users to vote in the attacker's favor. One can consider issuing query calls as update calls instead, see the next section. While that's easy to implement, it makes the calls significantly slower. To avoid this negative performance impact, for some use cases one can sign the data beforehand and provide [certified variables](/docs/current/developer-docs/security/security-best-practices/data-integrity-and-authenticity#certified-variables) in query responses. However, this can be complex to implement, depending on the specific use case. An example where certified variables are used is the [HTTP asset certification](/docs/current/developer-docs/security/security-best-practices/data-integrity-and-authenticity#use-http-asset-certification-and-avoid-serving-your-dapp-through-rawicp0io).
```
oggy-dfin, line 26:
```suggestion
Some applications may have higher authenticity requirements for query responses. For example, if a dapp queries a proposal description that the user then votes on, it may not be acceptable that a single replica or boundary node can tamper with that information, as they could trick users to vote in the attacker's favor. One can consider issuing query calls as update calls instead, see the next section. While that's easy to implement, it makes the calls significantly slower. To avoid this negative performance impact, for some use cases one can sign the data beforehand and provide *certified variables* (see [tutorial](https://internetcomputer.org/docs/current/tutorials/developer-journey/level-3/3.3-certified-data/), [security best practices](/docs/current/developer-docs/security/security-best-practices/data-integrity-and-authenticity#certified-variables)) in query responses. However, this can be complex to implement, depending on the specific use case. An example where certified variables are used is the [HTTP asset certification](/docs/current/developer-docs/security/security-best-practices/data-integrity-and-authenticity#use-http-asset-certification-and-avoid-serving-your-dapp-through-rawicp0io). 
```

docs/references/ingress-messages.mdx
```diff
@@ -0,0 +1,82 @@
...
+
+A [synchronous `call`](https://internetcomputer.org/docs/current/references/ic-interface-spec#http-call) follows the "call and await" pattern. In the success case, it maintains the HTTPS connection until a certified response is produced, as illustrated below. 
+
+![Dev containers 1](./_attachments/ingress_sync_call.png)
+
+Compared to the [asynchronous `call`](/docs/current/references/ic-interface-spec#http-async-call) endpoint, where the [`read_state`](/docs/current/references/ic-interface-spec#http-read-state) endpoint is used to poll for the request status and obtain the reply, the synchronous endpoint removes the need to poll for the response in the success case. 
+
+However, a response may not be produced until some implementation-specific timeout is reached. In that case, the synchronous endpoint behaves like the asynchronous one: it returns a `202` response, and the client needs to poll [read_state](/docs/current/references/ic-interface-spec#http-read-state) to get the request status and eventually obtain the reply, as illustrated below.
```
oggy-dfin, line 41: Should this be only "call v2" here? There's no async v3 call (which is a bit weird)

docs/references/ingress-messages.mdx
```diff
@@ -0,0 +1,82 @@
...
+### Update call errors and their implications
+
+A number of errors could occur in the above flow. The following are examples of errors, indicated using the numbers in the diagram:
+1. Due to misconfiguration, the agent might fail to successfully sign the update call and not even send it off.
+2. The call to the boundary node may fail, for example because the agent or the boundary node is offline and the call times out.
+3. Calls (`call` or `read_state`) to the IC node may fail, for example due to 
+   1. the node being offline
+   2. an execution pre-processing error (status `200` in case of the async call)
+   3. or the node being incapable of handling the request (status `5xx`)
+4. It could be that status responses are never received. For example, a malicious boundary node could block the status response. Thus, the agent would never see a [status like `received`, `processing`, etc.](/docs/current/references/ic-interface-spec#state-tree-request-status) upon calling `read_state`.
```
oggy-dfin, line 54:
```suggestion
4. It could be that status responses are never received. For example, the user's might lose network connectivity, or a malicious boundary node could block the status response. Thus, the agent would never see a [status like `received`, `processing`, etc.](/docs/current/references/ic-interface-spec#state-tree-request-status) upon calling `read_state`.
```

docs/references/ingress-messages.mdx
```diff
@@ -0,0 +1,82 @@
...
+
+As ICP rejects `ingress_expiry` that is too far in the future, the time window for retrying is limited. To maximize the window, one can maximize `ingress_expiry` in this case and set the timeout to `ingress_expiry` as well.
```
oggy-dfin, line 76: What timeout?

oggy-dfin (approved):
Looks very good overall, thank you! Left a few small comments and a question, PTAL

docs/references/ingress-messages.mdx
jessiemongeon1: @oggy-dfin No, there isn't any better way to link to it that would automatically detect if the link breaks. 

We've implemented a few different secondary workflows to catch broken links (hopefully) more often. These are:
- A sitemap analysis script that checks for broken links, and I run it periodically to catch any links that may have been changed. 
- Visitor traffic statistics that send me a weekly report on visitors that received a 404 page and what URL led them to a 404.

docs/references/ingress-messages.mdx
```diff
@@ -0,0 +1,82 @@
...
+
+# Ingress messages on ICP
+
+<MarkdownChipRow labels={["Intermediate","Reference", "Security"]} />
+
+## Overview
+
+Ingress messages are calls to ICP canisters that are sent by external entities, such as users, usually through an agent like [agent-js](https://github.com/dfinity/agent-js/tree/main) that runs within a web application, or [agent-rs](https://github.com/dfinity/agent-rs/tree/main) in a Rust application. See also the introductory [call overview](/docs/current/developer-docs/smart-contracts/call/overview). 
+
+For example, you can browse what ingress messages can be sent to the ckBTC ledger by browsing through its [API on the dashboard](https://dashboard.internetcomputer.org/canister/mxzaz-hqaaa-aaaar-qaada-cai). The tool also allows to send messages, using agent-js behind the scenes.
```
jessiemongeon1, line 15:
```suggestion
For example, you can browse what ingress messages can be sent to the ckBTC ledger by browsing through its [API using the dashboard](https://dashboard.internetcomputer.org/canister/mxzaz-hqaaa-aaaar-qaada-cai). The tool also allows you to send messages using agent-js behind the scenes.
```

docs/references/ingress-messages.mdx
```diff
@@ -0,0 +1,82 @@
...
+
+In the following, we discuss ICP's ingress message APIs. While these APIs are defined in detail in the [HTTPS interface specification](/docs/current/references/ic-interface-spec#http-interface), we provide a more high level and intuitive overview. In addition to discussing the happy path, we lay a special focus on error handling. That aspect is particularly important as it can be tricky to determine if an ingress message has actually been successfully executed. In the worst case, misinterpreting errors could lead to bugs such as double spending.
```
jessiemongeon1, line 17:
```suggestion
This page discusses ICP's ingress message APIs. While these APIs are defined in detail within the [HTTPS interface specification](/docs/current/references/ic-interface-spec#http-interface), this page provides a more high-level and intuitive overview, with a special focus on error handling. That aspect is particularly important, as it can be tricky to determine if an ingress message has actually been successfully executed. Misinterpreting errors could lead to bugs such as double spending.
```

docs/references/ingress-messages.mdx
```diff
@@ -0,0 +1,82 @@
...
+
+## Types of ingress messages
+
+ICP supports several types of calls, as defined in the [HTTPS interface specification](/docs/current/references/ic-interface-spec#http-interface). These include query and update calls. 
+
+## Query Calls
```
jessiemongeon1, line 23:
```suggestion
### Query calls
```

docs/references/ingress-messages.mdx
jessiemongeon1, line 24:
```suggestion
Queries don't modify a canister's state and are answered and signed by a single replica. Responses are obtained within milliseconds, but they are typically not signed by the subnet, meaning a malicious replica or boundary node could arbitrarily modify the response. 
```

docs/references/ingress-messages.mdx
jessiemongeon1, line 26:
```suggestion
Some applications may have higher authenticity requirements for query responses. For example, if a dapp queries a proposal description that the user then votes on, it may not be acceptable that the result could be tampered with, as it could trick users to vote in an attacker's favor. As an alternative, you can issue query calls as update calls instead. Although implementing this alternative is easy, it is significantly slow.

Another alternative approach is to sign the data beforehand and provide [certified variables](/docs/current/developer-docs/security/security-best-practices/data-integrity-and-authenticity#certified-variables) in query responses. While this approach has better performance than sending query calls as update calls, it can be complex to implement depending on the specific use case. An example where certified variables are used is within the [HTTP asset certification library](/docs/current/developer-docs/security/security-best-practices/data-integrity-and-authenticity#use-http-asset-certification-and-avoid-serving-your-dapp-through-rawicp0io). 
```

docs/references/ingress-messages.mdx
```diff
@@ -0,0 +1,82 @@
...
+
+Error handling for queries is typically simple: on error, one can simply retry until a successful response is received. 
+
+## Update calls
+Updates can modify a canister's state and thus go through consensus, typically taking 1-3 seconds. They return certified responses (signed by the subnet), providing authenticity within the subnet's corruption bounds.
```
jessiemongeon1, line 31:
```suggestion
Updates can modify a canister's state and thus must go through consensus, typically taking 1-3 seconds. They return certified responses (signed by the subnet), providing authenticity within the subnet's corruption bounds.
```

docs/references/ingress-messages.mdx
jessiemongeon1, line 41:
```suggestion
However, a response may not be produced until some implementation-specific timeout is reached. In that case, the synchronous endpoint behaves like the asynchronous one: it returns a `202` response, and the client must poll [`read_state`](/docs/current/references/ic-interface-spec#http-read-state) to get the request status and eventually obtain the reply, as illustrated below.
```

docs/references/ingress-messages.mdx
```diff
@@ -0,0 +1,82 @@
...
+Compared to the [asynchronous `call`](/docs/current/references/ic-interface-spec#http-async-call) endpoint, where the [`read_state`](/docs/current/references/ic-interface-spec#http-read-state) endpoint is used to poll for the request status and obtain the reply, the synchronous endpoint removes the need to poll for the response in the success case. 
+
+However, a response may not be produced until some implementation-specific timeout is reached. In that case, the synchronous endpoint behaves like the asynchronous one: it returns a `202` response, and the client needs to poll [read_state](/docs/current/references/ic-interface-spec#http-read-state) to get the request status and eventually obtain the reply, as illustrated below.
+
+![Dev containers 1](./_attachments/ingress_async_call.png)
+
+### Update call errors and their implications
+
+A number of errors could occur in the above flow. The following are examples of errors, indicated using the numbers in the diagram:
+1. Due to misconfiguration, the agent might fail to successfully sign the update call and not even send it off.
```
jessiemongeon1, line 48:
```suggestion
1. Due to misconfiguration, the agent might fail to successfully sign the update call and not even send it.
```

docs/references/ingress-messages.mdx
```diff
@@ -0,0 +1,82 @@
...
+2. The call to the boundary node may fail, for example because the agent or the boundary node is offline and the call times out.
```
jessiemongeon1, line 49:
```suggestion
2. The call to the boundary node may fail, because, for example, the agent or the boundary node is offline and the call times out.
```

docs/references/ingress-messages.mdx
```diff
@@ -0,0 +1,82 @@
...
+3. Calls (`call` or `read_state`) to the IC node may fail, for example due to
```
jessiemongeon1, line 50:
```suggestion
3. Calls (`call` or `read_state`) to the IC node may fail, due to possible reasons such as:
```

docs/references/ingress-messages.mdx
```diff
@@ -0,0 +1,82 @@
...
+
+### Update call errors and their implications
+
+A number of errors could occur in the above flow. The following are examples of errors, indicated using the numbers in the diagram:
+1. Due to misconfiguration, the agent might fail to successfully sign the update call and not even send it off.
+2. The call to the boundary node may fail, for example because the agent or the boundary node is offline and the call times out.
+3. Calls (`call` or `read_state`) to the IC node may fail, for example due to 
+   1. the node being offline
+   2. an execution pre-processing error (status `200` in case of the async call)
+   3. or the node being incapable of handling the request (status `5xx`)
```
jessiemongeon1, line 53:
```suggestion
   1. The node is offline.
   2. An execution pre-processing error (status `200` in case of the async call) occurs.
   3. The node is incapable of handling the request (status `5xx`).
```

docs/references/ingress-messages.mdx
```diff
@@ -0,0 +1,82 @@
...
+4. It could be that status responses are never received. For example, a malicious boundary node could block the status response. Thus, the agent would never see a [status like `received`, `processing`, etc.](/docs/current/references/ic-interface-spec#state-tree-request-status) upon calling `read_state`.
+5. The certificate in the `read_state` response could be invalid. For example, it could be malformed or have an invalid signature.
+6. Finally, the call could be `rejected` (see [rejected status](/docs/current/references/ic-interface-spec#state-tree-request-status) and [reject codes](/docs/current/references/ic-interface-spec#reject-codes)).
+
+When seeing errors 2-5, the call may still have been successfully executed. For example, a malicious boundary node or node could return an error, but still successfully execute the request. Only if a `rejected` response is received along with a valid certificate, the subnet guarantees that the call hasn’t been executed and never will be.
```
jessiemongeon1, line 58:
```suggestion
When seeing errors 2-5, the call may still be successfully executed. For example, a malicious boundary node or node could return an error but still successfully execute the request. Only if a `rejected` response is received along with a valid certificate, the subnet guarantees that the call hasn’t been executed and never will be.
```

docs/references/ingress-messages.mdx
```diff
@@ -0,0 +1,82 @@
...
+
+Some operations, such as transfers, must not be executed multiple times. If a call was `rejected` as described above, it is safe to retry the operation. In all other error cases however, retrying can be risky because the call could still successfully execute, potentially leading to severe security vulnerabilities such as double spending.
+
+In general there are cases where it is impossible to determine if a call was successful. For example, suppose
```
jessiemongeon1, line 62:
```suggestion
In general, there are cases where it is impossible to determine if a call was successful. For example, suppose:
```

docs/references/ingress-messages.mdx
```diff
@@ -0,0 +1,82 @@
...
+* The client loses its connection until the request status has been removed from the state tree. (Recall that ICP will remove the request from the system state tree some time after ingress expiry.)
+* There is no application specific way to tell if the call successfully executed. For example, the canister does not provide a way to find out if the call happened.
+
+In this situation, it is indistinguishable if the call was successful, but all information has already been removed from the system state tree, or the call never made it into the state tree in the first place and hasn't been executed. 
+
+### Recommendations and best practices for secure update call handling
+
+The best way to handle errors as discussed above is to implement safe retries using idempotent calls. Idempotency and several patterns to achieve it are discussed in the [best practices on safe retries and idempotency](/docs/current/developer-docs/smart-contracts/best-practices/idempotency).
```
jessiemongeon1, line 70:
```suggestion
The best way to handle the aforementioned errors is to implement safe retries using idempotent calls. Idempotency and several patterns to achieve it are discussed in the [best practices on safe retries and idempotency](/docs/current/developer-docs/smart-contracts/best-practices/idempotency).
```

docs/references/ingress-messages.mdx
```diff
@@ -0,0 +1,82 @@
...
+
+ICP provides a built-in deduplication feature. Namely, if two calls with identical [request id](/docs/current/references/ic-interface-spec#request-ids) are made, the transaction is processed at most once. For two calls to have the same `request_id`, they must have the same values for all fields `request_type`, `sender`, `ingress_expiry`, `nonce`, `canister_id`, `method_name`, and `arg`. This gives idempotency, but only until [ingress_expiry](/docs/current/references/ic-interface-spec#authentication) is reached, because after the message expires, ICP [may remove the request_id from the system state tree](/docs/current/references/ic-interface-spec#state-tree-request-status). Note that the same request will not be accepted later, as it expired. Thus, such requests can be safely retried analogously to how it is done for [idempotent calls](/docs/current/developer-docs/smart-contracts/best-practices/idempotency#idempotent-canister-apis), see the illustration on that page.
```
jessiemongeon1, line 74:
```suggestion
ICP provides a built-in deduplication feature. Namely, if two calls with identical [request id](/docs/current/references/ic-interface-spec#request-ids) are made, the transaction is processed at most once. For two calls to have the same `request_id`, they must have the same values for all fields `request_type`, `sender`, `ingress_expiry`, `nonce`, `canister_id`, `method_name`, and `arg`. This gives idempotency, but only until the [`ingress_expiry`](/docs/current/references/ic-interface-spec#authentication) is reached, because after the message expires, ICP [may remove the `request_id` from the system state tree](/docs/current/references/ic-interface-spec#state-tree-request-status). Note that the same request will not be accepted later, as it expired. Thus, such requests can be safely retried analogously to how it is done for [idempotent calls](/docs/current/developer-docs/smart-contracts/best-practices/idempotency#idempotent-canister-apis), see the illustration on that page.
```

docs/references/ingress-messages.mdx
```diff
@@ -0,0 +1,82 @@
...
+
+As ICP rejects `ingress_expiry` that is too far in the future, the time window for retrying is limited. To maximize the window, one can maximize `ingress_expiry` in this case and set the timeout to `ingress_expiry` as well.
+
+There are two options for retries:
+* Only poll `read_state` until `ingress_expiry` is reached. Note that this may be hard if the IC time and local time differ.
+* Poll `read_state` for e.g. `30s` and then retry the `call`, loop until `ingress_expiry` (e.g. `5min`). This can be beneficial because a call that didn’t make it would be retried.
+
+If the timeout is reached, there is no way for the client to determine if the transaction was successful, unless the canister’s API provides this information. In this case, after the message has expired, the user must be instructed to manually check the status of the transaction, using application specific APIs. There can be (hopefully exotic) cases where the APIs don't allow to check this, in which case the user simplay can’t tell if the transaction happened. Ideally, timeouts should be rare and not occur during normal operation.
```
jessiemongeon1, line 82:
```suggestion
If the timeout is reached, there is no way for the client to determine if the transaction was successful unless the canister’s API provides this information. In this case, after the message has expired, the user must be instructed to manually check the status of the transaction using application-specific APIs. There can be (hopefully uncommon) cases where the APIs don't allow checking this, in which case the user simply can’t tell if the transaction happened. Ideally, timeouts should be rare and not occur during normal operation.```

File: docs/references/ingress-messages.mdx (hunk 1 above)
Comment (robin-kunzler) [reply docs/references/ingress-messages.mdx]: Thanks, good point, added the link as suggested!

File: docs/references/ingress-messages.mdx (hunk 7 above)
Comment (robin-kunzler) [reply docs/references/ingress-messages.mdx]: Thanks, done.

File: docs/references/ingress-messages.mdx (hunk 8 above)
Comment (robin-kunzler) [reply docs/references/ingress-messages.mdx]: Done, thanks!

File: docs/references/ingress-messages.mdx (hunk 9 above)
Comment (robin-kunzler) [reply docs/references/ingress-messages.mdx]: Done, thanks!

File: docs/references/ingress-messages.mdx (hunk 2 above)
Comment (robin-kunzler) [reply docs/references/ingress-messages.mdx]: Adapted as you suggested. 

> Maybe hint that we can have pretty slow queries

Do you mean they could be slow in case complex computation is performed, or in case the system is under heavy load?

Comment (robin-kunzler) [reply docs/references/ingress-messages.mdx]: Thanks, done.

File: docs/references/ingress-messages.mdx (hunk 3 above)
Comment (robin-kunzler) [reply docs/references/ingress-messages.mdx]: Thanks for adding the tutorial link, done.

Comment (robin-kunzler) [reply docs/references/ingress-messages.mdx]: Very nice, thanks for streamlining this, it reads much better. 

Merged this with Ogi's input above.

File: docs/references/ingress-messages.mdx (hunk 4 above)
Comment (robin-kunzler) [reply docs/references/ingress-messages.mdx]: I'm not sure I understand: on timeout, v3 also returns a 202 (like v2) and in both cases we need to poll read_state, right? What change are you suggesting?

File: docs/references/ingress-messages.mdx (hunk 13 above)
Comment (robin-kunzler) [reply docs/references/ingress-messages.mdx]: Thanks, done

File: docs/references/ingress-messages.mdx (hunk 5 above)
Comment (robin-kunzler) [reply docs/references/ingress-messages.mdx:56]: Thanks, done.

File: docs/references/ingress-messages.mdx (hunk 15 above)
Comment (robin-kunzler) [reply docs/references/ingress-messages.mdx]: Thanks, done.

File: docs/references/ingress-messages.mdx (hunk 16 above)
Comment (robin-kunzler) [reply docs/references/ingress-messages.mdx]: Thanks, done.

File: docs/references/ingress-messages.mdx (hunk 17 above)
Comment (robin-kunzler) [reply docs/references/ingress-messages.mdx]: Nice, thanks.

File: docs/references/ingress-messages.mdx (hunk 18 above)
Comment (robin-kunzler) [reply docs/references/ingress-messages.mdx]: Done.

File: docs/references/ingress-messages.mdx (hunk 6 above)
Comment (robin-kunzler) [reply docs/references/ingress-messages.mdx]: Oh that was uncler, fixed it by saying: 
``` 
(...) and time out the retries after `ingress_expiry` as well.
```

File: docs/references/ingress-messages.mdx (hunk 19 above)
Comment (robin-kunzler) [reply docs/references/ingress-messages.mdx]: Thanks, done.

**Review by robin-kunzler — COMMENTED** [review]
Thanks for the reviews and good input @jessiemongeon1 and @oggy-dfin ! 

Addressed everything, only have two clarifying questions to you Ogi!

File: docs/references/_attachments/ingress_async_call.png
Comment (roelstorms) [inline docs/references/_attachments/ingress_async_call.png:1]: Isn't V3 the sync api or does it also have async calls?

File: docs/references/ingress-messages.mdx [hunk 20]
```diff
@@ -0,0 +1,84 @@
...
+   2. An execution pre-processing error (status `200` in case of the async call) occurs.
+   3. The node is incapable of handling the request (status `5xx`).
+4. It could be that status responses are never received. For example, a malicious boundary node could block the status response. Thus, the agent would never see a [status like `received`, `processing`, etc.](/docs/current/references/ic-interface-spec#state-tree-request-status) upon calling `read_state`.
+5. The certificate in the `read_state` response could be invalid. For example, it could be malformed or have an invalid signature.
+6. Finally, the call could be `rejected` (see [rejected status](/docs/current/references/ic-interface-spec#state-tree-request-status) and [reject codes](/docs/current/references/ic-interface-spec#reject-codes)).
+
+When seeing errors 2-5, the call may still be successfully executed. For example, the user's might lose network connectivity, or a malicious boundary node or node could return an error but still successfully execute the request. Only if a `rejected` response is received along with a valid certificate, the subnet guarantees that the call hasn’t been executed and never will be.
```
roelstorms, line 60:
```suggestion
When seeing errors 2-5, the call may still be successfully executed. For example, the user's might lose network connectivity, or a malicious boundary node or replica node could return an error but still successfully execute the request. Only if a `rejected` response is received along with a valid certificate, the subnet guarantees that the call hasn’t been executed and never will be.
```

docs/references/ingress-messages.mdx
```diff
@@ -0,0 +1,84 @@
...
+* There is no application specific way to tell if the call successfully executed. For example, the canister does not provide a way to find out if the call happened.
+
+In this situation, it is indistinguishable if the call was successful, but all information has already been removed from the system state tree, or the call never made it into the state tree in the first place and hasn't been executed. 
+
+### Recommendations and best practices for secure update call handling
+
+The best way to handle the aforementioned errors is to implement safe retries using idempotent calls. Idempotency and several patterns to achieve it are discussed in the [best practices on safe retries and idempotency](/docs/current/developer-docs/smart-contracts/best-practices/idempotency).
+
+It may not always be possible to make call endpoints idempotent. For example, an endpoint that is not under the developer’s control may simply not have this property. In this case, we still suggest retrying as a best effort. However, the guarantees in this case are weaker compared to using idempotent calls.
```
roelstorms, line 74: WDYM with "retrying as a best effort"?
If the endpoint can't be made idempotent, we still advise to retry which might lead to vulnerabilities?

roelstorms (approved):
Left a few comments. Otherwise LGTM.

docs/references/ingress-messages.mdx
oggy-dfin: I suggested "can be" to imply that it's possible that they are processed in milliseconds, but that they can also be slow (in both cases that you mention). So no further changes needed from my side.

docs/references/ingress-messages.mdx
oggy-dfin: Ah, I didn't realize that 202 applied to `v3` as well - sorry for the noise!

jessiemongeon1 (approved):

docs/references/ingress-messages.mdx
```diff
@@ -10,25 +10,27 @@ import { MarkdownChipRow } from "/src/components/Chip/MarkdownChipRow";
 
 ## Overview
 
-Ingress messages are calls to ICP canisters that are sent by external entities, such as users, usually through an agent like [agent-js](https://github.com/dfinity/agent-js/tree/main) that runs within a web application, or [agent-rs](https://github.com/dfinity/agent-rs/tree/main) in a Rust application. See also the introductory [call overview](/docs/current/developer-docs/smart-contracts/call/overview). 
+Ingress messages are calls to ICP canisters that are sent by external entities, such as users, usually through an [agent](/docs/current/developer-docs/developer-tools/dev-tools-overview/#agents) like [agent-js](https://github.com/dfinity/agent-js/tree/main) that runs within a web application, or [agent-rs](https://github.com/dfinity/agent-rs/tree/main) in a Rust application. See also the introductory [call overview](/docs/current/developer-docs/smart-contracts/call/overview).
```
yvonneanne, line 13: Are ingress messages limited to calls everywhere in our docs or are queries also ingress messages in some places? For sure our code is a bit ambiguous, e.g., https://github.com/dfinity/ic/blob/master/rs/validator/ingress_message/src/lib.rs

(if we're sure we use ingress message for calls only in our docs then that's great, I'm not implying we should rename code or the docs, I simply don't know if we're using it consistently)

yvonneanne: I see that we later use it for queries as well... So maybe should use "request" here instead of "call"?

docs/references/ingress-messages.mdx
```diff
@@ -38,45 +40,45 @@ A [synchronous `call`](https://internetcomputer.org/docs/current/references/ic-i
...
 
-If the timeout is reached, there is no way for the client to determine if the transaction was successful, unless the canister’s API provides this information. In this case, after the message has expired, the user must be instructed to manually check the status of the transaction, using application specific APIs. There can be (hopefully exotic) cases where the APIs don't allow to check this, in which case the user simplay can’t tell if the transaction happened. Ideally, timeouts should be rare and not occur during normal operation.
\ No newline at end of file
+If the timeout is reached, there is no way for the client to determine if the transaction was successful unless the canister’s API provides this information. In this case, after the message has expired, the user must be instructed to manually check the status of the transaction using application-specific APIs. There can be (hopefully uncommon) cases where the APIs don't allow checking this, in which case the user simply can’t tell if the transaction happened. Ideally, timeouts should be rare and not occur during normal operation.
```
yvonneanne, line 84:
```suggestion
If the expiry is reached, there is no way for the client to determine if the call was successful unless the canister’s API provides this information. In this case, after the message has expired, the user must be instructed to manually check the status of the call using application-specific APIs. There can be (hopefully uncommon) cases where the APIs don't allow checking this, in which case the user simply can’t tell if the message has been executed. Ideally, applications set expiry values such that under typical load timeouts are rare.
```

docs/references/ingress-messages.mdx
```diff
@@ -0,0 +1,84 @@
...
+
+ICP provides a built-in deduplication feature. Namely, if two calls with identical [request id](/docs/current/references/ic-interface-spec#request-ids) are made, the transaction is processed at most once. For two calls to have the same `request_id`, they must have the same values for all fields `request_type`, `sender`, `ingress_expiry`, `nonce`, `canister_id`, `method_name`, and `arg`. This gives idempotency, but only until the [`ingress_expiry`](/docs/current/references/ic-interface-spec#authentication) is reached, because after the message expires, ICP [may remove the `request_id` from the system state tree](/docs/current/references/ic-interface-spec#state-tree-request-status). Note that the same request will not be accepted later, as it expired. Thus, such requests can be safely retried analogously to how it is done for [idempotent calls](/docs/current/developer-docs/smart-contracts/best-practices/idempotency#idempotent-canister-apis), see the illustration on that page.
```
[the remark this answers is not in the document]
yvonneanne: replace transaction with message? (occurs in a few more places)

docs/references/ingress-messages.mdx
```diff
@@ -0,0 +1,84 @@
...
+
+As ICP rejects `ingress_expiry` that is too far in the future, the time window for retrying is limited. To maximize the window, one can maximize `ingress_expiry` in this case and time out the retries after `ingress_expiry` as well.
+
+There are two options for retries:
```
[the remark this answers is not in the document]
yvonneanne: do you mean the sdk team? (they are responsible for the agents).This kind of information is definitely helpful for them, they are not aware of all of the facts discussed here (I don't blame them, not all of this is obvious).

docs/references/ingress-messages.mdx
robin-kunzler: Thanks, done!

docs/references/ingress-messages.mdx
robin-kunzler: Good point, this is misleading. What I meant is essentially what is described below, i.e. retrying with the same request ID. I changed the wording here to explain this better, PTAL!

robin-kunzler:
Thanks for the additional feedback and input @roelstorms and @yvonneanne ! I addressed everything. 
* @roelstorms since you approved I took the liberty to resolve your comments after addressing them. But if you have anything to add please re-open them!
* @yvonneanne could you PTAL? If you're happy with the page I'd proceed and merge this :-)

docs/references/ingress-messages.mdx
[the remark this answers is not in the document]
yvonneanne: We definitely should change this behavior!!! So the agents stop polling read_state when the IC time is at expiry even when obtaining a certificate with "processing" state back?
cc @THLO

docs/references/ingress-messages.mdx
```diff
@@ -0,0 +1,84 @@
...
+* Only poll `read_state` until `ingress_expiry` is reached. Note that this may be hard if the IC time and local time differ. This is currently the behaviour of agent-js and agent-rs.
+* Poll `read_state` for e.g. `30s` and then retry the `call`, loop until `ingress_expiry` (e.g. `5min`). This can be beneficial because a call that didn’t make it would be retried.
+
+If the expiry is reached, there is no way for the client to determine if the call was successful unless the canister’s API provides this information. In this case, after the message has expired, the user must be instructed to manually check the status of the call using application-specific APIs. There can be (hopefully uncommon) cases where the APIs don't allow checking this, in which case the user simply can’t tell if the message has been executed. Ideally, applications set expiry values such that under typical load timeouts are rare.
```
yvonneanne, line 84: This is not entirely correct, expiry only determines when the canister execution can start. If processing takes 10min, then the response will be available after 10min for another 5min (assuming there is space in the ingress history, if not then it does not hold even before expiry is reached, iirc).

Maybe rephrase as follows:

```suggestion
Some time after expiry is reached, there is no way for the client to determine if the call was successful unless the canister’s API provides this information. In this case, the client can check the status of the call using application-specific APIs. There can be (hopefully uncommon) cases where the APIs don't allow checking this, in which case the user simply can’t tell if the message has been executed. Ideally, applications set expiry values such that under typical load timeouts are rare.
```

When do you use "user" and when "client"? Imho, checking the status does not need to be manual. If it's crucial for the app behavior, then the client/user-side code should do this automatically

robin-kunzler: Thanks, applied the suggestion!

> When do you use "user" and when "client"? Imho, checking the status does not need to be manual. If it's crucial for the app behavior, then the client/user-side code should do this automatically

"user" indicates the end user, and "client" can also be code such as an agent. I think here it's better to talk about the client. Changed "user" to "client" in the sentence above.

docs/references/ingress-messages.mdx
```diff
@@ -0,0 +1,84 @@
...
+
+As ICP rejects `ingress_expiry` that is too far in the future, the time window for retrying is limited. To maximize the window, one can maximize `ingress_expiry` in this case and time out the retries after `ingress_expiry` as well.
+
+There are two options for retries:
+* Only poll `read_state` until `ingress_expiry` is reached. Note that this may be hard if the IC time and local time differ. This is currently the behaviour of agent-js and agent-rs.
+* Poll `read_state` for e.g. `30s` and then retry the `call`, loop until `ingress_expiry` (e.g. `5min`). This can be beneficial because a call that didn’t make it would be retried.
```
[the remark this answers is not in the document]
robin-kunzler: Oh very good point, thanks a lot! I adapted the condition, PTAL!

robin-kunzler (commented):
Good catches @yvonneanne , thanks. Addressed your comments, PTAL!

docs/references/ingress-messages.mdx
```diff
@@ -0,0 +1,88 @@
...
+
+As ICP rejects `ingress_expiry` that is too far in the future, the time window for retrying is limited. To maximize the window, one can maximize `ingress_expiry` in this case and time out the retries after `ingress_expiry` as well.
+
+There are two options for retries:
+* Submit the `call` once, and poll `read_state` while at least one of the following conditions is met: 
+  * `ingress_expiry` is not reached.
+  * The call status is `processing`.
+
+  Evaluating the first condition may be hard if the IC time and local time differ. Currently, agent-js and agent-rs by default poll until `ingress_expiry` is reached.
```
yvonneanne, line 85:
```suggestion
  Evaluating the first condition may be hard if the subnet time and local time of the user differ. Currently, agent-js and agent-rs by default poll until `ingress_expiry` is reached. 
```
since every subnet can have a different time, it's better to use subnet time here.

docs/references/ingress-messages.mdx
```diff
@@ -0,0 +1,88 @@
...
+
+As ICP rejects `ingress_expiry` that is too far in the future, the time window for retrying is limited. To maximize the window, one can maximize `ingress_expiry` in this case and time out the retries after `ingress_expiry` as well.
+
+There are two options for retries:
+* Submit the `call` once, and poll `read_state` while at least one of the following conditions is met: 
+  * `ingress_expiry` is not reached.
+  * The call status is `processing`.
+
+  Evaluating the first condition may be hard if the IC time and local time differ. Currently, agent-js and agent-rs by default poll until `ingress_expiry` is reached. 
+* Do the same as above, but resubmit the initiall `call` (same `request_id`) e.g. every `30s`. This can be beneficial because a call that didn’t make it is retried.
```
yvonneanne, line 86:
```suggestion
* Do the same as above, but resubmit the initiall `call` (same `request_id`) e.g. every `30s` until expiry is reached by the subnet. This can be beneficial because a call that didn’t make it is retried.
```

robin-kunzler (commented):
@yvonneanne Thanks again for the additional input. Addressed the comments, could you PTAL?

docs/references/ingress-messages.mdx
```diff
@@ -0,0 +1,94 @@
...
+
+As ICP rejects `ingress_expiry` that is too far in the future, the time window for retrying is limited. To maximize the window, one can maximize `ingress_expiry` in this case and time out the retries after `ingress_expiry` as well.
+
+In summary, there are two options for a client:
+* _No retries, only poll `read_state`:_ Submit the `call` once, and poll `read_state` until at least one of the following conditions is met:
+  * A certified response was produced. That is, the call status is `replied`, `rejected`, or `done`.
+  * `ingress_expiry` is reached, and the call status is not `received` or `processing`.
```
yvonneanne, line 83:
```suggestion
  *  the subnet time exceeds `ingress_expiry`, and the call status is not `processing`.
```
if the message is received at this point in time, the canister will not start processing it later.

docs/references/ingress-messages.mdx
```diff
@@ -0,0 +1,94 @@
...
+  
+  Evaluating the second condition may be hard if the subnet time and local time of the user differ. 
+* _Retry the call while polling `read_state`_ Do the same as above, but resubmit the initial `call` (same `request_id`) periodically (e.g. every `30s`) until at least one of the following conditions is met: 
+  * The call is processing or a certified response was produced. That is, the call status is `received`, `processing`, `replied`, `rejected`, or `done`.
```
yvonneanne, line 87:
```suggestion
  * The call is known to the system. That is, the call status is `received`, `processing`, `replied`, `rejected`, or `done`.
```

Received is not considered a response, is it? If it is, then L82 needs to be rewritten...

docs/references/ingress-messages.mdx
```diff
@@ -0,0 +1,94 @@
...
+  * `ingress_expiry` is reached.
+
+  This can be beneficial because a call that didn’t make it is retried. For example, if the first call was dropped due to a network issue or by a malicious replica, a subsequent call could make it through e.g. if it is successfully routed to an honest replica. 
+
+At the time of writing, both agent-js (v2.1.3) and agent-rs (v0.39.1) don't retry `call`, so they follow the first option above. By default, agent-js uses `ingress_expiry` of five minutes and also polls for five minutes. agent-rs sets a default `ingress_expiry` of four minutes and polls for five minutes, but whenever it observes a `received` or `processing` status, it resets the timer to allow for an additional four minutes of polling.
```
yvonneanne, line 92: agent-js uses four minutes as well (5min - 1min) https://github.com/dfinity/agent-js/blob/main/packages/agent/src/agent/http/transforms.ts#L31 when setting the expiry. I don't know where in the code the polling is stopped.

robin-kunzler: Oh good catch, thanks. I missed the additional logic in the constructor. Adapted to 4 min. 

> I don't know where in the code the polling is stopped.

FYI this is done [here](https://github.com/dfinity/agent-js/blob/12b851f01fb0a0fc911089f37282adcb59e991bb/packages/agent/src/polling/index.ts#L28). [This is the default strategy.](https://github.com/dfinity/agent-js/blob/12b851f01fb0a0fc911089f37282adcb59e991bb/packages/agent/src/polling/strategy.ts#L19-L21)

robin-kunzler (commented):
Thanks for the review and additional feedback @yvonneanne . Addressed everything, PTAL!

yvonneanne (approved):
Can't find anything else :) 

Nice one!

PR #3577, opened 2024-10-07
chore: Add a compute allocation note

berestovskyy:
Make it clear that increasing allocation requires sufficient amount of cycles.


manudrijvers (approved):
Thanks!

PR #3708, opened 2024-11-08
Update resource limits

derlerd-dfinity:
Update the docs to appropriately reflect the total memory available per subnet and the Wasm stable memory limit after the recent [bump](https://github.com/dfinity/ic/commit/41af49d62b263bc12a14f288dfc885d8de7cd7e5) of those values.


derlerd-dfinity:
I think we should perhaps wait with merging until the proposal for the replica version containing the bump is accepted and rollout starts.

@dfinity/editorial Once the PR has the required approvals I can monitor things and hit merge as soon as the conditions above are met.

jessiemongeon1 (approved):

pakhomov-dfinity (approved):

PR #3734, opened 2024-11-13
chore: move Interface Specification to this repository

mraszyk:
This PR moves the Interface Specification to be hosted in this repository instead of pulling it as a git submodule pointing to a separate repository dfinity/interface-spec.


Dfinity-Bjoern (approved):
Thank you @mraszyk for moving the spec!

jessiemongeon1 (approved):

PR #3756, opened 2024-11-15
feat: specify ic0.canister_liquid_cycle_balance128

mraszyk:
This PR specifies a new system API ic0.canister_liquid_cycle_balance128 returning the current amount of cycles that is available for spending in calls and execution. This way, canisters can bound the amount of cycles attached to a call to prevent spurious traps when calling ic0.call_cycles_add or ic0.call_cycles_add128.


mraszyk:
Also useful for this: https://dfinity.atlassian.net/browse/PEN-426

Dfinity-Bjoern (approved):

PR #3757, opened 2024-11-15
feat: allow canister http outcalls to specify transform of another canister

mraszyk:
This PR allows canister http outcalls to specify the transform function exposed by another canister (on the same subnet).


jessiemongeon1 (approved):

mraszyk:
Canister http will be [disabled](https://github.com/dfinity/motoko-playground/pull/306) in the playground which seems to be the only user of this feature.

PR #3760, opened 2024-11-15
[FINAL] feat: Allow accepting and burning cycles in replicated queries

mraszyk:
Given a previous change that introduced formally the concepts of replicated and non-replicated queries, we can now extend the spec to allow accepting and burning cycles in replicated queries. This allows for creating reasonable business models for canisters by requiring callers in replicated mode to always make payments for accessing their endpoints.


jessiemongeon1 (approved):

mraszyk:
Merged into replica: https://github.com/dfinity/ic/commit/178aceab8793f1aedff9ade639a97e4b3b273669

Dfinity-Bjoern (approved):

PR #3761, opened 2024-11-15
feat: on low wasm memory hook

mraszyk:
This proposal is based on this [forum post](https://forum.dfinity.org/t/canister-lifecycle-hooks/17089) and has already been approved by [motion proposal 106146](https://dashboard.internetcomputer.org/proposal/106146).

Canister developers have to actively monitor their canisters to know if they are low on wasm memory. If detected too late, a canister can be completely stuck and forever un-upgradable.

To address this, we introduce a hook called on_low_wasm_memory. When defined, it is triggered whenever the canister's memory usage exceeds some developer-defined threshold.


mraszyk:
The hook should not run after an upgrade/reinstall/uninstall/install if the condition is not satisfied after the upgrade (although the condition was satisfied before the upgrade and the hook did not execute yet).

mraszyk:
> Because the hook condition is checked whenever the canister tries to grow wasm memory, can it ever happen that the canister does not try to grow wasm memory when running `upgrade/reinstall/uninstall/install`?

Yes, e.g., if the canister does not have any pre- and post-upgrade hooks (in which case there's no wasm execution during canister installation).

docs/references/ic-interface-spec.md
```diff
@@ -1390,6 +1394,18 @@ While an implementation will likely try to keep the interval between the value o
...
 
+#### On Low Wasm Memory {#on-low-wasm-memory}
+
+A canister can export a function with the name `canister_on_low_wasm_memory`, which is scheduled whenever the canister's remaining wasm memory size in bytes falls from at least a threshold `t` to strictly less than `t`.
+The threshold `t` can be defined in the [canister's settings](#ic-canister_status) and by default it is set to 0.
+
+:::note
+
+While the above function is scheduled immediately once the condition above is triggered, it may not necessarily be executed immediately if the canister does not have enough cycles.
+If the canister gets frozen immediately after the function is scheduled for execution, the function will run once the canister's unfrozen _if_ the canister's wasm memory remains above the threshold `t`.
```
luc-blaeser, line 1405:
```suggestion
If the canister gets frozen immediately after the function is scheduled for execution, the function will run once the canister's unfrozen _if_ the canister's wasm memory remains below the threshold `t`.
```
Shouldn't it mean "below threshold" meaning that the alert only trigger on unfreeze if it is still short of memory?

mraszyk: It should indeed be rephrased, but I'd suggest: _if_ the canister's remaining wasm memory size in bytes remains below the threshold `t`.

dragoljub-duric:
In slack conversation:

> 
> Luc
> Scenario:
> Canister sets memory threshold at 1 GB.
> Canister free memory is 0.99 GB -> hook is triggered
> Now, canister sets memory threshold to 0.5 GB, free memory is still 0.99 GB
> t1 > t remaining Wasm memory is greater than the Wasm memory threshold
> And now canister allocates more memory, until 0.49 GB free memory
> t2 > t1  remaining wasm memory falls below Wasm memory threshold
> Shouldn't the hook now be retriggered?
> 
> Dragoljub Duric
> That is a very good example. And I think that case may be missing. The reason is to check for the hook condition to be added after every `memory_grow` operation or after `install/reinstall/uninstall/upgrade` of the canister. But in this specific case, when memory is 0.99 GB hook will be executed and the status will be `Executed` since we do not have to check for the hook condition when we set a new threshold the status will remain `Executed`. Finally when additional memory is requested and the remaining memory becomes 0.49 GB, the condition for the hook will be satisfied, but because the status of the hook is already `Executed` we will not execute it again. It looks like we will need to add checking for the hook condition after we update `wasm_memory_threshold`.

Luc found out about this mistake, and I am going to implement the fix for it. After the `wasm_memory_threshold` is updated, I will check and update the hook condition, but I am between two alternatives for this:

1. Use the old hook status and the result of the checking for the hook condition with the new `wasm_memory_threshold` , i.e. update the hook status the same way as we do in any other case
2. Use only the result of the checking for the hook condition with the new wasm_memory_threshold, so the hook status after the update of `wasm_memory_threshold` can be either `ConditionNotSatisfied` or `Ready`. I believe this alternative will require spec changes.

The difference can be observed in the case if, before the update of the `wasm_memory_threshold` hook status was `Executed`, the old `wasm_memory_threshold` was 3GB, `remaining_wasm_memory` was 1GB, and the new `wasm_memory_threshold` is 2GB. The former approach will keep the hook status to `Executed`, while the latter will set the hook status to `Ready`.

I am leaning towards the latter option but I do not have a strong opinion. Do you have an opinion on this?

mraszyk:
> In slack conversation:
> 
> > Luc Bläser
> > Scenario:
> > Canister sets memory threshold at 1 GB.
> > Canister free memory is 0.99 GB -> hook is triggered
> > Now, canister sets memory threshold to 0.5 GB, free memory is still 0.99 GB
> > t1 > t remaining Wasm memory is greater than the Wasm memory threshold
> > And now canister allocates more memory, until 0.49 GB free memory
> > t2 > t1  remaining wasm memory falls below Wasm memory threshold
> > Shouldn't the hook now be retriggered?
> > Dragoljub Duric
> > That is a very good example. And I think that case may be missing. The reason is to check for the hook condition to be added after every `memory_grow` operation or after `install/reinstall/uninstall/upgrade` of the canister. But in this specific case, when memory is 0.99 GB hook will be executed and the status will be `Executed` since we do not have to check for the hook condition when we set a new threshold the status will remain `Executed`. Finally when additional memory is requested and the remaining memory becomes 0.49 GB, the condition for the hook will be satisfied, but because the status of the hook is already `Executed` we will not execute it again. It looks like we will need to add checking for the hook condition after we update `wasm_memory_threshold`.
> 
> Luc found out about this mistake, and I am going to implement the fix for it. After the `wasm_memory_threshold` is updated, I will check and update the hook condition, but I am between two alternatives for this:
> 
> 1. Use the old hook status and the result of the checking for the hook condition with the new `wasm_memory_threshold` , i.e. update the hook status the same way as we do in any other case
> 2. Use only the result of the checking for the hook condition with the new wasm_memory_threshold, so the hook status after the update of `wasm_memory_threshold` can be either `ConditionNotSatisfied` or `Ready`. I believe this alternative will require spec changes.
> 
> The difference can be observed in the case if, before the update of the `wasm_memory_threshold` hook status was `Executed`, the old `wasm_memory_threshold` was 3GB, `remaining_wasm_memory` was 1GB, and the new `wasm_memory_threshold` is 2GB. The former approach will keep the hook status to `Executed`, while the latter will set the hook status to `Ready`.
> 
> I am leaning towards the latter option but I do not have a strong opinion. Do you have an opinion on this?

Personally, I prefer the former approach because it is consistent in terms of updating the hook status. If we decide to potentially rerun an already executed hook after updating canister settings, then it is not clear why not rerun an already executed hook also in other special cases.

dragoljub-duric:
@mraszyk Thank you, that make sense, I implemented former approach [here](https://github.com/dfinity/ic/pull/3624).

dragoljub-duric:
@mraszyk all points you raised are addressed now. So the following release will include the ready feature.

docs/references/ic-interface-spec.md
mraszyk, line 1405:
```suggestion
If the canister gets frozen immediately after the function is scheduled for execution, the function will run once the canister's unfrozen _if_ the canister's wasm memory size in bytes remains strictly less than the threshold `t`.
```

docs/references/ic-interface-spec.md
```diff
@@ -1390,6 +1394,18 @@ While an implementation will likely try to keep the interval between the value o
...
 
+#### On Low Wasm Memory {#on-low-wasm-memory}
+
+A canister can export a function with the name `canister_on_low_wasm_memory`, which is scheduled whenever the canister's remaining wasm memory size in bytes falls from at least a threshold `t` to strictly less than `t`.
+The threshold `t` can be defined in the [canister's settings](#ic-canister_status) and by default it is set to 0.
+
+:::note
+
+While the above function is scheduled immediately once the condition above is triggered, it may not necessarily be executed immediately if the canister does not have enough cycles.
+If the canister gets frozen immediately after the function is scheduled for execution, the function will run once the canister's unfrozen _if_ the canister's wasm memory size in bytes remains strictly less than the threshold `t`.
```
dragoljub-duric, line 1405: Another case in which canister may not be executed immediately is if the used_wasm_memory becomes greater than wasm_memory_limit, in which case canister will execute the hook when wasm_memory_limit is increased here is [PR](https://github.com/dfinity/ic/pull/3631) that implemented that logic.

mraszyk: Thank you for the heads up! I've left a question on the PR. Based on your reply there and a potential discussion in the Interface Spec meeting, I'll come back to you.

mraszyk: Updated the spec: https://github.com/dfinity/portal/pull/3761/commits/6d37ea3f3ad650c44c6a7447bb040fb31124a860

Dfinity-Bjoern (approved):
Looks good, please merge in coordination with the roll-out to prod.

oggy-dfin (approved):

docs/references/ic-interface-spec.md
```diff
@@ -1390,6 +1394,17 @@ While an implementation will likely try to keep the interval between the value o
 
 :::
 
+#### On Low Wasm Memory {#on-low-wasm-memory}
+
+A canister can export a function with the name `canister_on_low_wasm_memory`, which is scheduled whenever the canister's remaining wasm memory size in bytes falls from at least a threshold `t` to strictly less than `t`.
```
dragoljub-duric, line 1399: @mraszyk I do not think this is true, hook is executed whenever after memory grows, remaining wasm memory is below `t`. Two cases: 
1. The remaining wasm memory size in bytes falls from at least a threshold `t` to strictly less than `t`
2. The remaining wasm memory size was less than threshold `t`, and after growing, the remaining wasm memory additionally decreased.

PR #3775, opened 2024-11-18
chore: improve http response specification in case of errors

mraszyk:
This PR improves the specification of HTTP responses in case of errors from `/api/v2` and `/api/v3` APIs.

Closes https://github.com/dfinity/interface-spec/issues/194


Dfinity-Bjoern (approved):
Thank you for bringing this in line with the implementation!

PR #3790, opened 2024-11-20
feat: new System API ic0.subnet_self

mraszyk:
This PR specifies a new System API `ic0.subnet_self_size` and `ic0.subnet_self_copy` that canisters can use to synchronously learn their own subnet ID. Requested on the [forum](https://forum.dfinity.org/t/how-stable-are-subnet-ranges-for-canister-ids/36437/3).


Dfinity-Bjoern:
Thank you Martin – that looks good from the spec side, let's have @dsarlis chime in for the execution team.

michael-weigelt:
Tangentially related: I assume the size part is necessary because principals are specified to have variable length. 
However, why not change the spec to commit to a fixed length for subnet_ids and simplify this interface? Same for canister_ids?

mraszyk:
> Tangentially related: I assume the size part is necessary because principals are specified to have variable length. However, why not change the spec to commit to a fixed length for subnet_ids and simplify this interface? Same for canister_ids?

For backwards-compatibility, this is practically impossible for existing APIs and thus we'd introduce an inconsistency with those by following the above suggestion.

docs/references/ic-interface-spec.md
```diff
@@ -7229,6 +7244,11 @@ ic0.canister_self_copy<es>(dst : I, offset : I, size : I) =
   if es.context = s then Trap {cycles_used = es.cycles_used;}
   copy_to_canister<es>(dst, offset, size, es.wasm_state.self_id)
 
+I ∈ {i32, i64}
+ic0.subnet_self_copy<es>(dst : I, offset : I, size : I) =
+  if es.context = s then Trap {cycles_used = es.cycles_used;}
+  copy_to_canister<es>(dst, offset, size, es.wasm_state.self_id)
```
mraszyk, line 7250:
```suggestion
  copy_to_canister<es>(dst, offset, size, es.params.sysenv.subnet_id)
```

docs/references/ic-interface-spec.md
```diff
@@ -7229,6 +7244,11 @@ ic0.canister_self_copy<es>(dst : I, offset : I, size : I) =
   if es.context = s then Trap {cycles_used = es.cycles_used;}
   copy_to_canister<es>(dst, offset, size, es.wasm_state.self_id)
 
+I ∈ {i32, i64}
+ic0.subnet_self_copy<es>(dst : I, offset : I, size : I) =
```
mraszyk, line 7248: ic0.subnet_self_size should be added

Dfinity-Bjoern (approved):

PR #3852, opened 2024-12-04
Update logo requirements

sea-snake:
Update logo requirements, include recommendations for optimizing file size.

  side navigation bar.

### DIFF
--- a/README.md
+++ b/README.md
@@ -588,11 +588,11 @@ The [Ecosystem Helper](https://mvw4g-yiaaa-aaaam-abnva-cai.icp0.io/) is an oncha
 
 ### Asset guidelines
 
-| Asset       |          | Requirements                    | Format               | Notes                                                                                   |
-|-------------|----------|---------------------------------|----------------------|-----------------------------------------------------------------------------------------|
-| logo        | required | 112x112px (1:1 aspect ratio)    | svg(recommended)/png | Optimize file size with [SVGOMG](https://svgomg.net/) or [TinyPNG](https://tinypng.com) |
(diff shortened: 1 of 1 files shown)


meodai:
Could you elaborate on why you want to remove WebP? It’s actually the smallest format of the three. If anything, I’d consider removing PNG instead and adding a max file size (KB) requirement to ensure consistency.

That said, keep in mind that not all tools export WebP natively, which might pose challenges for some contributors. Also, SVGs can be tricky since their size depends heavily on the complexity of the logo.

sea-snake:
> Could you elaborate on why you want to remove WebP? It’s actually the smallest format of the three.
My bad, didn't intend to remove webp, adding it back.

Since it's regarding logos, in most cases SVG would be a smaller file size and also scale when used in various sizes.

meodai:
@sea-snake I would typically agree but we had a few cases where the SVGs were huge. 

My proposal would be:

screenshots: jpeg / webP (specify the PX size and a max KB size maybe max 300?)
logo: webP or svg (50kb max?)

jennifertrin (approved):

sea-snake:
> @sea-snake I would typically agree but we had a few cases where the SVGs were huge.
> logo: webP or svg (50kb max?)
50kb is already too much right now with the II canister WASM size limitations, we're aiming for 4kb or less with max 10kb.

I added recommended online tooling for file size optimizations in this PR that saves roughly 70% in most cases, compared to explicit file size limits this seems like a more proactive approach to enable users to keep file sizes optimal.

meodai:
@sea-snake I don't feel strongly about it. LGTM

PR #3946, opened 2025-01-13
Add periodic confirmation to docs

LaraAS:
Modifying the docs to include periodic confirmation which will be launched today


jessiemongeon1 (approved):

marc0olo:
@LaraAS is this meanwhile in the docs? otherwise it would be good to fix the conflicts and merge this

PR #3962, opened 2025-01-20
content updates

herrpatrickmueller:
- add world computer share image to fix missing image in events from airtable
- update nes card content for world computer day on home


artkorotkikh-dfinity (approved):
LGTM

PR #4110, opened 2025-01-27
Add system API endpoints for cycle cost calculation

michael-weigelt:
[Design Doc](https://docs.google.com/document/d/1CGe779xUYU8uA37dOnG1ZXcdiMSHg2mNq43H3UaqpFs/edit?usp=sharing)

### DIFF
--- a/docs/references/ic-interface-spec.md
+++ b/docs/references/ic-interface-spec.md
@@ -1487,8 +1487,16 @@ defaulting to `I = i32` if the canister declares no memory.
     ic0.time : () -> (timestamp : i64);                                                   // *
     ic0.global_timer_set : (timestamp : i64) -> i64;                                      // I G U Ry Rt C T
     ic0.performance_counter : (counter_type : i32) -> (counter : i64);                    // * s
-    ic0.is_controller: (src : I, size : I) -> ( result: i32);                             // * s
-    ic0.in_replicated_execution: () -> (result: i32);                                     // * s
+    ic0.is_controller : (src : I, size : I) -> ( result : i32);                           // * s
+    ic0.in_replicated_execution : () -> (result : i32);                                   // * s
+    
+    ic0.cost_call : (method_name_size: i64, payload_size : i64, dst : I) -> ();           // * s
+    ic0.cost_create_canister : (dst : I) -> ();                                           // * s
+    ic0.cost_http_request(request_size : i64, max_res_bytes : i64, dst : I) -> ();        // * s
+    ic0.cost_ecdsa(src : I, size : I, dst : I) -> ();                                     // * s
+    ic0.cost_schnorr(src : I, size : I, dst : I) -> ();                                   // * s
+    ic0.cost_vetkey(src : I, size : I, dst : I) -> ();                                    // * s
+    ic0.replication_factor : (src : I, size : I) -> i32;                                  // * s
 
     ic0.debug_print : (src : I, size : I) -> ();                                          // * s
     ic0.trap : (src : I, size : I) -> ();                                                 // * s
@@ -2026,6 +2034,48 @@ When executing a query or composite query method via a query call (i.e. in non-r
     If this `certificate` includes a subnet delegation, then the id of the current canister will be included in the delegation's canister id range.
 
     This traps if `ic0.data_certificate_present()` returns `0`.
+
+### Cycle cost calculation
+
+Inter-canister calls have an implicit cost, and some calls to the management canister require the caller to attach cycles to the call explicitly.  
+The various cost factors may change over time, so the following system calls serve to give the canister programmatic, up-to-date information about the costs. 
+
+These system calls return costs in Cycles, represented by 128 bits, which will be written to the caller-specified `dst` pointer. 
+
+-   `ic0.cost_call : (method_name_size: i64, payload_size : i64, dst : I) -> () `I ∈ {i32, i64}`
+
+    The cost of an inter-canister call. `method_name_size` is the byte length of the method name, and `payload_size` is the length of the encoded argument to the method.  
+
+-   `ic0.cost_create_canister : (dst : I) -> (); `I ∈ {i32, i64}`
+
+    The cost of creating a canister on the same subnet as the calling canister via [`create_canister`](#ic-create_canister). Note that canister creation via a call to the CMC can have a different cost if the target subnet has a different replication factor. In order to facilitate conversions, see `ic0.replication_factor`.
+
+-   `ic0.cost_http_request(request_size : i64, max_res_bytes : i64, dst : I) -> (); `I ∈ {i32, i64}`
+
+    The cost of a request via [`http_request`](#ic-http_request). `request_size` is the sum of the byte lengths of the following components of an http request: 
+    - url
+    - method
+    - headers - i.e., the sum of the lengths of all keys and values 
+    - body
+    - transform - i.e., the sum of the transform method name length and the length of the transform context
+    
+    `max_res_bytes` is the maximum response length the caller wishes to accept. 
+
+-   `ic0.cost_ecdsa(src : I, size : I, dst : I) -> (); `I ∈ {i32, i64}`
+
+-   `ic0.cost_schnorr(src : I, size : I, dst : I) -> (); `I ∈ {i32, i64}`
+
+-   `ic0.cost_vetkey(src : I, size : I, dst : I) -> (); `I ∈ {i32, i64}`
+
+    These system calls accept a key name via `src` + `size` for the specific signing scheme / key. See [`sign_with_ecdsa`](#ic-sign_with_ecdsa), [`sign_with_schnorr`](#ic-sign_with_schnorr) and [`vetkd_encrypted_key`](#ic-vetkd_encrypted_key) for more information. 
+
+    These system calls trap if `src` + `size` do not decode to a valid key name. 
+
+-   `ic0.replication_factor : (src : I, size : I) -> i32; `I ∈ {i32, i64}`
+
+    Returns the replication factor (subnet size) of the subnet identified by the `Principal` at `src` + `size`. 
+
+    This system call traps if `src` + `size` do not decode to a valid `Principal` or the given `Principal` is not a subnet.   
 
 ### Debugging aids
 


docs/references/ic-interface-spec.md
```diff
@@ -2027,6 +2035,48 @@ When executing a query or composite query method via a query call (i.e. in non-r
 
     This traps if `ic0.data_certificate_present()` returns `0`.
 
+### Cycle cost calculation {#system-api-cycle-cost}
+
+Inter-canister calls have an implicit cost, and some calls to the management canister require the caller to attach cycles to the call explicitly.  
+The various cost factors may change over time, so the following system calls serve to give the canister programmatic, up-to-date information about the costs.
```
mraszyk, line 2041:
```suggestion
The various cost factors may change over time, so the following system calls give the canister programmatic, up-to-date information about the costs.
```

docs/references/ic-interface-spec.md
```diff
@@ -2027,6 +2035,48 @@ When executing a query or composite query method via a query call (i.e. in non-r
 
     This traps if `ic0.data_certificate_present()` returns `0`.
 
+### Cycle cost calculation {#system-api-cycle-cost}
+
+Inter-canister calls have an implicit cost, and some calls to the management canister require the caller to attach cycles to the call explicitly.  
+The various cost factors may change over time, so the following system calls serve to give the canister programmatic, up-to-date information about the costs. 
+
+These system calls return costs in Cycles, represented by 128 bits, which will be written to the caller-specified `dst` pointer.
```
mraszyk, line 2043:
```suggestion
These system calls return costs in Cycles, represented by 128 bits, which will be written to the heap memory starting at offset `dst`. 
```

docs/references/ic-interface-spec.md
```diff
@@ -2027,6 +2035,48 @@ When executing a query or composite query method via a query call (i.e. in non-r
...
     This traps if `ic0.data_certificate_present()` returns `0`.
 
+### Cycle cost calculation {#system-api-cycle-cost}
+
+Inter-canister calls have an implicit cost, and some calls to the management canister require the caller to attach cycles to the call explicitly.  
+The various cost factors may change over time, so the following system calls serve to give the canister programmatic, up-to-date information about the costs. 
+
+These system calls return costs in Cycles, represented by 128 bits, which will be written to the caller-specified `dst` pointer. 
+
+-   `ic0.cost_call : (method_name_size: i64, payload_size : i64, dst : I) -> () `I ∈ {i32, i64}`
```
mraszyk, line 2045:
```suggestion
-   `ic0.cost_call : (method_name_size: i64, payload_size : i64, dst : I) -> ()`; `I ∈ {i32, i64}`
```
analogously below

docs/references/ic-interface-spec.md
```diff
@@ -2027,6 +2035,48 @@ When executing a query or composite query method via a query call (i.e. in non-r
...
+
+    The cost of an inter-canister call. `method_name_size` is the byte length of the method name, and `payload_size` is the length of the encoded argument to the method.
```
mraszyk, line 2047:
```suggestion
    The cost of an inter-canister call. `method_name_size` is the byte length of the method name, and `payload_size` is the byte length of the argument to the method. 
```
I'd drop "encoded" as it doesn't matter what the argument looks like (it can as well be plain text) and "byte length" should be unambiguous.

docs/references/ic-interface-spec.md
```diff
@@ -2027,6 +2035,48 @@ When executing a query or composite query method via a query call (i.e. in non-r
...
+
+These system calls return costs in Cycles, represented by 128 bits, which will be written to the caller-specified `dst` pointer. 
+
+-   `ic0.cost_call : (method_name_size: i64, payload_size : i64, dst : I) -> () `I ∈ {i32, i64}`
+
+    The cost of an inter-canister call. `method_name_size` is the byte length of the method name, and `payload_size` is the length of the encoded argument to the method.  
+
+-   `ic0.cost_create_canister : (dst : I) -> (); `I ∈ {i32, i64}`
+
+    The cost of creating a canister on the same subnet as the calling canister via [`create_canister`](#ic-create_canister). Note that canister creation via a call to the CMC can have a different cost if the target subnet has a different replication factor. In order to facilitate conversions, see `ic0.replication_factor`.
```
mraszyk, line 2051: the conversions from cost on a fiduciary subnet to the cost on a regular app subnet might be imprecise due to rounding: replica always scales up from 13 to n; scaling down might give a different result.

docs/references/ic-interface-spec.md
```diff
@@ -2027,6 +2035,48 @@ When executing a query or composite query method via a query call (i.e. in non-r
...
+
+-   `ic0.cost_http_request(request_size : i64, max_res_bytes : i64, dst : I) -> (); `I ∈ {i32, i64}`
+
+    The cost of a request via [`http_request`](#ic-http_request). `request_size` is the sum of the byte lengths of the following components of an http request:
```
mraszyk, line 2055:
```suggestion
    The cost of a canister http outcall via [`http_request`](#ic-http_request). `request_size` is the sum of the byte lengths of the following components of an http request: 
```

docs/references/ic-interface-spec.md
```diff
@@ -2027,6 +2035,48 @@ When executing a query or composite query method via a query call (i.e. in non-r
...
+-   `ic0.cost_http_request(request_size : i64, max_res_bytes : i64, dst : I) -> (); `I ∈ {i32, i64}`
+
+    The cost of a request via [`http_request`](#ic-http_request). `request_size` is the sum of the byte lengths of the following components of an http request: 
+    - url
+    - method
+    - headers - i.e., the sum of the lengths of all keys and values 
+    - body
+    - transform - i.e., the sum of the transform method name length and the length of the transform context
+    
+    `max_res_bytes` is the maximum response length the caller wishes to accept.
```
mraszyk, line 2062:
```suggestion
    `max_res_bytes` is the maximum response length the caller wishes to accept or the default value of `...` if no maximum response length is provided in the actual request. 
```

docs/references/ic-interface-spec.md
```diff
@@ -2027,6 +2035,48 @@ When executing a query or composite query method via a query call (i.e. in non-r
...
+
+-   `ic0.cost_ecdsa(src : I, size : I, dst : I) -> (); `I ∈ {i32, i64}`
+
+-   `ic0.cost_schnorr(src : I, size : I, dst : I) -> (); `I ∈ {i32, i64}`
+
+-   `ic0.cost_vetkey(src : I, size : I, dst : I) -> (); `I ∈ {i32, i64}`
+
+    These system calls accept a key name via `src` + `size` for the specific signing scheme / key. See [`sign_with_ecdsa`](#ic-sign_with_ecdsa), [`sign_with_schnorr`](#ic-sign_with_schnorr) and [`vetkd_encrypted_key`](#ic-vetkd_encrypted_key) for more information. 
+
+    These system calls trap if `src` + `size` do not decode to a valid key name.
```
mraszyk, line 2072:
```suggestion
    These system calls trap if the textual representation does not decode to a valid key name. 
```

docs/references/ic-interface-spec.md
```diff
@@ -2027,6 +2035,48 @@ When executing a query or composite query method via a query call (i.e. in non-r
...
+
+    These system calls accept a key name via `src` + `size` for the specific signing scheme / key. See [`sign_with_ecdsa`](#ic-sign_with_ecdsa), [`sign_with_schnorr`](#ic-sign_with_schnorr) and [`vetkd_encrypted_key`](#ic-vetkd_encrypted_key) for more information. 
+
+    These system calls trap if `src` + `size` do not decode to a valid key name. 
+
+-   `ic0.replication_factor : (src : I, size : I) -> i32; `I ∈ {i32, i64}`
+
+    Returns the replication factor (subnet size) of the subnet identified by the `Principal` at `src` + `size`. 
+
+    This system call traps if `src` + `size` do not decode to a valid `Principal` or the given `Principal` is not a subnet.
```
mraszyk, line 2078: "decode" sounds like the argument is the textual representation of a principal while I'd assume the binary representation to be used

mraszyk (commented):
Could you please also add placeholders into the formal [semantics](https://internetcomputer.org/docs/current/references/ic-interface-spec#system-imports): the actual return value can be abstracted over as `arbitrary()`, but the rest should follow, e.g., `ic0.canister_cycles_balance128`.

docs/references/ic-interface-spec.md
mraszyk: or do we use a different representation which is not textual?

mraszyk: an example might be useful to illustrate this

docs/references/ic-interface-spec.md
michael-weigelt: Do you mean that something like C' = C * 13/34 might round down so the true cost will be slightly higher than calculated? Sure, the caller should be careful about rounding or add a safety margin. But are you suggesting to do anything about that in the system API?

docs/references/ic-interface-spec.md
michael-weigelt: In this syscall, the max_res_bytes is not optional, so there is no default value here. If the user follows the documentation of http_request, they will get the full picture. But I think here this formulation is misleading. 

In other words: The function signature tells the caller that they MUST provide this data explicitly.

michael-weigelt: I added a sentence to clarify, let me know what you think.

docs/references/ic-interface-spec.md
michael-weigelt: I used encode because the user provides bytes, whereas rust demands a valid utf-8 string. But I am fine changing the language if you think decode would imply more.

docs/references/ic-interface-spec.md
mraszyk: I suggest to either not go into details and not comment on conversions here or point out the rounding issue in the spec. The system API itself can stay as is.

docs/references/ic-interface-spec.md
mraszyk: My comment on "encode" does not apply here.

mraszyk: an example would still be useful

docs/references/ic-interface-spec.md
mraszyk: > In this syscall, the max_res_bytes is not optional, so there is no default value here. 

There is a default value in the replica implementation that the caller has to supply here because system API doesn't allow easily modelling optional values.

mraszyk: > I added a sentence to clarify, let me know what you think.

I don't see it. Could you please share a link?

michael-weigelt: pushed now

michael-weigelt: Another suggestion: We can specify that the passed value is clamped to the true replica's maximum. So if the user does not want to bound the max_response_bytes in the call to the management canister, they can supply u64::MAX here and we clamp it to the true 2_000_000 (or whatever this evolves to). 

This would save the caller from having to know the true maximum when they only want to call mgmt with None.

docs/references/ic-interface-spec.md
michael-weigelt: pushed

docs/references/ic-interface-spec.md
```diff
@@ -7465,6 +7515,52 @@ ic0.in_replicated_execution<es>() : i32 =
...
+    Trap {cycles_used = es.cycles_used;}
+  copy_cycles_to_canister<es>(dst, arbitrary())
+
+I ∈ {i32, i64}
+ic0.replication_factor<es, S>(src: I, size: I, dst: I) : i32 = 
+  principal_bytes = copy_from_canister<es>(src, size)
+  if not principal_bytes encode a principal then
+    Trap {cycles_used = es.cycles_used;}
+  return S.canister_subnet[principal_bytes].subnet_size
+
```
michael-weigelt, line 7563: @mraszyk not sure how to refer to `S` from here. It does not seem to be contained in `es`. 
(Also, it is confusing that S above always uses `CanisterId` where I would expect `Principal`)

michael-weigelt: Code was updated since, but the question remains.

mraszyk: the subnet size is available is es.params.sysenv.subnet_size

docs/references/ic-interface-spec.md
mraszyk: I think it's a reasonable suggestion, but I'll bring this up in the spec meeting to gather more feedback.

docs/references/ic-interface-spec.md
michael-weigelt: Of the subnet the canister is on, right? That's how I interpreted `es`. 
But I need the subnet size of a subnet specified by principal_bytes.

mraszyk: right, then let's just use arbitrary() to leave it an implementation detail

docs/references/ic-interface-spec.md
```diff
@@ -2027,6 +2035,48 @@ When executing a query or composite query method via a query call (i.e. in non-r
...
+    - transform - i.e., the sum of the transform method name length and the length of the transform context
+    
+    `max_res_bytes` is the maximum response length the caller wishes to accept. Note that this argument is not optional like in the call to the management canister. The cost depends on `max_res_bytes`, so the caller must provide it explicitly. See the [`http_request`](#ic-http_request) call to the management canister API to learn about the current default and maximum values. 
+
+-   `ic0.cost_ecdsa(src : I, size : I, dst : I) -> ()`; `I ∈ {i32, i64}`
+
+-   `ic0.cost_schnorr(src : I, size : I, dst : I) -> ()`; `I ∈ {i32, i64}`
+
+-   `ic0.cost_vetkey(src : I, size : I, dst : I) -> ()`; `I ∈ {i32, i64}`
+
```
michael-weigelt, line 2069: TODO: Turns out, the (signing scheme, key_name)-tuple is not enough to determine the right key, because at least Schnorr has several curves in use, and they can have the same name.

oggy-dfin:
Here's an idea for a different approach:
1. We introduce just two new API functions, `ic0.cost_new`, and `ic0.cost_perform`, that correspond to `call_new` and `call_perform`; the idea is that we simulate executing the call
2. `cost_new` takes the same arguments as `call_new`
3. Between `cost_new` and `cost_perform`, we can either allow `ic0.call_data_append` or a new `ic0.cost_data_append` function, whatever is simpler
4. `cost_perform` would return two values, the cost in cycles, and the number of cycles that should be attached to the call. We could return this as two pairs of i64 values, or by specifying the destination on the heap, whatever is deemed simpler.

The advantages as I see them:
1. Simpler implementation for the CDKs and/or users (depending on how we choose to expose this in the CDKs). They don't need to understand or do anything special for the different call types.
2. No API additions needed in the future (unlike the current proposal where we'd have to add new System API calls whenever we add new functionality to the management canister) 
3. The cost computation is as precise as possible

The only downside that I see is that the cost of the `cost_*` functions themselves can't be determined in advance, but I think that will be the case whenever we need to parse the payload.

michael-weigelt:
@oggy-dfin thanks for the input. I like that this approach leads to a smaller API surface and covers current and future use cases. 

The problem I see is with parsing the payload, as you have pointed out yourself: This is cheap and constant effort for the `ecdsa`, `schnorr` and `vetkey` use cases. But for `http_request`, we have to get `max_response_bytes` out of the payload, which means copying and deserializing a user-provided blob of up to 2MB. So the cost of `cost_*` scales with the payload, and might be significant (it takes ~900M instructions to deserialize 1.9MB). 
It also means that we have to depend on candid and the mgmt types in the system API impl. 

So we would buy extensibility with 1) more complexity in the impl and 2) more costs to the caller. But is extensibility really worth that? I would consider adding a new system API endpoint of an existing type is a small effort already. Although granted, it is less pleasant than an API that just covers it by design. 

Another downside of simulating a call is that you have to have your payload ready to know the cost. If you have access to the proposed API, where you can just provide some payload lengths, you can make decisions based on costs before deciding on the details of a call. Not sure how interesting this is in reality, but it would suck if people had to compose some dummy data just to predict the cost of a similar call.

mraszyk:
> ~900M instructions to deserialize 1.9MB

is this benchmark using serde_bytes for decoding Vec<u8>?

> It also means that we have to depend on candid and the mgmt types in the system API impl.

a dependency on candid already exists in the system-api crate for the sake of routing

> more complexity in the impl

this complexity can't be avoided, it can only be pushed to the CDKs

michael-weigelt:
>is this benchmark using serde_bytes for decoding Vec?

I measured the instructions on a canister that turns a Vec<u8> into a struct using candid::Decode!(). Whether candid uses serde_bytes itself, idk. 

>this complexity can't be avoided, it can only be pushed to the CDKs

Yes. But there, the precise call type is already known without inspecting a blob.

mraszyk:
> I measured the instructions on a canister that turns a Vec into a struct using candid::Decode!(). Whether candid uses serde_bytes itself, idk.

serde_bytes are used if the type you want to decode contains corresponding serde_bytes annotation. Could you please share the type definition you're decoding to?

docs/references/ic-interface-spec.md
```diff
@@ -1487,8 +1487,16 @@ defaulting to `I = i32` if the canister declares no memory.
...
-    ic0.in_replicated_execution: () -> (result: i32);                                     // * s
+    ic0.is_controller : (src : I, size : I) -> ( result : i32);                           // * s
+    ic0.in_replicated_execution : () -> (result : i32);                                   // * s
+    
+    ic0.cost_call : (method_name_size: i64, payload_size : i64, dst : I) -> ();           // * s
+    ic0.cost_create_canister : (dst : I) -> ();                                           // * s
+    ic0.cost_http_request(request_size : i64, max_res_bytes : i64, dst : I) -> ();        // * s
+    ic0.cost_sign_with_ecdsa(src : I, size : I, ecdsa_curve: i32, dst : I) -> ();         // * s
+    ic0.cost_sign_with_schnorr(src : I, size : I, algorithm: i32, dst : I) -> ();         // * s
+    ic0.cost_vetkd_derive_encrypted_key(src : I, size : I, vetkd_curve: i32, dst : I) -> ();  // * s
```
michael-weigelt, line 1498:
```suggestion
    ic0.cost_http_request : (request_size : i64, max_res_bytes : i64, dst : I) -> ();     // * s
    ic0.cost_sign_with_ecdsa : (src : I, size : I, ecdsa_curve: i32, dst : I) -> ();      // * s
    ic0.cost_sign_with_schnorr : (src : I, size : I, algorithm: i32, dst : I) -> ();      // * s
    ic0.cost_vetkd_derive_encrypted_key : (src : I, size : I, vetkd_curve: i32, dst : I) -> ();  // * s
```

michael-weigelt:
While I agree with your analysis, I'd make a different trade-off. Having an API that is agnostic to the details has these disadvantages: 
- The system API impl gets a blob -> the logic of understanding mgmt canister payloads is now in the system API impl too. Note that if we push this to the caller, i.e., the CDK, it does not add more coupling, because the CDK already needs to understand these payloads anyway. But having a set of switch statements that fans out all the possible blob types in the system API impl is not pretty, especially considering the caller had the precise information available in the first place. 
- The difference in cost of passing a blob versus passing a few u64s is significant, both in terms of cycles and actual work performed by the replica. 
- Asking the user to provide a well-formed payload when they potentially just want to make a decision based on some sizes and lengths is not that ergonomic.

docs/references/ic-interface-spec.md
```diff
@@ -1487,8 +1487,16 @@ defaulting to `I = i32` if the canister declares no memory.
...
+    ic0.is_controller : (src : I, size : I) -> ( result : i32);                           // * s
+    ic0.in_replicated_execution : () -> (result : i32);                                   // * s
+    
+    ic0.cost_call : (method_name_size: i64, payload_size : i64, dst : I) -> ();           // * s
+    ic0.cost_create_canister : (dst : I) -> ();                                           // * s
+    ic0.cost_http_request : (request_size : i64, max_res_bytes : i64, dst : I) -> ();     // * s
+    ic0.cost_sign_with_ecdsa : (src : I, size : I, ecdsa_curve: i32, dst : I) -> ();      // * s
+    ic0.cost_sign_with_schnorr : (src : I, size : I, algorithm: i32, dst : I) -> ();      // * s
+    ic0.cost_vetkd_derive_encrypted_key : (src : I, size : I, vetkd_curve: i32, dst : I) -> ();  // * s
+    ic0.replication_factor : (src : I, size : I) -> i32;                                  // * s
```
michael-weigelt, line 1499: `is_controller`, `sign_with_*` and `vetkd_..` can fail due to 1) malformed inputs and 2) invalid destinations. While these endpoints should trap on 1, they should not trap on 2. Instead, I propose to specify a special return type as follows: 
```
0: call failed
n: actual return value n
```
The caller then has to check if the result was 0 and pass the error to the user. I suggest the corresponding `RejectCode` should be 3, `DestinationInvalid`. 
0 as a guard value is fine because it cannot be a replication factor and it presumably cannot be an actual cost.

docs/references/ic-interface-spec.md
```diff
@@ -2027,6 +2035,53 @@ When executing a query or composite query method via a query call (i.e. in non-r
...
+
+-   `ic0.cost_create_canister : (dst : I) -> ()`; `I ∈ {i32, i64}`
+
+    The cost of creating a canister on the same subnet as the calling canister via [`create_canister`](#ic-create_canister). Note that canister creation via a call to the CMC can have a different cost if the target subnet has a different replication factor. In order to facilitate conversions, see `ic0.replication_factor`. When converting costs using ratios of subnet sizes, be mindful of rounding issues and consider adding a safety margin. 
+
+-   `ic0.cost_http_request(request_size : i64, max_res_bytes : i64, dst : I) -> ()`; `I ∈ {i32, i64}`
+
+    The cost of a canister http outcall via [`http_request`](#ic-http_request). `request_size` is the sum of the byte lengths of the following components of an http request: 
+    - url
+    - method
```
mraszyk, line 2057:
```suggestion
```
this is not part of the formula and it'd be unclear anyway how to compute the byte length of the enumeration variant

docs/references/ic-interface-spec.md
```diff
@@ -2027,6 +2035,53 @@ When executing a query or composite query method via a query call (i.e. in non-r
...
+    - headers - i.e., the sum of the lengths of all keys and values 
+    - body
+    - transform - i.e., the sum of the transform method name length and the length of the transform context
+    
+    `max_res_bytes` is the maximum response length the caller wishes to accept. Note that this argument is not optional like in the call to the management canister. The cost depends on `max_res_bytes`, so the caller must provide it explicitly. See the [`http_request`](#ic-http_request) call to the management canister API to learn about the current default and maximum values.
```
mraszyk, line 2062: Given that the spec actually contains the default value, let's trap if max_res_bytes exceeds that value (2,000,000).

docs/references/ic-interface-spec.md
```diff
@@ -2027,6 +2035,53 @@ When executing a query or composite query method via a query call (i.e. in non-r
...
+    - body
+    - transform - i.e., the sum of the transform method name length and the length of the transform context
+    
+    `max_res_bytes` is the maximum response length the caller wishes to accept. Note that this argument is not optional like in the call to the management canister. The cost depends on `max_res_bytes`, so the caller must provide it explicitly. See the [`http_request`](#ic-http_request) call to the management canister API to learn about the current default and maximum values. 
+
+-   `ic0.cost_sign_with_ecdsa(src : I, size : I, ecdsa_curve: i32, dst : I) -> ()`; `I ∈ {i32, i64}`
+
+-   `ic0.cost_sign_with_schnorr(src : I, size : I, algorithm: i32, dst : I) -> ()`; `I ∈ {i32, i64}`
+
+-   `ic0.cost_vetkd_derive_encrypted_key(src : I, vetkd_curve: i32, size : I, dst : I) -> ()`; `I ∈ {i32, i64}`
```
mraszyk, line 2068: Shouldn't src and size be next to each other?

docs/references/ic-interface-spec.md
mraszyk: Rephrased the suggestion to
```
(the caller should provide the default value of `2,000,000` if no maximum response length is provided in the actual request).
```
(as discussed separately)

docs/references/ic-interface-spec.md
michael-weigelt: Never mind. It's better not to introduce special return values. 
Removing replication_factor from this PR will remove one case. 
Discussion about the key names is ongoing.

docs/references/ic-interface-spec.md
```diff
@@ -2083,13 +2082,10 @@ These system calls return costs in Cycles, represented by 128 bits, which will b
...
     See [`sign_with_ecdsa`](#ic-sign_with_ecdsa), [`sign_with_schnorr`](#ic-sign_with_schnorr) and [`vetkd_encrypted_key`](#ic-vetkd_encrypted_key) for more information.
 
-    These system calls trap if the string represented by `src` + `size` does not correspond to a valid key name, such as `dfx_test_key`, `test_key_1` or `key_1`, or if the provided curve enum is an unspecified variant. 
-
--   `ic0.replication_factor : (src : I, size : I) -> i32`; `I ∈ {i32, i64}`
-
-    Returns the replication factor (subnet size) of the subnet identified by the `Principal` at `src` + `size`. 
-
-    This system call traps if `src` + `size` do not represent a valid `Principal` or the given `Principal` is not a subnet.   
+    These system calls trap if `src` + `size` exceeds the size of the WebAssembly memory. Otherwise, they return an `i32` with the following meaning:
```
mraszyk, line 2085:
```suggestion
    These system calls trap if `src` + `size`  or `dst` + 8 exceed the size of the WebAssembly memory. Otherwise, they return an `i32` with the following meaning:
```

docs/references/ic-interface-spec.md
```diff
@@ -2083,13 +2082,10 @@ These system calls return costs in Cycles, represented by 128 bits, which will b
...
+    - `0 (00)`: Success. The result can be found at the memory address `dst`.
+    - `1 (01)`: Invalid curve or algorithm. Memory at `dst` is left unchanged. 
+    - `2 (10)`: Invalid key name for the given scheme/curve. Memory at `dst` is left unchanged.
```
mraszyk, line 2088: What if both key and curve are invalid? Why don't we return 3 (11).

docs/references/ic-interface-spec.md
```diff
@@ -7558,40 +7554,34 @@ ic0.cost_cost_http_request<es>(request_size: i64, max_res_bytes: i64, dst: I) :
...
 I ∈ {i32, i64}
-ic0.cost_sign_with_ecdsa<es>(src: I, size: I, ecdsa_curve: i32, dst: I) : () = 
-  principal_bytes = copy_from_canister<es>(src, size)
-  if not principal_bytes encode a principal then
-    Trap {cycles_used = es.cycles_used;}
-  if principal_bytes ∉ es.subnets then
-    Trap {cycles_used = es.cycles_used;}
+ic0.cost_sign_with_ecdsa<es>(src: I, size: I, ecdsa_curve: i32, dst: I) : i32 = 
+  key_name = copy_from_canister<es>(src, size)
+  if ecdsa_curve ∉ known_curves then
```
mraszyk, line 7559: maybe say that `known_curves = arbitrary()` and `known_keys = arbitrary()` so that these two identifiers are well-defined; also below

docs/references/ic-interface-spec.md
```diff
@@ -1490,8 +1490,15 @@ defaulting to `I = i32` if the canister declares no memory.
...
     ic0.global_timer_set : (timestamp : i64) -> i64;                                      // I G U Ry Rt C T
     ic0.performance_counter : (counter_type : i32) -> (counter : i64);                    // * s
-    ic0.is_controller: (src : I, size : I) -> ( result: i32);                             // * s
-    ic0.in_replicated_execution: () -> (result: i32);                                     // * s
+    ic0.is_controller : (src : I, size : I) -> ( result : i32);                           // * s
+    ic0.in_replicated_execution : () -> (result : i32);                                   // * s
+    
+    ic0.cost_call : (method_name_size: i64, payload_size : i64, dst : I) -> ();           // * s
+    ic0.cost_create_canister : (dst : I) -> ();                                           // * s
+    ic0.cost_http_request : (request_size : i64, max_res_bytes : i64, dst : I) -> ();     // * s
```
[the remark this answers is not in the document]
mraszyk: > Now the user has charged the canister with C cycles needlessly.

Cycles are [refunded](https://github.com/dfinity/ic/blob/665c75015ad5afcd701565580f812a04862384d5/rs/execution_environment/src/execution_environment.rs#L938) in this case.

docs/references/ic-interface-spec.md
michael-weigelt: A key name's validity depends on the signing scheme and curve. So returning 3 is just not a sensible code path.
First we find all keys that have the given curve. Then we check if one of them has the given name. If the first check fails, the second one is irrelevant, and performing the check anyway is useless. 
Imagine you want Ecdsa/C1/"key_2" and the system tells you that it does not know this, but why not use Schnorr instead because Schnorr/C2/"key_2" exists? In other words, "there exists a key with this name, but for a different curve / scheme" is useless, because the user 1) specifies an scheme endpoint and 2) constructs a typed argument with the desired scheme and curve, so they already know what scheme / curve they want.

docs/references/ic-interface-spec.md
```diff
@@ -7490,6 +7541,54 @@ ic0.in_replicated_execution<es>() : i32 =
...
+I ∈ {i32, i64}
+ic0.cost_call<es>(method_name_size: i64, payload_size: i64, dst: I) : () = 
+  copy_cycles_to_canister<es>(dst, arbitrary())
+
+I ∈ {i32, i64}
+ic0.cost_create_canister<es>(dst: I) : () = 
+  copy_cycles_to_canister<es>(dst, arbitrary())
+
+I ∈ {i32, i64}
+ic0.cost_cost_http_request<es>(request_size: i64, max_res_bytes: i64, dst: I) : () =
```
mraszyk, line 7553:
```suggestion
ic0.cost_http_request<es>(request_size: i64, max_res_bytes: i64, dst: I) : () = 
```

docs/references/ic-interface-spec.md
```diff
@@ -2036,6 +2043,50 @@ When executing a query or composite query method via a query call (i.e. in non-r
...
+    - url
+    - headers - i.e., the sum of the lengths of all keys and values 
+    - body
+    - transform - i.e., the sum of the transform method name length and the length of the transform context
+    
+    `max_res_bytes` is the maximum response length the caller wishes to accept (the caller should provide the default value of `2,000,000` if no maximum response length is provided in the actual request to the management canister). 
+
+-   `ic0.cost_sign_with_ecdsa(src : I, size : I, ecdsa_curve: i32, dst : I) -> i32`; `I ∈ {i32, i64}`
+
+-   `ic0.cost_sign_with_schnorr(src : I, size : I, algorithm: i32, dst : I) -> i32`; `I ∈ {i32, i64}`
```
mraszyk, line 2074:
```suggestion
-   `ic0.cost_sign_with_schnorr(src : I, size : I, schnorr_algorithm: i32, dst : I) -> i32`; `I ∈ {i32, i64}`
```

docs/references/ic-interface-spec.md
```diff
@@ -2036,6 +2043,50 @@ When executing a query or composite query method via a query call (i.e. in non-r
...
+
+-   `ic0.cost_vetkd_derive_encrypted_key(src : I, size : I, vetkd_curve: i32, dst : I) -> i32`; `I ∈ {i32, i64}`
+
+    These system calls accept a key name via a textual representation for the specific signing scheme / key of a given size stored in the heap memory starting at offset `src`. They also accept an `i32` with the following interpretations:
+    - `ecdsa_curve: 0 → secp256k1`
+    - `schnorr_algorithm: 0 → bip340secp256k1, 1 → ed25519`
```
mraszyk, line 2080:
```suggestion
    - `algorithm: 0 → bip340secp256k1, 1 → ed25519`
```

docs/references/ic-interface-spec.md
```diff
@@ -2036,6 +2043,50 @@ When executing a query or composite query method via a query call (i.e. in non-r
...
+    - `vetkd_curve: 0 → bls12_381`
+
+    See [`sign_with_ecdsa`](#ic-sign_with_ecdsa), [`sign_with_schnorr`](#ic-sign_with_schnorr) and [`vetkd_encrypted_key`](#ic-vetkd_encrypted_key) for more information.
+
+    These system calls trap if `src` + `size` or `dst` + 16 exceed the size of the WebAssembly memory. Otherwise, they return an `i32` with the following meaning:
+    - `0 (00)`: Success. The result can be found at the memory address `dst`.
+    - `1 (01)`: Invalid curve or algorithm. Memory at `dst` is left unchanged. 
+    - `2 (10)`: Invalid key name for the given scheme/curve. Memory at `dst` is left unchanged.
```
mraszyk, line 2088:
```suggestion
    - `2 (10)`: Invalid key name for the given curve/algorithm. Memory at `dst` is left unchanged. 
```

docs/references/ic-interface-spec.md
```diff
@@ -7490,6 +7541,54 @@ ic0.in_replicated_execution<es>() : i32 =
...
+    return 1
+  if key_name ∉ known_keys then
+    return 2
+  copy_cycles_to_canister<es>(dst, arbitrary())
+  return 0
+
+I ∈ {i32, i64}
+ic0.cost_sign_with_schnorr<es>(src: I, size: I, algorithm: i32, dst: I) : i32 = 
+  known_keys = arbitrary()
+  known_curves = arbitrary()
```
mraszyk, line 7571:
```suggestion
  known_algorithms = arbitrary()
```

docs/references/ic-interface-spec.md
michael-weigelt: pushed something more precise

docs/references/ic-interface-spec.md
```diff
@@ -2036,6 +2043,50 @@ When executing a query or composite query method via a query call (i.e. in non-r
 
     This traps if `ic0.data_certificate_present()` returns `0`.
 
+### Cycle cost calculation {#system-api-cycle-cost}
+
+Inter-canister calls have an implicit cost, and some calls to the management canister require the caller to attach cycles to the call explicitly.  
+The various cost factors may change over time, so the following system calls give the canister programmatic, up-to-date information about the costs.
+
+These system calls return costs in Cycles, represented by 128 bits, which will be written to the heap memory starting at offset `dst`.
```
mraszyk, line 2051: Remark that the results are only correct if the input values are within the size bounds for the corresponding calls.

michael-weigelt: left a comment

docs/references/ic-interface-spec.md
```diff
@@ -2048,7 +2048,7 @@ When executing a query or composite query method via a query call (i.e. in non-r
 Inter-canister calls have an implicit cost, and some calls to the management canister require the caller to attach cycles to the call explicitly.  
 The various cost factors may change over time, so the following system calls give the canister programmatic, up-to-date information about the costs.
 
-These system calls return costs in Cycles, represented by 128 bits, which will be written to the heap memory starting at offset `dst`. 
+These system calls return costs in Cycles, represented by 128 bits, which will be written to the heap memory starting at offset `dst`. Note that the cost calculation is only correct for correct inputs, e.g., a method name length argument should not exceed 10'000, because such an argument would be rejected by ic0.call_perform. The cost API will still return a number in this case, but it would not have a real meaning.
```
mraszyk, line 2051: Where did you get the threshold of 10,000 from (the spec [mentions](https://internetcomputer.org/docs/current/references/ic-interface-spec#system-api-module) 20,000)? Also the [validation](https://github.com/dfinity/ic/blob/0d9bdcb16515d8b8194d9bcce21e5caac8cca666/rs/system_api/src/request_in_prep.rs#L87) of method name length happens already in ic0.call_new.

docs/references/ic-interface-spec.md
mraszyk: Thank you! I've opened a follow-up conversation.

docs/references/ic-interface-spec.md
michael-weigelt: fixed

Dfinity-Bjoern (approved):

docs/references/ic-interface-spec.md
```diff
@@ -2036,6 +2043,50 @@ When executing a query or composite query method via a query call (i.e. in non-r
...
+    - `ecdsa_curve: 0 → secp256k1`
+    - `algorithm: 0 → bip340secp256k1, 1 → ed25519`
+    - `vetkd_curve: 0 → bls12_381`
+
+    See [`sign_with_ecdsa`](#ic-sign_with_ecdsa), [`sign_with_schnorr`](#ic-sign_with_schnorr) and [`vetkd_encrypted_key`](#ic-vetkd_encrypted_key) for more information.
+
+    These system calls trap if `src` + `size` or `dst` + 16 exceed the size of the WebAssembly memory. Otherwise, they return an `i32` with the following meaning:
+    - `0 (00)`: Success. The result can be found at the memory address `dst`.
+    - `1 (01)`: Invalid curve or algorithm. Memory at `dst` is left unchanged. 
+    - `2 (10)`: Invalid key name for the given combination of signing scheme and curve/algorithm. Memory at `dst` is left unchanged.
```
lwshang, line 2088: I noticed the return codes in the spec are formatted as `0 (00)` and `1 (01)`, which reminded me of the `bitflags` approach we discussed in the spec meeting. If we’re sticking with that, `3 (11)` would mean both "invalid curve/algorithm" and "invalid key name."

However, I checked the current implementation, and it doesn’t follow the `bitflags` approach. For example, when both the curve and key name are invalid in `ic0.cost_sign_with_ecdsa`, it returns `1 (01)` — meaning "invalid curve/algorithm" takes precedence.

To clarify this section, we should either:
- **Adopt** `bitflags`: Update the implementation to match the intention from the meeting and specify it here.
- **Drop** `bitflags`: Remove the bits representation in parentheses and explicitly state which code is returned when both fields are invalid.

mraszyk: I've removed the bitflags and emphasized that the return value 2 is only returned if the curve/algorithm is valid.

michael-weigelt: And [here's](https://github.com/dfinity/portal/pull/4110#discussion_r1959433477) the reasoning

oggy-dfin (approved):

PR #4753, opened 2025-02-04
replace thumbnails in video slider on library page with youtube high-…

herrpatrickmueller:
replace thumbnails in video slider on library page with youtube high-res images


artkorotkikh-dfinity (approved):
LGTM

PR #5080, opened 2025-02-11
Clarify ingress expiry semantics

yvonneanne:

  side navigation bar.

### DIFF
--- a/docs/references/ic-interface-spec.md
+++ b/docs/references/ic-interface-spec.md
@@ -588,7 +588,7 @@ The Internet Computer has two HTTPS APIs for canister calling:
 - [*Asynchronous*](#http-async-call-overview) canister calling, where the user must poll the Internet Computer for the status of the canister call by _separate_ HTTPS requests.
 - [*Synchronous*](#http-sync-call-overview) canister calling, where the status of the canister call is in the response of the original HTTPS request.
 
-#### Asynchronous canister calling {#http-async-call-overview}
+#### Canister call state transition {#http-async-call-overview}
 
 1.  A user submits a call via the [HTTPS Interface](#http-interface). No useful information is returned in the immediate response (as such information cannot be trustworthy anyways).
 
@@ -663,7 +663,7 @@ In order to call a canister, the user makes a POST request to `/api/v3/canister/
 
 -   `request_type` (`text`): Always `call`
 
--   `sender`, `nonce`, `ingress_expiry`: See [Authentication](#authentication)
+-   `sender`, `nonce`, `ingress_expiry`: See [Authentication](#authentication) and [. 
 
 -   `canister_id` (`blob`): The principal of the canister to call.
 
@@ -6004,9 +6004,9 @@ S with
 
 NB: The refunded cycles, `RM.refunded_cycles` are, by construction, empty.
 
-#### Request clean up
+#### Update call request clean up
 
-The IC will keep the data for a completed or rejected request around for a certain, implementation defined amount of time, to allow users to poll for the data. After that time, the data of the request will be dropped:
+The IC will keep the data for a completed or rejected update <code>call</code> request around for a certain, implementation defined amount of time, to allow users to poll for the data with <code>read_state<code> requests . After that time, the data of the request will be dropped:
 
 Conditions  
 


docs/references/ic-interface-spec.md
```diff
@@ -642,16 +642,20 @@ The characteristic property of the `processing` state is that *the initial effec
 A call may be rejected by the IC or the canister. In either case, there is no guarantee about how much processing of the call has happened.
 
 To avoid replay attacks, the transition from `done` or `received` to `pruned` must happen no earlier than the call's `ingress_expiry` field.
+If a subnet's time is greater than a call's `ingress_expiry` field and it is still unknown to the IC (i.e, it was never in state `received`, `processing`, `replied`, `rejected`, or `done`), then the call will never be in one of these states.
 
 Calls must stay in `replied` or `rejected` long enough for polling users to catch the response.
```
yvonneanne, line 647:
```suggestion
Calls stay in `replied` or `rejected` long enough for polling users to catch the response under good networking conditions with low load.
```

docs/references/ic-interface-spec.md
```diff
@@ -642,16 +642,20 @@ The characteristic property of the `processing` state is that *the initial effec
...
+A synchronous update call, also known as a "call and await", is a type of update call where the replica will attempt to respond to the HTTPS request with a certificate of the call status. 
+On the replica, a synchronous call request goes through the same states (`received`, `processing`, `replied`, `rejected`, or `done`) as the ones depicted for the [asynchronous API](#http-async-call-overview).
+If the returned certificate indicates that the update call is in a terminal state (`replied`, `rejected`, or `done`), then the user __does not need to poll__ (using [`read_state`](#http-read-state) requests) 
+to determine the result of the call. A terminal state means the call has completed its execution.
 
-The synchronous call endpoint is useful for users as it removes the networking overhead of polling the IC to determine the status of their call.
+The synchronous call endpoint is useful for users as it removes the networking overhead of polling the IC to determine the status of their call if a terminal state is reached quickly.
```
THLO, line 658: I'm not sure how to best improve it but "reached quickly" is vague, right?

docs/references/ic-interface-spec.md
```diff
@@ -596,7 +596,7 @@ The Internet Computer has two HTTPS APIs for canister calling:
 
 3.  The IC asks the targeted canister if it is willing to accept this message and be charged for the expense of processing it. This uses the [Ingress message inspection](#system-api-inspect-message) API for normal calls. For calls to the management canister, the rules in [The IC management canister](#ic-management-canister) apply.
 
-4.  At some point, the IC may accept the call for processing and set its status to `received`. This indicates that the IC as a whole has received the call and plans on processing it (although it may still not get processed if the IC is under high load). Furthermore, the user should also be able to ask any endpoint about the status of the pending call.
+4.  At some point, the IC may accept the call for processing and set its status to `received`. This indicates that the IC as a whole has received the call and plans on processing it (although it may still not get processed if the IC is under high load). At the time of this transition, the call has not yet expired. Furthermore, the user should also be able to ask any endpoint about the status of the pending call.
```
lmuntaner, line 599: I find point 8 a bit confusing as well:

> In the case that the call has been retained for long enough, but the request has not expired yet

Can we be specific on "long enough" and "has not yet expired"?

Stupid question: does this relate to the "ingress_expiriry" field?

docs/references/ic-interface-spec.md
```diff
@@ -642,16 +642,20 @@ The characteristic property of the `processing` state is that *the initial effec
 A call may be rejected by the IC or the canister. In either case, there is no guarantee about how much processing of the call has happened.
 
 To avoid replay attacks, the transition from `done` or `received` to `pruned` must happen no earlier than the call's `ingress_expiry` field.
+If a subnet's time is greater than a call's `ingress_expiry` field and it is still unknown to the IC (i.e, it was never in state `received`, `processing`, `replied`, `rejected`, or `done`), then the call will never be in one of these states.
 
-Calls must stay in `replied` or `rejected` long enough for polling users to catch the response.
+Calls stay in `replied` or `rejected` long enough for polling users to catch the response under good networking conditions with low load.
```
lmuntaner, line 647: Can we also be specific of what it means "long enough"?

docs/references/ic-interface-spec.md
yvonneanne: Yes, this relates to the ingress_expiry field. 
I added  a few sentences about subnet time and call expiry above. Does this make it more clear what "has not expired" means here? If not, can you make a suggestion?

(let's discuss "long enough" in L646/647)

docs/references/ic-interface-spec.md
yvonneanne: iirc, this is vague on purpose (I asked the same question some time ago and I don't remember the answer). Can someone more familiar with the interface spec history comment on this?

docs/references/ic-interface-spec.md
THLO: I like the alternative better! Thanks!

docs/references/ic-interface-spec.md
```diff
@@ -751,7 +758,7 @@ In order to read parts of the [The system state tree](#state-tree), the user mak
 
 -   `request_type` (`text`): Always `read_state`
 
--   `sender`, `nonce`, `ingress_expiry`: See [Authentication](#authentication)
+-   `sender`, `nonce`, `ingress_expiry`: See [Authentication](#authentication). `ingress_expiry` refers to this request's expiry, not the expiry of an earlier call request.
```
mraszyk, line 761: This request might not refer to any earlier call request in which case the note can be confusing.

docs/references/ic-interface-spec.md
mraszyk: It should be [five minutes](https://github.com/dfinity/ic/blob/3bc41ef43481abd0f82a684229745913d74927f8/rs/limits/src/lib.rs#L17). Is there a reason to keep this private, @derlerd-dfinity ?

derlerd-dfinity: I don't remember details of this discussion, but I could imagine the following reasons why it was decided to remain vague:
1. Not specifying how long we keep things around would allow to easily change it in the future.
2. Under high load (i.e., when hitting the memory limits of the ingress history) both `rejected` and `replied` statuses might immediately be transitioned to `done`. So even if we specified 5 minutes here I can't see how it could be useful to an application, as the 5 minutes are just a good weather property rather than a guarantee. Applications should make sure to not rely on being able to observe `replied` or `rejected` anyways.

I think (2) is the more important reason.

I don't feel strongly about mentioning / not mentioning the 5 minutes, as long as it is clear that even if the application manages to get a request through and an answer back within 5 minutes it might miss the `rejected`/`replied` status nevertheless and instead only see the `done` status.

mraszyk: > Under high load (i.e., when hitting the memory limits of the ingress history) both rejected and replied statuses might immediately be transitioned to done. So even if we specified 5 minutes here I can't see how it could be useful to an application, as the 5 minutes are just a good weather property rather than a guarantee.

What's the motivation behind distinguishing "done" and "pruned" then?

mraszyk: > By "pruned" you mean it is no longer in the tree?

Indeed.

> If so, the motivation is that done tells you that the message was processed and it might have been a rejected or a replied. For a status that is no longer in the tree (i.e., was "pruned") you know nothing, i.e., not even if it was processed.

If the message times out [here](https://github.com/dfinity/ic/blob/6e327863fd0e72d8cf9c5c46fc1263f548fad4f5/rs/execution_environment/src/scheduler.rs#L837), then it can end up in the "done" status although it was never processed. So the "done" status only seems to make a difference if the subnet time is less than the ingress expiry.

docs/references/ic-interface-spec.md
```diff
@@ -642,16 +645,20 @@ The characteristic property of the `processing` state is that *the initial effec
 A call may be rejected by the IC or the canister. In either case, there is no guarantee about how much processing of the call has happened.
 
 To avoid replay attacks, the transition from `done` or `received` to `pruned` must happen no earlier than the call's `ingress_expiry` field.
+If a subnet's time is greater than a call's `ingress_expiry` field and it is still unknown to the IC (i.e, it was never in state `received`, `processing`, `replied`, `rejected`, or `done`), then the call will never be in one of these states.
```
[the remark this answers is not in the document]
mraszyk: CC @derlerd-dfinity

docs/references/ic-interface-spec.md
derlerd-dfinity: > So the "done" status only seems to make a difference if the subnet time is less than the ingress expiry.

Yeah, good point.

docs/references/ic-interface-spec.md
```diff
@@ -596,7 +599,7 @@ The Internet Computer has two HTTPS APIs for canister calling:
 
 3.  The IC asks the targeted canister if it is willing to accept this message and be charged for the expense of processing it. This uses the [Ingress message inspection](#system-api-inspect-message) API for normal calls. For calls to the management canister, the rules in [The IC management canister](#ic-management-canister) apply.
 
-4.  At some point, the IC may accept the call for processing and set its status to `received`. This indicates that the IC as a whole has received the call and plans on processing it (although it may still not get processed if the IC is under high load). Furthermore, the user should also be able to ask any endpoint about the status of the pending call.
+4.  At some point, the IC may accept the call for processing and set its status to `received`. This indicates that the IC as a whole has received the call and plans on processing it (although it may still not get processed if the IC is under high load). At the time of this transition, the call has not yet expired. Furthermore, the user should also be able to ask any endpoint about the status of the pending call.
```
oggy-dfin, line 602:
```suggestion
4.  At some point, the IC may accept the call for processing and set its status to `received`. This indicates that the IC as a whole has received the call and plans on processing it (although it may still not get processed if the IC is under high load). This transition can only happen before the target canister's time (as visible in the [state tree](#state-tree-time)) exceeds the [`ingress_time`](#http-call) field of the request which submitted the call. Furthermore, the user should also be able to ask any endpoint about the status of the pending call.
```

docs/references/ic-interface-spec.md
```diff
@@ -588,6 +588,9 @@ The Internet Computer has two HTTPS APIs for canister calling:
 - [*Asynchronous*](#http-async-call-overview) canister calling, where the user must poll the Internet Computer for the status of the canister call by _separate_ HTTPS requests.
 - [*Synchronous*](#http-sync-call-overview) canister calling, where the status of the canister call is in the response of the original HTTPS request.
 
+The publicly exposed state of a subnet (including the status of calls to its canisters) evolves in rounds. Each round is associated with a discrete timestamp, see [Time](#state-tree-time).
```
oggy-dfin, line 591: I'd propose to delete this discussion here and move it to where `ingress_expiry` is used, I'll make a suggestion below.

docs/references/ic-interface-spec.md
oggy-dfin: Not introduced by this PR, but what does "the user should be able to ask any endpoint about the status of the pending call" mean here? Doesn't the user have to go to `read_state` endpoint?

docs/references/ic-interface-spec.md
mraszyk: ingress_time => ingress_expiry

oggy-dfin: thanks, fixed

docs/references/ic-interface-spec.md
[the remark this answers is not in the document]
Dfinity-Bjoern: Indeed, that would be the right time window where this check is guaranteed to be correct. Before the subnet hits `ingress_expiry` there's no guarantee of the message not being executed, and 5 minutes after that time there's no guarantee that the data would still be around.

docs/references/ic-interface-spec.md
Dfinity-Bjoern: I think I would suggest that we add a bit more detail here, as per this discussion.
- Make "long enough" more explicit, saying that it's 5 minutes.
- State that it's not guaranteed, but that this doesn't make a major difference since robust application design means that the client side needs to be able to deal with the fact that the network may just be gone for 5 minutes and thus anyway has to handle the case of not being able to fetch the response.

yvonneanne:
Thanks for improving the PR while I was away. It looks like all comments have been addressed, we're missing a review though before being able to merge it.

docs/references/ic-interface-spec.md
```diff
@@ -590,23 +590,21 @@ The Internet Computer has two HTTPS APIs for canister calling:
 
 #### Asynchronous canister calling {#http-async-call-overview}
 
-1.  A user submits a call via the [HTTPS Interface](#http-interface). No useful information is returned in the immediate response (as such information cannot be trustworthy anyways).
+1.  A user submits a call via the [HTTPS Interface](#http-interface). The node receiving the request asks the targeted canister if it is willing to accept this message and be charged for the expense of processing it. For calls to the management canister, the rules in [The IC management canister](#ic-management-canister) apply. Otherwise, the node uses the [Ingress message inspection](#system-api-inspect-message) API.
```
oggy-dfin, line 593: Reading this, I think it needs improvement. Which node (could be the boundary node)? What does it mean to "ask the canister" - it's not clear that this is not a replicated execution. 

Overall, I think what we should get across (but I don't think we're there yet) is that these checks are best-effort, as the node may both (a) throw stuff out even if it passes the checks, (b) let stuff through that shouldn't pass the checks (at least for message inspection, I guess not for the management canister, though I'm not sure which rules are meant here).

How about something like:
> A user submits a call via the [HTTPS Interface](#http-interface), and the call gets forwarded to a replica. That  replica decides whether to accept the call. A honest replica will do this by locally evaluating the [ingress message inspection](#system-api-inspect-message) API for ordinary calls, and <insert the relevant part of the management canister API here>

oggy-dfin: Re-reading the points below, maybe we should merge point 2 here too, I guess this is exactly the time where "the IC behaves as if it does not know about the call"?

mraszyk: Some more refactoring: https://github.com/dfinity/portal/pull/5080/commits/3c6bb378df504f0eeeb373c718bec442a89b865b

yvonneanne: Nice, thanks to both of you!

mraszyk: One more [refactor](https://github.com/dfinity/portal/pull/5080/commits/fce8ec584da8a03177b711595c126e4b2c7cd9f9) so that we can also refer to the formal section of the spec.

docs/references/ic-interface-spec.md
```diff
@@ -590,23 +590,27 @@ The Internet Computer has two HTTPS APIs for canister calling:
...
+1.  A user submits a call via the [HTTPS Interface](#http-interface) and the call is received by a replica (a node belonging to an IC subnet). The receiving replica decides whether it accepts the call. An honest replica does so by checking that the target canister is not frozen and
 
-2.  For a certain amount of time, the IC behaves as if it does not know about the call.
+  - checking that the target canister is not empty and running and performing [ingress message inspection](#system-api-inspect-message) for calls to a regular canister;
 
-3.  The IC asks the targeted canister if it is willing to accept this message and be charged for the expense of processing it. This uses the [Ingress message inspection](#system-api-inspect-message) API for normal calls. For calls to the management canister, the rules in [The IC management canister](#ic-management-canister) apply.
+  - checking that the management canister method can be called via ingress messages and that the caller is a controller of the target canister for calls to the management canister.
```
yvonneanne, line 597:
```suggestion
  - checking that the management canister method can be called via ingress messages and that the caller is a controller of the target canister for calls to the management canister.
  
  The replica's immediate response go the HTTPS call about the acceptance cannot be trusted, because a single replica may deviate from the protocol.
```

@mraszyk In the current version of the PR there is no information anymore about the response the sender gets immediately when submitting a call. I think it is important to mention something about this aspect. What do you think about my proposal to still write a sentence about this?

mraszyk: That statement does not apply if the immediate response (when requesting /api/v3) contains a certificate for the response. There might also be no "immediate" response in case of /api/v3 endpoint. We could try to go more into details and list the cases which cannot be trusted, but I'm not sure if it's worthwhile.

yvonneanne: But this is in the part of the asynchronous call (v2), see to Line 591

mraszyk: Sure, but the synchronous call part also refers to this section since the flow is similar (except that in many cases, the immediate response already contains the trustworthy certified response).

mraszyk: How about: "Note that you can only trust valid IC certificates in IC responses since malicious replicas could spoof their responses."

yvonneanne:
Yep, that fits for both! Great suggestion.

docs/references/ic-interface-spec.md
yvonneanne: That's a great suggestion, thanks Martin!

mraszyk: https://github.com/dfinity/portal/pull/5080/commits/d620aa8aca3f0daf9da59f5cde0ef9c7a50ec2f3

docs/references/ic-interface-spec.md
oggy-dfin: WDYT we add a pointer that an error may be signaled to the user in step 1 (forward pointing to the HTTP 200 etc), but that the error is not trustworthy unless accompanied by a certificate?

mraszyk: Not only an error without a certificate is untrustworthy, also HTTP 202 (request accepted; keep polling) is untrustworthy (although there's no harm in polling).

docs/references/ic-interface-spec.md
```diff
@@ -590,23 +590,29 @@ The Internet Computer has two HTTPS APIs for canister calling:
...
 
-4.  At some point, the IC may accept the call for processing and set its status to `received`. This indicates that the IC as a whole has received the call and plans on processing it (although it may still not get processed if the IC is under high load). Furthermore, the user should also be able to ask any endpoint about the status of the pending call.
+  Note that you can only trust valid IC certificates in IC responses since malicious replicas could spoof their responses.
```
oggy-dfin, line 599:
```suggestion
  Note that any reply from the IC at this point (for example, errors, or replies saying that the message is accepted) isn't yet accompanied by a valid IC certificate and thus cannot be trusted, since it could come from a malicious replica.
```

mraszyk: I'm worried that this can be confusing w.r.t. /api/v3/.../call endpoint whose response could contain a valid IC certificate.

oggy-dfin: The problem with the current statement that I have is that it's disconnected from the rest, there's no mention that there might be a response. So I'd like us to just mention that there might be a response even at this point, but you shouldn't necessarily trust it. Maybe we could also move the notice to a different place.

mraszyk: Find a suitable place is tricky as the /api/v3/.../call endpoint can reply at any stage.

yvonneanne: Maybe we can make this part more explicitly about one of the endpoints and then explain the differences of the other later?

oggy-dfin: I made another attempt and updated the comment, hopefully there's no confusion now.

mraszyk: > Only responses that contain a valid IC certificate should be trusted

=> "Only valid IC certificates should be trusted in such responses" (as attaching a valid IC certificate doesn't make everything else trustworthy).

> other responses come from a single replica that can be either honest or malicious

=> I'd drop "other" (as all responses come from a single replica)

mraszyk (approved):

PR #5489, opened 2025-03-03
feat: Specify a hard lower limit for canister's freezing threshold

dsarlis:
This PR updates the allowed lower bound for a canister's freezing threshold.

The new lower bound is proposed to be 1 week. Lower values for a freezing threshold are considered dangerous as the canister can be quickly drained of cycles and uninstalled without giving enough time to react to canister developers.

This change will make it harder for canister developers to make accidental changes to their canister's freezing threshold and risk their canister being uninstalled too quickly.


docs/references/ic-interface-spec.md
```diff
@@ -2121,7 +2121,9 @@ The optional `settings` parameter can be used to set the following settings:
 
 -   `freezing_threshold` (`nat`)
 
-    Must be a number between 0 and 2<sup>64</sup>-1, inclusively, and indicates a length of time in seconds.
+    Must be a number between 86400 (or equivalent of 1 day in seconds) and 2<sup>64</sup>-1, inclusively, and indicates a length of time in seconds.
```
mraszyk, line 2124: Why a hard lower bound instead of relying on the unsafe settings flag to allow very low freezing threshold? It could be useful to have a freezing threshold of zero, e.g., to move all cycles out of a canister.

dsarlis: There's general agreement internally that a freezing threshold of 0 (or anything below a day) is extremely dangerous and the added value of saving a few more cycles (like in the use case you describe) does not outweigh the risks. In fact, some people wanted to completely disallow anything below 2 weeks (or even 30 days) but that certainly felt way too aggressive.

mraszyk:
How do we plan to deal with existing canisters violating the new conditions?

dsarlis:
> How do we plan to deal with existing canisters violating the new conditions?

Existing canisters would be allowed to maintain their current settings. If they wish to change them though, they would have to conform to the new restrictions.

docs/references/ic-interface-spec.md
mraszyk: Moving cycles out of a canister will anyway be facilitated by the new ic0.canister_liquid_cycle_balance128.

docs/references/ic-interface-spec.md
```diff
@@ -2121,7 +2121,9 @@ The optional `settings` parameter can be used to set the following settings:
...
+
+    Values between 1 day and 2 weeks, or more formally any value in the range [1, 14) days, can only be set if the special `allow-unsafe-settings` is used in the relevant management canister APIs.
```
mraszyk, line 2126:
```suggestion
    Values between 1 day and 2 weeks, or more formally any value that is at least 1 day and less than 14 days, can only be set if the special `allow-unsafe-settings` is used in the relevant management canister APIs.
```

mraszyk: As one could interpret [1, 14) days as [1, 13] days.

docs/references/ic-interface-spec.md
```diff
@@ -2121,7 +2121,9 @@ The optional `settings` parameter can be used to set the following settings:
 
 -   `freezing_threshold` (`nat`)
 
-    Must be a number between 0 and 2<sup>64</sup>-1, inclusively, and indicates a length of time in seconds.
+    Must be a number between 86400 (or equivalent of 1 day in seconds) and 2<sup>64</sup>-1, inclusively, and indicates a length of time in seconds.
+
+    Values between 1 day and 2 weeks, or more formally any value that is at least 1 day and less than 14 days, can only be set if the special `allow-unsafe-settings` is used in the relevant management canister APIs.
```
mraszyk, line 2126: From the discussion in the Interface Spec meeting: how about introducing a "hard" lower bound of one week and dropping the "soft" lower bound (and the additional fields `allow_unsafe_settings`)?

dsarlis: Sounds good to me. I'll update accordingly.

dsarlis: Done.

mraszyk: Thank you!

Dfinity-Bjoern (approved):
Looks good, please merge in coordination with roll-out to mainnet.

PR #5514, opened 2025-03-06
point to to imperative Map contribution instead of functional (pure) map

ByronBecker:
In the https://internetcomputer.org/blog/features/new-motoko-base announcement, the post references the adapted StableHeapBTreeMap data structure as a functional (pure implementation), but it’s actually an imperative implementation.

Here’s the imperative implementation of Map for reference.
https://github.com/dfinity/new-motoko-base/blob/35f2ce2580f089c925d768452043c4ed25ee8847/src/Map.mo#L24

I've updated this in the post.


jessiemongeon1 (approved):

rvanasa (approved):
Thanks for catching that!

PR #5521, opened 2025-03-10
feat: specify ic0.root_key

mraszyk:
This PR specifies new system API  `ic0.root_key_size` and `ic0.root_key_copy` for fetching the public key of the IC root key. This API can only be called in replicated execution to make canisters refrain from fetching the public key of the IC root key in an untrustworthy way.


Dfinity-Bjoern (approved):
Thanks, Martin! Please proceed with merging in line with the deployment to mainnet.

PR #5564, opened 2025-03-20
chore(ICRC_Ledger): Add quotes to search string in ICRC ledger suite release link

mbjorkqvist:
Without quotes, the current link to list the latest `ledger-suite-icrc` releases may also find ICP ledger suite releases. When adding the quotes, the search should only return ICRC ledger suite releases, as expected.


mbjorkqvist:
> 🤖 Here's your preview: https://hada6-4yaaa-aaaam-abaha-cai.icp0.io

In particular, the link in [this section](https://hada6-4yaaa-aaaam-abaha-cai.icp0.io/docs/defi/token-ledgers/setup/icrc1_ledger_setup#deploying-the-icrc-ledger-locally).

mbjorkqvist:
This actually does not yet affect the `ledger-suite-icrc`, but it does happen that if you look for `ledger-suite-icp` [without quotes](https://github.com/dfinity/ic/releases?q=%22ledger-suite-icp%22&expanded=true), an ICRC ledger suite release is found, whereas this does not happen when querying [with quotes](https://github.com/dfinity/ic/releases?q=%22ledger-suite-icp%22&expanded=true). It is likely that this would also happen in the future for ICRC ledger suite releases (depending on the commit messages), so adjusting the link in the docs does still make sense.

maciejdfinity (approved):

PR #5582, opened 2025-03-26
chore: Create submodule CI check 

jessiemongeon1:

  side navigation bar.


.github/workflows/submodule.yml
```diff
@@ -0,0 +1,46 @@
...
+        run: |
+          git config --global user.email "github-actions[bot]@users.noreply.github.com"
+          git config --global user.name "GitHub Actions"
+
+      - name: Update submodules to latest commit
+        run: git submodule update --remote --force
+
+      - name: Commit changes
+        run: |
+          git add .
```
marko-k0, line 26: Did you test this? Different scenarios? What if there's no changes?

.github/workflows/submodule.yml
```diff
@@ -0,0 +1,46 @@
...
+          git commit -m "Automated submodule update"
+          git push origin HEAD:automated-submodule-update
+
+      - name: Create Pull Request using peter-evans/create-pull-request
+        uses: peter-evans/create-pull-request@v3
```
marko-k0, line 31: You should pin this action to specific sha256. I think you can search and use the same sha as we have in IC repo.

.github/workflows/submodule.yml
jessiemongeon1: Yes, I've tested it in a fork of this repo. If there are no changes, it will simply fail.

jessiemongeon1: Added some logic to succeed if no changes necessary

marko-k0: Does this mean that PR will not be created if there are no changes to commit? It seems that that's not handled so I'd guess that PR with empty change-set might get created.

jessiemongeon1: No, the CI job exits with a success status and does not create a PR if there are no changes.

marc0olo (approved):

.github/workflows/check_submodule.yml
```diff
@@ -0,0 +1,43 @@
...
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          submodules: recursive
+          fetch-depth: 0
+
+      - name: Validate Submodule Commits
+        run: |
+          # Define latest release commits
```
sesi200, line 19:
```suggestion
          # Submodules are very easy to accidentally overwrite with old versions.
          # To make submodule updates more explicit and to prevent accidental changes we check the revisions explicitly here.
```

Maybe we can add a bit of context. Otherwise it looks quite pointless to do this check

sesi200 (approved):

PR #5651, opened 2025-04-15
A proposal for a System API exposing error codes to canisters

oggy-dfin:
This is a proposal for a System API extension exposing error codes to users.

  side navigation bar.

### DIFF
--- a/docs/references/ic-interface-spec.md
+++ b/docs/references/ic-interface-spec.md
@@ -1158,7 +1158,7 @@ When canisters explicitly reject a message (see [Public methods](#system-api-req
 
 ### Error codes {#error-codes}
 
-Implementations of the API can provide additional details for rejected messages in the form of a textual label identifying the error condition. API clients can use these labels to handle errors programmatically or suggest recovery paths to the user. The specification reserves error codes matching the regular expression `IC[0-9]+` (e.g., `IC502`) for the DFINITY implementation of the API.
+Implementations of the HTTPS and System APIs can provide additional details for rejected messages in the form of a textual label identifying the error condition. API clients and canisters can use these labels to handle errors programmatically or suggest recovery paths to the user. In the HTTPS API, the specification reserves error codes matching the regular expression `IC[0-9]+` (e.g., `IC502`) for the DFINITY implementation of the API.
 
 ### Status endpoint {#api-status}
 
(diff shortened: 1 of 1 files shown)


docs/references/ic-interface-spec.md
```diff
@@ -1158,7 +1158,7 @@ When canisters explicitly reject a message (see [Public methods](#system-api-req
 
 ### Error codes {#error-codes}
 
-Implementations of the API can provide additional details for rejected messages in the form of a textual label identifying the error condition. API clients can use these labels to handle errors programmatically or suggest recovery paths to the user. The specification reserves error codes matching the regular expression `IC[0-9]+` (e.g., `IC502`) for the DFINITY implementation of the API.
+Implementations of the HTTPS and System APIs can provide additional details for rejected messages in the form of a textual label identifying the error condition. API clients and canisters can use these labels to handle errors programmatically or suggest recovery paths to the user. In the HTTPS API, the specification reserves error codes matching the regular expression `IC[0-9]+` (e.g., `IC502`) for the DFINITY implementation of the API.
```
dsarlis, line 1161: Is this accurate given that the proposed system API would return only a numerical value? Unless I missed something you are proposing that instead of returning `IC502`, the system API would return `502`.

oggy-dfin: Oops. I rephrased this now to specifically say that labels are only used in the HTTPS API. Thanks!

mraszyk, line 1161:
```suggestion
Implementations of the HTTPS and System APIs can provide additional details for rejected messages in the form of a textual and numerical label identifying the error condition. API clients and canisters can use these labels to handle errors programmatically or suggest recovery paths to the user. In the HTTPS API, the specification reserves error codes matching the regular expression `IC[0-9]+` (e.g., `IC502`) for the DFINITY implementation of the API.
```

marc0olo:
This repository has been replaced by [dfinity/developer-docs](https://github.com/dfinity/developer-docs), live at [docs.internetcomputer.org](https://docs.internetcomputer.org), and is being archived, so no further merges will happen here. Closing this PR — if the change is still relevant, please open an equivalent PR against [dfinity/developer-docs](https://github.com/dfinity/developer-docs). Thanks for the contribution.

PR #5690, opened 2025-04-29
fix: improve stop canister specification

mraszyk:
This PR improves the specification of the management canister endpoint (method) to stop a canister:
- emphasizes that the canister is stopped if this method successfully returns (unless restarted by a separate concurrent call);
- points out that the call might return an error upon timeout and the effect on the canister status.


Dfinity-Bjoern (approved):
Thanks for the clarification!

dsarlis (approved):

PR #5728, opened 2025-05-09
feat: preconditions for installing code

mraszyk:
This PR specifies preconditions for installing canister code via the corresponding management canister endpoints that must be satisfied (if provided) for the call to succeed. The motivation is to prevent canisters that are not stopped from being installed and to enable compare-and-swap semantics for installing canisters (by atomically "comparing" the canister history before "swapping" the canister code).

### DIFF
--- a/docs/references/_attachments/ic.did
+++ b/docs/references/_attachments/ic.did
@@ -210,11 +210,17 @@ type canister_install_mode = variant {
     };
 };
 
+type install_code_preconditions = record {
+    stopped : opt bool;
+    total_num_changes : opt nat64;
+};
+
 type install_code_args = record {
     mode : canister_install_mode;
     canister_id : canister_id;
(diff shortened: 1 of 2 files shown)


Dfinity-Bjoern (approved):
Looks good, please proceed and merge in line with roll-out

mraszyk:
> Should we do the same for `load_canister_snapshot_args`?

Makes sense to me.

marc0olo:
This repository has been replaced by [dfinity/developer-docs](https://github.com/dfinity/developer-docs), live at [docs.internetcomputer.org](https://docs.internetcomputer.org), and is being archived, so no further merges will happen here. Closing this PR — if the change is still relevant, please open an equivalent PR against [dfinity/developer-docs](https://github.com/dfinity/developer-docs). Thanks for the contribution.

PR #5749, opened 2025-05-13
update nns hero

reigj1:
Updated nns hero

image


artkorotkikh-dfinity (approved):
Let's update the mobile image and it's ready to go!

PR #5796, opened 2025-05-28
feat: EXC-2060: Introduce environment variables

AlexandraZapuc:
TODO


docs/references/_attachments/ic.did
```diff
@@ -57,6 +61,9 @@ type change_details = variant {
     controllers_change : record {
         controllers : vec principal;
     };
+    environment_variables_change: record {
```
AlexandraZapuc, line 64: @raymondk The PR is still in progress, but could you please confirm whether the proposed solution meets your requirements? You might also want to sync with @keplervital, since he had input regarding the visibility and exposure of environment variables.

Proposed Solution Overview:

- 	Reading variables:

Canisters will be able to read environment variables directly. (Note: the System API for this is not yet included in the PR—it’s still in progress.)

- 	Setting variables:

Environment variables are configured via canister settings, using the existing update_settings and create_canister management APIs.

- 	Exposing variables:

	-	The canister_status call will expose all environment variables and their values.
	-	Only controllers will be able to view this information.
	-	A hash of the environment variables will be stored in the canister history, and a new entry will be added each time the env vars are changed. 
	 
From previous discussions, there seems to be two key concerns around exposing environment variables:
	1. Users want to verify which environment variables are set, as these directly impact canister behavior.
	2. Change detection: There needs to be a mechanism for users to detect when something has changed.

Given that only controllers have permission to set environment variables, it seems reasonable that only they can read the actual values since they will have the power to do anything with them(changing the behaviour of canister by changing the values). 
Meanwhile, the hash in the canister history serves as a signal that a change has occurred. 

Please let me know if this addresses the use cases you had in mind. I don't see a need for distinguishing between private and public variables if the concerns are covered by this solution.

raymondk: So the implication here is that there will be no such thing as a public environment variable. The environment variables and their values will always be private but the hash of them will always be public.

docs/references/ic-interface-spec.md
```diff
@@ -1545,6 +1545,14 @@ defaulting to `I = i32` if the canister declares no memory.
     ic0.cost_sign_with_schnorr : (src : I, size : I, algorithm: i32, dst : I) -> i32;     // * s
     ic0.cost_vetkd_derive_encrypted_key : (src : I, size : I, vetkd_curve: i32, dst : I) -> i32;  // * s
 
+    ic0.env_var_count() -> I;                                                             // *
+
+    ic0.env_var_name_size(index: I) -> I;                                                 // *
+    ic0.env_var_name_copy(dst: I, offset: I, size: I, index: I) -> ();                    // *
+
+    ic0.env_var_value_size(src: i, size: I) -> I;                                         // *
```
mraszyk, line 1553:
```suggestion
    ic0.env_var_value_size(src: I, size: I) -> I;                                         // *
```

docs/references/ic-interface-spec.md
```diff
@@ -2148,6 +2156,35 @@ These system calls return costs in Cycles, represented by 128 bits, which will b
...
+
+-   `ic0.env_var_name_size(index: I) -> I`; `I ∈ {i32, i64}`
+
+  Gets the size in bytes of the name of the environment variable at the given index.
+  
+-   `ic0.env_var_name_copy(dst: I, offset: I, size: I, index: I) -> ()`; `I ∈ {i32, i64}`
+
+  Copies the name of the environment variable at the given index into memory.
+
+-   `ic0.env_var_value_size(src: i, size: I) -> I`; `I ∈ {i32, i64}`
```
mraszyk, line 2175:
```suggestion
-   `ic0.env_var_value_size(src: I, size: I) -> I`; `I ∈ {i32, i64}`
```

docs/references/ic-interface-spec.md
```diff
@@ -2148,6 +2156,35 @@ These system calls return costs in Cycles, represented by 128 bits, which will b
...
+
+  Gets the size in bytes of the value for the environment variable with the given name.
+
+-   `ic0.env_var_value_copy(dst: I, offset: I, size: I, src: I, size: I) -> ()`; `I ∈ {i32, i64}`
```
mraszyk, line 2179:
```suggestion
-   `ic0.env_var_value_copy(dst: I, offset: I, size: I, index: I) -> ()`; `I ∈ {i32, i64}`
```

docs/references/ic-interface-spec.md
```diff
@@ -2278,6 +2315,14 @@ The optional `settings` parameter can be used to set the following settings:
 
     Default value: 0 (i.e., the "on low wasm memory" hook is never scheduled).
 
+-   `environment_variables` (`environment_variables`)
+
+    A record containing a vector of key-value pairs where both key and value are text strings. These variables are accessible to the canister during execution and can be used to configure canister behavior without code changes. Each key must be unique.
+
+    The maximum number of environment variables is implementation-defined. The maximum length of keys and values is implementation-defined.
+
+    Default value: `null` (i.e., no explicit environment variables provided).
```
mraszyk, line 2324:
```suggestion
    Default value: `null` (i.e., no environment variables provided).
```
I guess there are no "implicit" env variables, right?

docs/references/_attachments/ic.did
oggy-dfin: Note that if the hash is visible this makes the variables potentially vulnerable to a brute force hash preimage search, if they have limited entropy. It's not the end of the world because the controllers can always add entropy by prefixing the actual value with a salt, for example, but it's a potential footgun.

mraszyk:
I'd also propose the following additional changes:
- record env variables in the canister history when creating a canister;
- handle env variables in `provisional_create_canister_with_cycles`;
- add conditions (in the less formal part) under which the system APIs trap (index of out bounds, memory pointers out of bounds) analogously to other system APIs;
- add formal definitions of the new system APIs (section https://internetcomputer.org/docs/references/ic-interface-spec#concrete-canisters) incl. their formal trap conditions.

docs/references/ic-interface-spec.md
```diff
@@ -2451,7 +2498,7 @@ This method can only be called by canisters, i.e., it cannot be called by extern
...
+The canister history consists of a list of canister changes (canister creation, code uninstallation, code deployment, snapshot restoration, or controllers or environment variable change). Every canister change consists of the system timestamp at which the change was performed, the canister version after performing the change, the change's origin (a user or a canister), and its details. The change origin includes the principal (called *originator* in the following) that initiated the change and, if the originator is a canister, the originator's canister version when the originator initiated the change (if available). Code deployments are described by their mode (code install, code reinstall, code upgrade) and the SHA-256 hash of the newly deployed canister module. Loading a snapshot is described by the canister version, snapshot ID and timestamp at which the snapshot was taken. Canister creations and controllers changes are described by the full new set of the canister controllers after the change. The order of controllers stored in the canister history may vary depending on the implementation.
```
mraszyk, line 2501: Please also include a high-level description of what is recorded for env variable changes.

docs/references/ic-interface-spec.md
```diff
@@ -4884,17 +4940,39 @@ S.canister_history[A.canister_id] = {
...
       };
+    }]
+  else
+    [])
+
+if A.settings.controllers is not null or A.settings.environment_variables is not null:
+  New_canister_history = {
+    total_num_changes = N + 
+      (if A.settings.controllers is not null then 1 else 0) +
+      (if A.settings.environment_variables is not null then 1 else 0);
```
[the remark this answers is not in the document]
mraszyk: Note that adding a new variant (for a change of env variables) is already a breaking change anyway.

docs/references/ic-interface-spec.md
```diff
@@ -2278,6 +2315,14 @@ The optional `settings` parameter can be used to set the following settings:
 
     Default value: 0 (i.e., the "on low wasm memory" hook is never scheduled).
 
+-   `environment_variables` (`environment_variables`)
+
+    A record containing a vector of key-value pairs where both key and value are text strings. These variables are accessible to the canister during execution and can be used to configure canister behavior without code changes. Each key must be unique.
```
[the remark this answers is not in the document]
AlexandraZapuc: @raymondk should answer this.

I suggested blob, however he mentioned that because of usability he is proposing text instead.

letmejustputthishere:
this sounds pretty cool, where can i learn more about it? searching on slack or drive didn't yield any results :(

docs/references/_attachments/ic.did
AlexandraZapuc: I am not sure how much of a security issue is this? @oggy-dfin These are some env variables that can be manipulated by controllers only, the keys are not really predictable ( do not follow a specific pattern). Are you proposing to take an action item here  for this PR or simply raising a concern?

AlexandraZapuc: @mraszyk fyi, because you do a super in-depth review, I am still waiting on the confirmation from @raymondk. However, you're input is appreciated here.

marc0olo:
it would also be good to update the canister status example output to reflect the change. I don't think that is the case yet.

docs/references/_attachments/ic.did
oggy-dfin: The fact that the map differs per canister changes anything materially - an attacker may well know exactly what environment variables a canister expects to have. It's true that a single high-entropy value in the map is enough, but if we don't tell people to put one there they may shoot themselves in the foot. We could add docs and say it's their own problem. We could also nudge them by changing the API for setting env variables such that the controllers need to provide a salt together with the variables, and we add that to the hash. We could also not publish the hash in the first place.

> Regarding your example of storing credentials in environment variables: while technically possible, it’s generally not considered a best practice.

I disagree, it's quite common - it's the reason why GH secrets exist for example.

mraszyk: > We could also nudge them by changing the API for setting env variables such that the controllers need to provide a salt together with the variables, and we add that to the hash

I'm not sure we should make the API more complex just to nudge people to do the "right" thing.

dsarlis: > @dsarlis We talked about adding a hash of the init args to the history also, aren't the risks similar in both cases?

Yes they are indeed. And as we discussed the fact that the hash of init args is not present in the history is a miss as it does not allow for full verification of what's deployed.

AlexandraZapuc: To conclude the topic about the risk of exposing the hash, I am not against making the documention clearer. I will work with Martin on this to get something decent. But I wanted to emphasise that is a minimal risk and a quite expensive attack even with a small example.  

I thank you all for the input, I will put the PR in review, once I have the changes done.

docs/references/_attachments/ic.did
```diff
@@ -54,9 +62,10 @@ type change_details = variant {
       snapshot_id : snapshot_id;
       taken_at_timestamp : nat64;
     };
-    controllers_change : record {
-        controllers : vec principal;
-    };
```
mraszyk, line 59: Do we want to migrate existing instances of this variant in the state to the new variant (e.g., upon replica upgrade)? If not, then we should keep this variant in the spec.

docs/references/ic-interface-spec.md
```diff
@@ -2148,6 +2156,70 @@ These system calls return costs in Cycles, represented by 128 bits, which will b
     - `1`: Invalid curve or algorithm. Memory at `dst` is left unchanged.
     - `2`: Invalid key name for the given combination of signing scheme and (valid) curve/algorithm. Memory at `dst` is left unchanged.
 
+ This system call traps if `src+size` exceeds the size of the WebAssembly memory or `offset+size` exceeds the size of the stable memory.
+
+-   `ic0.stable64_read : (dst : i64, offset : i64, size : i64) → ()`
+
+    Copies the data from location \[offset, offset+size) of the stable memory to the location \[dst, dst+size) in the canister memory.
+
+     This system call traps if `dst+size` exceeds the size of the WebAssembly memory or `offset+size` exceeds the size of the stable memory.
```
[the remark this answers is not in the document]
AlexandraZapuc: Yes, forgot to remove it.

docs/references/ic-interface-spec.md
```diff
@@ -2148,6 +2156,70 @@ These system calls return costs in Cycles, represented by 128 bits, which will b
...
+      - The name does not correspond to a valid environment variable
+
+-   `ic0.env_var_value_copy(name_src: I, name_size: I, dst: I, offset: I, size: I) -> ()`; `I ∈ {i32, i64}`
+
+    Copies the value of the environment variable with the given name into memory.
+
+    This system call traps if:
+      - Invoked from the `(start)` function
+      - `name_src+name_size` exceeds the size of the WebAssembly memory
+      - If the data referred to by `name_src`/`name_size` is not valid UTF8.
```
mraszyk, line 2213: also applies for the previous system API

docs/references/ic-interface-spec.md
```diff
@@ -2451,7 +2533,7 @@ This method can only be called by canisters, i.e., it cannot be called by extern
...
+The canister history consists of a list of canister changes (canister creation, code uninstallation, code deployment, snapshot restoration, or canister settings change). Every canister change consists of the system timestamp at which the change was performed, the canister version after performing the change, the change's origin (a user or a canister), and its details. The change origin includes the principal (called *originator* in the following) that initiated the change and, if the originator is a canister, the originator's canister version when the originator initiated the change (if available). Code deployments are described by their mode (code install, code reinstall, code upgrade) and the SHA-256 hash of the newly deployed canister module. Loading a snapshot is described by the canister version, snapshot ID and timestamp at which the snapshot was taken. Canister settings changes are described by the complete updated set of canister controllers following the change, along with a hash of the environment variables, if any variables were modified. Note that the order of controllers stored in the canister history may vary depending on the implementation.
```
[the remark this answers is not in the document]
AlexandraZapuc: I did not link it because we use hash_of_map in a lot of other places. If you really want this link, I would prefer to link it in one single place. Open to suggestion where that would be.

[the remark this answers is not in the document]
mraszyk: > I did not link it because we use hash_of_map in a lot of other places.

There's no reference to `hash_of_map` in this paragraph though. It only says "hash of the environment variables" without making that precise.

docs/references/ic-interface-spec.md
```diff
@@ -2148,6 +2156,70 @@ These system calls return costs in Cycles, represented by 128 bits, which will b
...
+
+
+-   `ic0.env_var_value_size(name_src: I, name_size: I) -> I`; `I ∈ {i32, i64}`
+
+    Gets the size in bytes of the value for the environment variable with the given name.
+
+    This system call traps if:
+      - Invoked from the `(start)` function
+      - `name_src+name_size` exceeds the size of the WebAssembly memory
+      - The name does not correspond to a valid environment variable
```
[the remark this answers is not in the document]
AlexandraZapuc: Let me rephrase = > The name does not match any existing environment variable.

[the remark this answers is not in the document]
mraszyk: > The name does not match any existing environment variable.

This is indeed a better formulation, but the issue still exists: how would a CDK handle accessing an environment variable whose existence is not guaranteed, e.g., in
```
fn read_env_var(name: String) -> Option<String>
```

mraszyk:
> it would also be good to update the canister status example output to reflect the change. I don't think that is the case yet.

This PR is a core protocol specification PR. Developer tooling and examples will be updated in a follow up PR (the core protocol change must be rolled out first and only then developer tooling will be updated).

docs/references/ic-interface-spec.md
```diff
@@ -2148,6 +2156,56 @@ These system calls return costs in Cycles, represented by 128 bits, which will b
...
+    This system call traps if:
+      - `name_src+name_size` exceeds the size of the WebAssembly memory
+      - If the data referred to by `name_src`/`name_size` is not valid UTF8. 
+      - The name does not match any existing environment variable.
+      - `offset+size` is greater than the size of the environment variable value
+      - `dst+size` exceeds the size of the WebAssembly memory
+
+These system calls allow canisters to:
+- Enumerate all environment variables
+- Access variable names by index
```
mraszyk, line 2206: This seems subsumed by the previous point.

docs/references/ic-interface-spec.md
```diff
@@ -4949,17 +5033,20 @@ S.canister_history[A.canister_id] = {
...
+    recent_changes = H · SettingsChange {
+      controllers = if A.settings.controllers is not null then
+          opt A.settings.controllers;
+        else 
+          null;
+      environment_variables_hash = if A.settings.environment_variables is not null then
+          opt hash_of_map(A.settings.environment_variables);
+        else
+          null;
+    }
```
mraszyk, line 5049:
```suggestion
    recent_changes = H · H · {
        timestamp_nanos = S.time[A.canister_id];
        canister_version = S.canister_version[A.canister_id] + 1;
        origin = change_origin(M.caller, A.sender_canister_version, M.origin);
        details = SettingsChange {
          controllers = if A.settings.controllers is not null then
            opt A.settings.controllers;
          else 
            null;
          environment_variables_hash = if A.settings.environment_variables is not null then
            opt hash_of_map(A.settings.environment_variables);
          else
            null;
        };
      };
```

docs/references/_attachments/ic.did
```diff
@@ -8,6 +8,11 @@ type log_visibility = variant {
     allowed_viewers : vec principal;
 };
 
+type environment_variables = vec record {
```
raymondk, line 11: Does the max size of the value count as an implementation detail or does it need to specified in the spec?

skilesare:
Just as documentation, I've updated ICRC-105 to specify how to handle environment variable and their changes.  Ideally(as with other replica changes) we'd love a lifecycle event to make this much less complicated to track.  While I see that the hash of these changes are covered in the canister history API, the aim of 105 is to make full transparency of all config(software or environment variables), installs, configs, etc first class citizens in the ICRC-3 log.  Enumeration is not a must as hopefully the developer knows which variables he is using and can review any changes to the hash, but enumeration would certainly be welcome. See more: https://github.com/dfinity/ICRC/issues/105

mraszyk:
> enumeration would certainly be welcome

Depending on the size limits on canister environment variables, this could bloat up the canister history size substantially. Hence, a hash is preferrable (analogously to the WASM module hash that must also be reproduced separately).

mraszyk:
> we'd love a lifecycle event to make this much less complicated to track

What do you mean more concretely? And what would the workflow look like?

skilesare:
> What do you mean more concretely? And what would the workflow look like?

From:
https://forum.dfinity.org/t/canister-lifecycle-hooks/17089/38?u=skilesare

While we are here, there are a number of 'other' hooks that would be very helpful and one topic that would be worth discussing here in regards to the work I've been doing on ICRC-105. https://github.com/dfinity/ICRC/issues/105

The 'spirit' of ICRC-105 is to give public utility canisters a spot in the ICRC-3 log to record any configuration, upgrade, or installation changes that may be relevant to the execution of the canister.  The ideal 'best-case' scenario would be that given an ICRC-105 + ICRC-3 canister, the set of used wasms, properly recorded ICRC-3 block(proper being that they recorded all information relevant to the deterministic execution program/state changes) that one *could* reconstruct the state of the canister from the ICRC-3 log.  This means we need to record *any* change to the canister the would affect how the code runs.  Since you can write any code there may be a number of unknown unknowns that a dev would be responsible for, but it would be nice to take care of the known possible changes via canister life cycle events.

The obvious events would be
- Change of Controller(give me the new list)
- Change of canister settings(like **compute_allocation**, **memory_allocation**, **freezing_threshold**, **reserved_cycles_limit**, **wasm_memory_limit**, **log_visibility** (give me the new setting)
- A canister freeze(let this run once if my canister is frozen)
- A change of the environment variables(I understand this is under development) (let me enumerate the items or give them to me)
- A subnet change(when subnet migration is possible)(new subnet id)
- Edit: Canister Installed/upgraded (give me the new hash) - right now I have to do an timer+async call from my initializer and it delays and theoretically, something else could get called before I log the new hash...thus I have two blocks, one to record the install/upgrade and then a follow on to record the new hash.

One that is not so obvious would be something that fires if the version of the replica changes.  I'm not sure what the process or workflow is there, but I would imagine that if a replica version is updated it might include code that could change the way a canister behaves(something like the cycle cost of certain operations or even system API level information).  It may not be a great feature to have every canister on the subnet firing an event each time a new replica version starts up, but perhaps this could fire the first time that canister is called in the future?

docs/references/ic-interface-spec.md
```diff
@@ -1515,38 +1515,46 @@ defaulting to `I = i32` if the canister declares no memory.
...
+    ic0.cost_sign_with_ecdsa : (src : I, size : I, ecdsa_curve: i32, dst : I) -> i32;                 // * s
+    ic0.cost_sign_with_schnorr : (src : I, size : I, algorithm: i32, dst : I) -> i32;                 // * s
+    ic0.cost_vetkd_derive_key : (src : I, size : I, vetkd_curve: i32, dst : I) -> i32;                // * s
 
-    ic0.debug_print : (src : I, size : I) -> ();                                          // * s
-    ic0.trap : (src : I, size : I) -> ();                                                 // * s
+    ic0.env_var_count : () -> I;                                                                      // *
+
+    ic0.env_var_name_size : (index: I) -> I;                                                          // *
+    ic0.env_var_name_copy : (dst: I, offset: I, size: I, index: I) -> ();                             // *
```
mraszyk, line 1551: Maybe it'd make more sense to put `index` first: analogously to `name_src` and `name_size` being first for `ic0.env_var_value_{size,copy}`. @AlexandraZapuc

docs/references/ic-interface-spec.md
```diff
@@ -4841,6 +4855,15 @@ M.arg = candid(A)
 is_system_assigned Canister_id
 Canister_id ∉ dom(S.canisters)
 SubnetId ∈ Subnets
+// Environment variables validation conditions
```
mraszyk: please also include in provisional_create_canister_with_cycles

docs/references/ic-interface-spec.md
```diff
@@ -2156,6 +2177,67 @@ These system calls return costs in Cycles, represented by 128 bits, which will b
...
+
+    This system call traps if:
+      - The index is out of bounds (>= than value provided by `ic0.env_var_count`)
+      - `offset+size` is greater than the size of the environment variable name
+      - `dst+size` exceeds the size of the WebAssembly memory
+
+
+-   `ic0.env_var_name_exists : (name_src: I, name_size: I) -> i32`; `I ∈ {i32, i64}`
+
+    Checks if an environment variable with the given name exists.
```
lwshang, line 2207:
```suggestion
    Checks if an environment variable with the given name exists. If yes, then a value of 1 is returned, otherwise a 0 is returned.
```

To align with `is_controller` and `in_replicated_execution`.

Dfinity-Bjoern (approved):
Let's go!

docs/building-apps/canister-management/settings.mdx
```diff
@@ -13,7 +13,7 @@ import { GlossaryTooltip } from "/src/components/Tooltip/GlossaryTooltip";
 
 Each <GlossaryTooltip>canister</GlossaryTooltip> has a group of possible settings that control its behavior. Only a controller of the canister can read and modify a canister's settings.
 
-Currently, there are eight canister setting fields:
```
mraszyk, line 16: This PR should be merged when the roll-out is complete and changelog should be updated before merging.

docs/references/ic-interface-spec.md
```diff
@@ -2202,6 +2223,67 @@ These system calls return costs in Cycles, represented by 128 bits, which will b
...
+    Returns the number of environment variables set for this canister.
+
+-   `ic0.env_var_name_size : (index: I) -> I`; `I ∈ {i32, i64}`
+
+    Gets the size in bytes of the name of the environment variable at the given index.
+
+    This system call traps if:
+      - If the index is out of bounds (>= than value provided by `ic0.env_var_count`)
+
+-   `ic0.env_var_name_copy : (index: I, dst: I, offset: I, size: I) -> ()`; `I ∈ {i32, i64}`
```
letmejustputthishere, line 2241: it would be helpful to elaborate on the arguments here. i would have assumed the index and size are enough to retrieve the  copy. what are `dst` and `offset` needed for?

sesi200 (approved):

docs/references/ic-interface-spec.md
```diff
@@ -2202,6 +2223,67 @@ These system calls return costs in Cycles, represented by 128 bits, which will b
...
+
+    This system call traps if:
+      - The index is out of bounds (>= than value provided by `ic0.env_var_count`)
+      - `offset+size` is greater than the size of the environment variable name
+      - `dst+size` exceeds the size of the WebAssembly memory
+
+
+-   `ic0.env_var_name_exists : (name_src: I, name_size: I) -> i32`; `I ∈ {i32, i64}`
+
+    Checks if an environment variable with the given name exists. If yes, then a value of 1 is returned, otherwise a 0 is returned.
```
letmejustputthishere, line 2253: I think this is a bit misleading. I would have expected to pass the actual name of an environment variable to `env_var_name_exists` reading the text, or is that encoded as `i32` / `i64` in `name_src`? In that case, why is `name_size` needed? in general i feel like i'm missing context on how the storage of the variables is implemented. i thought it's an array of byte arrays

mraszyk: This is a low-level WASM system API that takes the name of the environment variable as input encoded by a pointer to the WASM memory `name_src` at which the name is stored (as UTF-8 encoded string) and the (binary) size `name_size` of the string. The result value is either 0 or 1 (encoded as `i32`). The low-level WASM system API cannot deal with high-level types (such as byte arrays) which is why they need to be represented by a pair of a pointer and size and this is how all existing system APIs are implemented.

PR #5835, opened 2025-06-12
EXC-2064: Add env_var_name_exists in SystemAPI 

AlexandraZapuc:
This PR adds a new System API function `ic0.env_var_name_exists` that allows canisters to check for the existence of environment variables without having to handle traps that would occur when trying to access non-existent variables.

### DIFF
--- a/docs/references/ic-interface-spec.md
+++ b/docs/references/ic-interface-spec.md
@@ -1549,7 +1549,8 @@ defaulting to `I = i32` if the canister declares no memory.
 
     ic0.env_var_name_size : (index: I) -> I;                                                          // *
     ic0.env_var_name_copy : (index: I, dst: I, offset: I, size: I) -> ();                             // *
-
+    ic0.env_var_name_exists : (name_src: I, name_size: I) -> I;                                       // *
+    
     ic0.env_var_value_size : (name_src: I, name_size: I) -> I;                                        // *
     ic0.env_var_value_copy : (name_src: I, name_size: I, dst: I, offset: I, size: I) -> ();           // *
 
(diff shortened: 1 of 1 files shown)


docs/references/ic-interface-spec.md
```diff
@@ -1549,7 +1549,8 @@ defaulting to `I = i32` if the canister declares no memory.
 
     ic0.env_var_name_size : (index: I) -> I;                                                          // *
     ic0.env_var_name_copy : (index: I, dst: I, offset: I, size: I) -> ();                             // *
-
+    ic0.env_var_name_exists : (name_src: I, name_size: I) -> I;                                       // *
```
mraszyk, line 1552:
```suggestion
    ic0.env_var_name_exists : (name_src: I, name_size: I) -> i32;                                       // *
```
since the return type is a Boolean (not a pointer or variable size)

docs/references/ic-interface-spec.md
```diff
@@ -2181,6 +2182,15 @@ The following system calls provide access to the canister's environment variable
       - `dst+size` exceeds the size of the WebAssembly memory
 
 
+-   `ic0.env_var_name_exists : (name_src: I, name_size: I) -> I`; `I ∈ {i32, i64}`
+
+    Checks if an environment variable with the given name exists.
+
+    This system call traps if:
+       - `name_src+name_size` exceeds the size of the WebAssembly memory
```
mraszyk, line 2190: There should also be a maximum length of the name (the concrete value omitted in the spec).

mraszyk (approved):

PR #5841, opened 2025-06-13
fix: out of memory error

mraszyk:
This PR fixes the section on out of memory error, based this replica implementation [fix](https://github.com/dfinity/ic/pull/5358) that has been rolled out to the ICP mainnet.


adambratschikaye (approved):
Thanks!

PR #5844, opened 2025-06-16
fix(ICP_Rosetta): FI-1767: Update example bytes for construction/combine response

mbjorkqvist:
This PR proposes to update the example bytes for the `signed_transaction` in the response to a `"construction/combine"` request. The reason for this is the change in ICP Rosetta in [this commit](https://github.com/dfinity/ic/commit/d3adf4958fc0dae7db3493294dfa71a70542c1ca), where
```
pub type SignedTransaction = Vec<Request>;
```
was changed to
```
pub struct SignedTransaction
    pub requests: Vec<Request>,
}
```
effectively wrapping the earlier bytes in an object. The content of the example response remains the same, it is only for representation that is slightly adjusted.


mbjorkqvist:
> 🤖 Here's your preview: https://aoaad-piaaa-aaaam-aekja-cai.icp0.io

Specifically here: https://aoaad-piaaa-aaaam-aekja-cai.icp0.io/docs/defi/rosetta/icp_rosetta/construction_api/operations-flow#fetching-a-signed-transaction

maciejdfinity (approved):

PR #5889, opened 2025-06-27
chore: CRP-2857 replace `set_user_rights` with `get_encrypted_vetkey` in DKMS dev docs

altkdf:
Getting an encrypted vetKeys is the more interesting part of the API for the devs.


fspreiss (approved):
Thanks, @altkdf!

jessiemongeon1 (approved):

PR #5907, opened 2025-07-02
Update `subnet_info` mgmt call

michael-weigelt:


michael-weigelt:
Implementation merged https://github.com/dfinity/ic/pull/5800

PR #5916, opened 2025-07-03
fix: Clarify cycles ledger info

jessiemongeon1:

  side navigation bar.

### DIFF
--- a/docs/defi/token-ledgers/cycles-ledger.mdx
+++ b/docs/defi/token-ledgers/cycles-ledger.mdx
@@ -22,6 +22,10 @@ It is important to note that the cycles ledger does not provide the ability to m
 
 If you need to make arbitrary calls to other canisters with cycles, a cycles wallet can be used.
 
+:::caution 
+When deploying a new canister, if a cycles wallet is configured, then `dfx` will use that, otherwise the cycles ledger will be used. 
+:::
+
 The cycles ledger complies with the ICRC-1, ICRC-2, and ICRC-3 standards. In addition to standard ledger functionality, it also interacts with the <GlossaryTooltip>NNS</GlossaryTooltip> canisters and user canisters to provide specific functions:
 
 - Accepting incoming cycles sent from other canisters.
@@ -119,9 +123,16 @@ All commands shown below are for workflows involving mainnet development. For lo
 |-----------------------|-----------------------|-----------------------|
 | Checking the cycles balance | `dfx wallet balance --network ic`  | `dfx cycles balance --network ic` |
 | Converting ICP tokens into cycles | `dfx ledger create-canister PRINCIPAL_ID --amount AMOUNT --network ic` |  `dfx cycles convert --amount AMOUNT --network ic` |
-| Transferring cycles |  `dfx wallet send PRINCIPAL_ID AMOUNT --network ic`   | `dfx cycles transfer AMOUNT PRINCIPAL_ID (--subaccount [SUBACCOUNT]) --network ic` |
 | Redeeming a cycles coupon |  `dfx wallet redeem-faucet-coupon 'ABCDE-ABCDE-ABCDE'`     |  `dfx cycles redeem-faucet-coupon ABCDE-ABCDE-ABCDE --network ic`  |
 | Topping up a canister     | `dfx wallet send PRINCIPAL_ID AMOUNT --network ic` | `dfx cycles top-up CANISTER_ID AMOUNT --network ic` |
 | Creating a canister with cycles | `dfx canister create CANISTER_NAME --with-cycles AMOUNT --network ic` |  `dfx canister create CANISTER_NAME --with-cycles AMOUNT --no-wallet --network ic` |
 
+### Transferring cycles from a wallet to the cycles ledger
 
+To transfer all cycles from a cycles wallet to the cycles ledger, then stop using the cycles wallet, run the following command:
+
+```
+dfx canister call --ic um5iw-rqaaa-aaaaq-qaaba-cai deposit "(record { to = record { owner = principal \"$(dfx identity get-principal)\"} })" --with-cycles 1_000_000_000_000 --wallet $(dfx identity get-wallet --ic)
+```
+
+Change the number of cycles to the amount you'd like to transfer.


docs/defi/token-ledgers/cycles-ledger.mdx
```diff
@@ -22,6 +22,10 @@ It is important to note that the cycles ledger does not provide the ability to m
 
 If you need to make arbitrary calls to other canisters with cycles, a cycles wallet can be used.
 
+:::caution 
+When deploying a new canister, if a cycles wallet is configured, then `dfx` will use that, otherwise the cycles ledger will be used.
```
mbjorkqvist, line 26: If I've previously used a cycles wallet, than transferred the cycles to the cycles ledger (as per the instructions added below), how can I then "unconfigure" the cycles wallet from `dfx`? From [the cycles wallet docs](https://internetcomputer.org/docs/building-apps/canister-management/cycles-wallet), would this be "removing a controller", "de-authorizing a principal", or something else? I think it would make sense to add information about this here and/or in the cycles wallet docs.

docs/defi/token-ledgers/cycles-ledger.mdx
```diff
@@ -119,9 +123,16 @@ All commands shown below are for workflows involving mainnet development. For lo
...
 | Checking the cycles balance | `dfx wallet balance --network ic`  | `dfx cycles balance --network ic` |
 | Converting ICP tokens into cycles | `dfx ledger create-canister PRINCIPAL_ID --amount AMOUNT --network ic` |  `dfx cycles convert --amount AMOUNT --network ic` |
-| Transferring cycles |  `dfx wallet send PRINCIPAL_ID AMOUNT --network ic`   | `dfx cycles transfer AMOUNT PRINCIPAL_ID (--subaccount [SUBACCOUNT]) --network ic` |
 | Redeeming a cycles coupon |  `dfx wallet redeem-faucet-coupon 'ABCDE-ABCDE-ABCDE'`     |  `dfx cycles redeem-faucet-coupon ABCDE-ABCDE-ABCDE --network ic`  |
 | Topping up a canister     | `dfx wallet send PRINCIPAL_ID AMOUNT --network ic` | `dfx cycles top-up CANISTER_ID AMOUNT --network ic` |
 | Creating a canister with cycles | `dfx canister create CANISTER_NAME --with-cycles AMOUNT --network ic` |  `dfx canister create CANISTER_NAME --with-cycles AMOUNT --no-wallet --network ic` |
 
+### Transferring cycles from a wallet to the cycles ledger
+
+To transfer all cycles from a cycles wallet to the cycles ledger, then stop using the cycles wallet, run the following command:
```
mbjorkqvist, line 132: Related to the above comment, regarding the "stop using the cycles wallet" part, what does this mean in practice? Would the recommended approach be to transfer out all the cycles and delete the canister, or would there be reasons to not spell out all the steps (here or somewhere else)? E.g., if the wallet is on a busy subnet where no more canisters can be created (except if spawned from existing canisters), this could be a reason to not delete the canister.

I was reminded of [this internal Slack discussion](https://dfinity.slack.com/archives/CGA566TPV/p1748353019160289), regarding transferring out all the cycles (to a wallet) and then deleting a canister. When calling `deposit` on the cycles ledger, is it the case that you cannot deposit the whole canister cycles balance, but you need to leave some "margin" to process the response (and something for the freezing threshold etc.)?

mbjorkqvist (commented):
Thanks for the fixes @jessiemongeon1! I left some clarifying/understanding questions which I believe would be mainly for the @sesi200 and/or @THLO.

docs/defi/token-ledgers/cycles-ledger.mdx
sesi200: Yes, you need to leave enough margin, otherwise you'll get a `Failed to send` error with no good error message.

I would suggest people leave the wallet around so that if they later notice they still need access to a canister that they don't control directly they can still control it

mbjorkqvist:
@jessiemongeon1 Would you be able to add the SDK team as codeowners for `docs/defi/token-ledgers/cycles-ledger.mdx` in `.github/CODEOWNERS`:
```
docs/defi/token-ledgers/cycles-ledger.mdx @dfinity/finint @dfinity/sdk
```
(I believe this needs to be added after the line with `docs/defi/ @dfinity/finint`.)

And also consider adapting the documentation based on @sesi200's replies to my questions (or creating another ticket/PR)?

docs/defi/token-ledgers/cycles-ledger.mdx
```diff
@@ -119,9 +123,16 @@ All commands shown below are for workflows involving mainnet development. For lo
 |-----------------------|-----------------------|-----------------------|
 | Checking the cycles balance | `dfx wallet balance --network ic`  | `dfx cycles balance --network ic` |
 | Converting ICP tokens into cycles | `dfx ledger create-canister PRINCIPAL_ID --amount AMOUNT --network ic` |  `dfx cycles convert --amount AMOUNT --network ic` |
-| Transferring cycles |  `dfx wallet send PRINCIPAL_ID AMOUNT --network ic`   | `dfx cycles transfer AMOUNT PRINCIPAL_ID (--subaccount [SUBACCOUNT]) --network ic` |
 | Redeeming a cycles coupon |  `dfx wallet redeem-faucet-coupon 'ABCDE-ABCDE-ABCDE'`     |  `dfx cycles redeem-faucet-coupon ABCDE-ABCDE-ABCDE --network ic`  |
 | Topping up a canister     | `dfx wallet send PRINCIPAL_ID AMOUNT --network ic` | `dfx cycles top-up CANISTER_ID AMOUNT --network ic` |
 | Creating a canister with cycles | `dfx canister create CANISTER_NAME --with-cycles AMOUNT --network ic` |  `dfx canister create CANISTER_NAME --with-cycles AMOUNT --no-wallet --network ic` |
 
+### Transferring cycles from a wallet to the cycles ledger
+
```
jessiemongeon1, line 131:
```suggestion

To transfer cycles from a cycles wallet to the cycles ledger, you must call the ledger’s deposit method. When doing so, avoid transferring the wallet's entire balance and ensure a small buffer of cycles remains to cover the transaction's execution cost. Failing to leave this margin will result in a `Failed to send` error.

Additionally, it’s recommended not to delete the cycles wallet, as it may still control a canister you might need to access in the future.
```

docs/defi/token-ledgers/cycles-ledger.mdx
jessiemongeon1, line 26:
```suggestion
When deploying a new canister, if a cycles wallet is configured, then `dfx` will use that, otherwise the cycles ledger will be used. 

A cycles wallet can be removed using `rm ~/.config/dfx/identity/<identity name>/wallets.json`.
```

jessiemongeon1:
Thanks @mbjorkqvist , I've applied @sesi200 's comments.

reigj1 (approved):

PR #5939, opened 2025-07-10
Add IPv4 support documentation

mihailjianu1:
IPv4 support was recently added, and it works by proxying relevant requests via servers running the SOCKS protocol.

We reflect those changes in the documentation


letmejustputthishere (approved):

docs/references/https-outcalls-how-it-works.mdx
```diff
@@ -39,8 +39,9 @@ As you can see, the transformation function and the Internet Computer Protocol's
...
-* It is assumed that the HTTP service is honest; otherwise, it can provide any responses to any of the calling replicas of the subnet. This is also the case for calls to the service from a Web 2.0 service or an oracle; thus, the situation is the same as in our scenario of a blockchain making the call to the service.
-* The trust model of ICP assumes that at least $2/3$ of the replicas are honest. In this case, the honest replicas will obtain the same response (after optional transformation), and consensus will be able to agree on the response so that the response can be provided back to the calling canister.
+* It is assumed that the HTTP service is honest; otherwise, it can provide any response to the calling replicas. This is a standard assumption when interacting with any external API.
+* The trust model of ICP assumes that at least 2/3 of the replicas are honest. This trust in the protocol extends to the IC-managed infrastructure that proxies certain requests (e.g., to IPv4 addresses). This ensures that honest replicas can reach a consensus on the response to be returned to the canister.
```
oggy-dfin, line 43: We only need to trust the proxies for liveness, I think this is worth mentioning. How about:
```suggestion
* The trust model of ICP assumes that at least 2/3 of the replicas are honest. Moreover, the IC-managed infrastructure that proxies certain requests (e.g., to IPv4 addresses) is assumed to not block traffic. This ensures that honest replicas can reach a consensus on the response to be returned to the canister. Note that the proxy infrastructure cannot tamper with the response, it can only block it.
```

oggy-dfin (approved):
Thanks for this! Left one suggestion.

PR #5994, opened 2025-08-06
CRP-2808 Add the master public keys for ECDSA, Schnorr, and VetKD to the documentation

randombit:


letmejustputthishere:
would be great if we can point to examples/logic on how to do the derivation for t-ecdsa & schnorr as well, an maybe provide a way for users to verify the authenticity of the master public key (also discussed [here](https://dfinity.slack.com/archives/CH07MRJQ7/p1745481208429349))

letmejustputthishere (approved):

randombit:
> would be great if we can point to examples/logic on how to do the derivation for t-ecdsa & schnorr as well

That is coming - just at the moment there aren't any easily accessible libraries for actually doing this for ECDSA or Schnorr. I'm working on adding support to the relevant packages; once these are modified and released I'll do a followup PR here adding examples.

PR #6003, opened 2025-08-12
Update docs to reflect 2TiB subnet memory capacity and 750GiB threshold

ShuoWangNSL:

  side navigation bar.


ShuoWangNSL:
The subnet memory capacity is increased to 2TiB in this [PR](https://github.com/dfinity/ic/pull/6086) and it has been rolled to mainnet. This PR updates the number in docs.

letmejustputthishere:
we should make sure we catch all places that refer to this limit, e.g. [here](https://github.com/dfinity/portal/blob/db8980161e78f1703b83298aef7917617455ffb2/docs/building-apps/essentials/gas-cost.mdx?plain=1#L124). ideally we find a solution that does not require this and we define constant in a single place we can refer to

ShuoWangNSL:
> we should make sure we catch all places that refer to this limit, e.g. [here](https://github.com/dfinity/portal/blob/db8980161e78f1703b83298aef7917617455ffb2/docs/building-apps/essentials/gas-cost.mdx?plain=1#L124)

Thanks for pointing it out. I also updated other places.

ShuoWangNSL:
> ideally we find a solution that does not require this and we define constant in a single place we can refer to

@letmejustputthishere  That's a great idea. It seems I did not find existing patterns of setting constants and using it everywhere. Do you have any suggestions and I will happy to follow. Or I could figure out a new way of doing this.

letmejustputthishere:
> That's a great idea. It seems I did not find existing patterns of setting constants and using it everywhere. Do you have any suggestions and I will happy to follow. Or I could figure out a new way of doing this.

I don't think we're currently doing it in our docs, at least I haven't come across it. Would be really cool if you can find a way to achieve this 🤩

ShuoWangNSL:
> I don't think we're currently doing it in our docs, at least I haven't come across it. Would be really cool if you can find a way to achieve this 🤩

Hi @letmejustputthishere , I will try to find a way this week. What do you think of merging this PR first so that the doc becomes update-to-date as we have already publicly announced this change through media. And I will have some more time trying new ways.

letmejustputthishere (approved):

ShuoWangNSL:
> sounds good to me, feel free to merge this

Thanks. FYI, I have tried several approaches in my new [PR](https://github.com/dfinity/portal/pull/6047) with the help of LLM as I did not have much experience with JS, but the preview link suggested they did not work well so far.

PR #6005, opened 2025-08-12
feat: snapshot source in canister history

mraszyk:
This PR specifies that canister history provides the source of a snapshot in the entry for loading the snapshot.

Implementation PR: https://github.com/dfinity/ic/pull/6163


Dfinity-Bjoern (approved):

docs/references/_attachments/ic.did
```diff
@@ -53,6 +53,10 @@ type change_details = variant {
         canister_version : nat64;
         snapshot_id : snapshot_id;
         taken_at_timestamp : nat64;
+        source : variant {
```
mraszyk, line 56: Changelog should be updated (shortly - to avoid conflicts) before merging this PR. The PR should be merged when the implementation of the feature is rolled out to all subnets.

PR #6008, opened 2025-08-13
Add a note on sig existence guarantees

oggy-dfin:
Previously we didn't specify anything about what happens when a call to a signing point fails for whatever reason. Clarify to say that we don't guarantee that no signature was computed, so that applications don't rely on this signature being unique.

  side navigation bar.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -22,7 +22,7 @@ docs/defi/chain-key-tokens/ @dfinity/cross-chain-team @dfinity/editorial
 docs/building-apps/chain-fusion/ @dfinity/cross-chain-team @dfinity/editorial
 
 # ProdSec
-docs/building-apps/best-practices/security/ @dfinity/product-security @dfinity/editorial
+docs/building-apps/security/ @dfinity/product-security @dfinity/editorial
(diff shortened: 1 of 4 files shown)


gregorydemay:
> When we say "Note that, in some cases" we could maybe mention/link to bounded-wait calls? In general, my understanding was that we would make the documentation more explicit about using bounded-wait calls and threshold signatures. Do you plan to add anything specific in the documentation of bounded-wait calls?

I also find the addition, while worthwhile, not specific enough. I think we should have a table comparing threshold signatures with bounded-wait calls vs unbounded-wait calls.

oggy-dfin:
Sorry, dropped the ball on this one. Would the following formulation be satisfactory?

> Note that, in some cases, a signature may be computed when requested, but not returned to the requesting canister. Thus, canisters should not rely on the signature not existing even if the signing request fails. In particular, the signature existing despite the signing request failing is more likely to happen if the request is made using a bounded-wait call to the management canister which times out. However, it may also happen in other request failure cases too (regardless of the whether the request was done using a bounded- or unbounded-wait call).

@andreacerulli @gregorydemay

gregorydemay (approved):
Thanks @oggy-dfin for driving this!

PR #6075, opened 2025-09-16
feat: allow loading a canister snapshot onto a different canister

mraszyk:
This PR specifies that a canister snapshot can be loaded onto a canister that belongs to the same subnet as the canister for which the snapshot was taken.


Dfinity-Bjoern (commented):
Looks good, thanks!

PR #6076, opened 2025-09-17
fix: specification of ic0.debug_print

mraszyk:
This PR
- fixes the specification of `ic0.debug_print` which is no longer a no-op, but appends data to canister logs;
- fixes the specification of cleanup callbacks by not enumerating allowed system API calls (the list was outdated).

### DIFF
--- a/docs/references/ic-interface-spec.md
+++ b/docs/references/ic-interface-spec.md
@@ -1804,7 +1804,7 @@ Subsequent calls to the following functions set further attributes of that call,
 
 If a cleanup callback (of type `(env : I) -> ()`) is specified for this call, it is executed if and only if the `reply` or the `reject` callback was executed and trapped (for any reason).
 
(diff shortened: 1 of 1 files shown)


docs/references/ic-interface-spec.md
```diff
@@ -2184,21 +2184,21 @@ These system calls return costs in Cycles, represented by 128 bits, which will b
 
 ### Debugging aids
 
-In a local canister execution environment, the canister needs a way to emit textual trace messages. On the "real" network, these do not do anything.
+Canister can produce logs available through the management canister endpoint [`fetch_canister_logs`](#ic-fetch_canister_logs).
```
dsarlis, line 2187:
```suggestion
Canister can fetch logs available through the management canister endpoint [`fetch_canister_logs`](#ic-fetch_canister_logs).
```

dsarlis (approved):
Seems like we need a more generic title given that we're performing fixes across some more system APIs than just `ic0.debug_print`.

Dfinity-Bjoern (approved):
Thanks @mraszyk and @dsarlis for these cleanups!

PR #6078, opened 2025-09-23
feat: Improve explanation for sharded canister ranges in spec

Andrei1998:
Following up on https://github.com/dfinity/portal/pull/5993#issuecomment-3210198830, this PR provides a more easy-to-digest description for the sharded canister ranges path in the state tree.

See also this related PR that moved the explanation further down in the spec: https://github.com/dfinity/portal/pull/6065.


Dfinity-Bjoern (approved):
Thanks, @Andrei1998!

PR #6087, opened 2025-09-25
docs(defi): move rosetta to bottom of sidebar

ielashi:
Moves "ICP Rosetta" and "ICRC1 Rosetta" to the bottom of the sidebar of the documentation. Rosetta is a very niche service that only concerns a small number of institutional clients, so it doesn't make sense to be at the top of the sidebar.

Before:
grafik

After:
grafik

nit: I also formatted the JSON in this document as it wasn't properly aligned.


marc0olo (approved):
thanks! :)

PR #6101, opened 2025-10-06
New HTTP outcalls pricing model

oggy-dfin:
This PR describes the interface for the new, finer grained pricing model for HTTP outcalls, which should enable lowering the costs of outcalls by around 4x on average.

### DIFF
--- a/docs/references/ic-interface-spec.md
+++ b/docs/references/ic-interface-spec.md
@@ -2202,14 +2202,14 @@ These system calls return costs in Cycles, represented by 128 bits, which will b
 
     The cost of creating a canister on the same subnet as the calling canister via [`create_canister`](#ic-create_canister). Note that canister creation via a call to the CMC can have a different cost if the target subnet has a different replication factor.
 
-:::note
-
-The `ic0.cost_http_request` System API call is DEPRECATED. Canister developers are advised to use the `ic0.cost_http_request_v2` call instead.
-
-:::
-
 -   `ic0.cost_http_request(request_size : i64, max_res_bytes : i64, dst : I) -> ()`; `I ∈ {i32, i64}`
 
+    :::note
+    
+    The `ic0.cost_http_request` System API call is DEPRECATED. Canister developers are advised to use the `ic0.cost_http_request_v2` call instead.
+    
+    :::
+    
     The cost of a canister HTTP outcall via [`http_request`](#ic-http_request) with the pricing version set to `1` (currently the default). `request_size` is the sum of the byte lengths of the following components of an http request:
     - url
     - headers - i.e., the sum of the lengths of all keys and values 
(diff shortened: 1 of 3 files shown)


docs/references/ic-interface-spec.md
```diff
@@ -2197,14 +2204,48 @@ These system calls return costs in Cycles, represented by 128 bits, which will b
...
     
     `max_res_bytes` is the maximum response length the caller wishes to accept (the caller should provide the default value of `2,000,000` if no maximum response length is provided in the actual request to the management canister). 
 
+-   `ic0.cost_http_request_v2(params_src: I, params_size: I, dst : I) -> ();    I ∈ {i32, i64}`
+
+    The cost of a canister HTTP outcall via [`http_request`](#ic-http_request) with the pricing version set to `2`. The blob described by `params_src` and `params_size` must be a valid Candid encoding of a value of the following type:
+    ```
+    record {
+        request_bytes : nat64;
+        request_time_ms : nat64;
```
mihailjianu1, line 2227: Nit: maybe response_time_ms ?

mihailjianu1 (approved):

docs/references/ic-interface-spec.md
```diff
@@ -2918,11 +2959,13 @@ The following parameters should be supplied for the call:
...
-
-The `is_replicated` field is considered EXPERIMENTAL.
+    :::note
+    
+    The `is_replicated` field is considered EXPERIMENTAL.
+    
+    :::
 
-:::
+-   `pricing_version` - the version of the pricing mechanism for HTTP outcalls that should be applied to this call; it can be either `1` or `2`. For compatibility reasons, the default is `1`; however, version `1` is deprecated.
```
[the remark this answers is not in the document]
Dfinity-Bjoern: How big is the additional effort of failing in the case that `max_reponse_bytes` is set and `pricing_version` is 2? If the effort is one `if` statement (because we have to test the combination anywhere), then I'd suggest we just do that check.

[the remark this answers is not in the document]
Dfinity-Bjoern: After Oggy's explanation that we plan to drop pricing version 1 eventually but we'd want to keep supporting old clients in a best-effort way going forward, I agree it makes sense to silently ignore the combination for now (and not return an error). Thanks for the clarification!

docs/references/ic-interface-spec.md
```diff
@@ -2221,14 +2228,48 @@ These system calls return costs in Cycles, represented by 128 bits, which will b
...
+-   `ic0.cost_http_request_v2(params_src: I, params_size: I, dst : I) -> ();    I ∈ {i32, i64}`
+
+    The cost of a canister HTTP outcall via [`http_request`](#ic-http_request) with the pricing version set to `2`. The blob described by `params_src` and `params_size` must be a valid Candid encoding of a value of the following type:
+    ```
+    record {
+        request_bytes : nat64;
+        http_roundtrip_time_ms : nat64;
+        raw_response_bytes : nat64;
+        transformed_response_bytes : nat64;
+        transform_instructions: nat64;
```
gregorydemay: It would be more developer-friendly to have all those fields as `opt nat64` and describe the default values used in case they are not specified. I think it's a bit unclear for a developer what reasonable values should be put in `http_roundtrip_time_ms` or `transform_instructions`. WDYT?

mraszyk: That makes sense to me. What do you think, @mihailjianu1 ? (Switching from `nat64` to `opt nat64` is backward-compatible so it should be fine although the API has already been rolled out in production.)

PR #6102, opened 2025-10-07
feat: canister status query call

mraszyk:
This PR specifies management canister _query_ calls to the method `canister_status`.

### DIFF
--- a/docs/references/ic-interface-spec.md
+++ b/docs/references/ic-interface-spec.md
@@ -2529,6 +2529,7 @@ The optional `sender_canister_version` parameter can contain the caller's canist
 ### IC method `canister_status` {#ic-canister_status}
 
 This method can be called by canisters as well as by external users via ingress messages.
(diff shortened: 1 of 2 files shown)


docs/references/ic-interface-spec.md
```diff
@@ -2529,6 +2529,7 @@ The optional `sender_canister_version` parameter can contain the caller's canist
 ### IC method `canister_status` {#ic-canister_status}
 
 This method can be called by canisters as well as by external users via ingress messages.
+This method can also be called by external users via non-replicated (query) calls, but it cannot be called from composite query calls.
```
mraszyk, line 2532: It is true that it cannot be called from composite query calls, right? @dsarlis

dsarlis: Correct, it is not possible until we fix an existing issue that would allow in general any management canister calls to happen via composite queries (so it would also apply to `fetch_canister_logs`).

PR #6213, opened 2026-03-18
Disallow canister signatures from engines on mainnet

andreacerulli:
Canister signatures whose certificate includes a delegation from a cloud_engine subnet must now be rejected. This is enforced by adding a check in the canister signature verification rules: if the certificate includes a subnet delegation, the subnet type (at /subnet/<subnet_id>/type in the delegation's certificate) must not be cloud_engine. In particular, certificates omitting the subnet type from the delegation should not be accepted. 

This restriction applies only to canister signatures used for IC request authentication. Subnet delegations to cloud_engine subnets remain valid for other uses (e.g. certified data, query responses).


docs/references/ic-interface-spec.md
```diff
@@ -411,7 +411,7 @@ This section makes forward references to other concepts in this document, in par
 
     where `signing_canister_id` is the id of the signing canister and `reconstruct` is a function that computes a root-hash for the tree.
 
-    -   If the `certificate` includes a subnet delegation, then the `signing_canister_id` must be included in the delegation's canister id range (see [Delegation](#certification-delegation)).
+    -   If the `certificate` includes a subnet delegation, then the `signing_canister_id` must be included in the delegation's canister id range (see [Delegation](#certification-delegation)), and the subnet type (obtained by looking up `/subnet/<subnet_id>/type` in the delegation's certificate) *must not* be `cloud_engine`.
```
mraszyk:
```suggestion
    -   If the `certificate` includes a subnet delegation, then the `signing_canister_id` must be included in the delegation's canister id range (see [Delegation](#certification-delegation)), and the subnet type (obtained by looking up the path `/subnet/<subnet_id>/type` in the delegation's certificate which *must* be present) *must not* be `cloud_engine`.
```

mraszyk (commented):
Proposed a little clarification. Otherwise LGTM and let's merge when the change is rolled-out.

randombit (approved):

schneiderstefan (approved):

marc0olo (approved):

PR #6223, opened 2026-04-14
feat: query call for subnet admins to list all canisters on a subnet

mraszyk:
This PR                                                                                                                                 
  - Adds a new `list_canisters` query method to the IC Management Canister (`aaaaa-aa`) in the interface specification and Candid interface definition.
  - The method returns all canisters on a subnet as a list of consecutive canister ID ranges (`start`/`end` principal pairs); deleted canisters are excluded.
  - **Subnet admins only** — the caller must be a subnet admin for the relevant subnet.
  - **External users only** — not callable by canisters, not via replicated calls, and not from composite query calls.                                        
  - Any principal may be used as the effective canister ID for routing, consistent with other subnet-admin methods like `create_canister`.

### DIFF
--- a/docs/references/_attachments/ic.did
+++ b/docs/references/_attachments/ic.did
@@ -511,6 +511,15 @@ type fetch_canister_logs_result = record {
     canister_log_records : vec canister_log_record;
 };
 
+type canister_id_range = record {
+    start : canister_id;
+    end : canister_id;
+};
+
+type list_canisters_result = record {
+    canisters : vec canister_id_range;
+};
+
 type read_canister_snapshot_metadata_args = record {
     canister_id : canister_id;
     snapshot_id : snapshot_id;
@@ -683,4 +692,7 @@ service ic : {
 
     // canister logging
     fetch_canister_logs : (fetch_canister_logs_args) -> (fetch_canister_logs_result) query;
+
+    // subnet admin methods
+    list_canisters : () -> (list_canisters_result) query;
 };
--- a/docs/references/ic-interface-spec.md
+++ b/docs/references/ic-interface-spec.md
@@ -7884,7 +7884,7 @@ Query response `R`:
 
 ```
 
-where `CanisterIdRanges` is the list of all canister IDs on the subnet encoded as consecutive ranges (excluding deleted canisters), and the query `Q`, the response `R`, and a certificate `Cert` that is obtained by requesting the path `/subnet` in a **separate** read state request to `/api/v3/canister/<ECID>/read_state` satisfy the following:
+where `CanisterIdRanges` is the list of all canister IDs on the subnet encoded as consecutive canister ID ranges (excluding deleted canisters), and the query `Q`, the response `R`, and a certificate `Cert` that is obtained by requesting the path `/subnet` in a **separate** read state request to `/api/v3/canister/<ECID>/read_state` satisfy the following:
 
 ```html
 
--- a/docs/references/ic-interface-spec.md
+++ b/docs/references/ic-interface-spec.md
@@ -1058,6 +1058,8 @@ It must be contained in the canister ranges of a subnet, otherwise the correspon
 
 -   If the request is a query call to the Management Canister (`aaaaa-aa`), then:
 
+    -   If the call is to the `list_canisters` method, then any principal can be used as the effective canister id for this call.
+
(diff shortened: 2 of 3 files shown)


docs/references/_attachments/ic.did
```diff
@@ -683,4 +692,7 @@ service ic : {
 
     // canister logging
     fetch_canister_logs : (fetch_canister_logs_args) -> (fetch_canister_logs_result) query;
+
+    // subnet admin methods
+    list_canisters : () -> (list_canisters_result) query;
```
Dfinity-Bjoern, line 697: Do we have to take care of the situation where the return value does not fit into a single query response?

As we write below that deleted canisters are not part of the result, wouldn't that possibly make the data structure grow substantially, if I, say, create many canisters and then delete every other canister?

docs/references/ic-interface-spec.md
```diff
@@ -3272,6 +3274,24 @@ Replica-signed queries may improve security because the recipient can verify the
 
 :::
 
+### IC method `list_canisters` {#ic-list_canisters}
+
+This method can only be called by external users with subnet admin privileges via non-replicated (query) calls, i.e., it cannot be called by canisters, cannot be called via replicated calls, and cannot be called from composite query calls.
```
Dfinity-Bjoern, line 3279: Why is this only available in non-replicated mode? Doesn't it make sense to have a subnet admin that is a canister, such as supporting subnet administration via Orbit or from an SNS?

docs/references/_attachments/ic.did
mraszyk: It is very unlikely that the response would not fit into a single query call response (up to 3MB): one canister ID range takes ~20 bytes so it'd require ~150k canister ranges, i.e., ~150k "holes" due to deleted canisters.

In the future, we could introduce pagination by adding a new argument to the input type:
```
canister_ranges : opt vec canister_id_range
```
which is backward-compatible and could be used as a filter for the return value.

docs/references/ic-interface-spec.md
mraszyk: Enabling this API in replicated mode is not a requirement (at the moment) and it would require to benchmark its performance to determine how many instructions should be accounted for its execution (not to charge the admins, but to determine how many such calls to execute in a single round to not get a FR drop if admins crawl heavily). Hence, I'd suggest to enable replicated execution when this becomes a requirement.

docs/references/ic-interface-spec.md
```diff
@@ -4351,6 +4371,7 @@ A `Request` has an effective canister id according to the rules in [Effective ca
 ```
 is_effective_canister_id(Request {canister_id = ic_principal, method = create_canister, …}, p)
 is_effective_canister_id(Request {canister_id = ic_principal, method = provisional_create_canister_with_cycles, …}, p)
+is_effective_canister_id(Request {canister_id = ic_principal, method = list_canisters, …}, p)
```
mraszyk:
```suggestion
is_effective_canister_id(CanisterQuery {canister_id = ic_principal, method = list_canisters, …}, p)
```

docs/references/ic-interface-spec.md
```diff
@@ -7822,9 +7843,58 @@ verify_response(Q, R, Cert) ∧ lookup(["time"], Cert) = Found S.system_time //
 
 ```
 
+#### IC Management Canister: List canisters (query call) {#ic-mgmt-canister-list-canisters}
+
+This section specifies the `list_canisters` management canister query call.
+It is a call to `/api/v3/canister/<ECID>/query`
```
schneiderstefan, line 7849: Does this endpoint not work with the `/v2/` variants?

mraszyk: It works with them, too, but the (formal) part of the spec omits deprecated endpoints for simplicity.

mraszyk: If you find this information useful, I'd add it consistently in a separate PR.

schneiderstefan: I think that was a good idea to omit the deprecated endpoints. I was mostly wondering for my own benefit, and I don't think that you should change that (in another PR or otherwise).

schneiderstefan (approved):

PR #6227, opened 2026-04-15
Update CODEOWNERS for document ownership assignments

raymondk:


marc0olo:
@raymondk can we revert the removal of the defi team? I assume they didn't have write access and this might be fixed now. otherwise we should tell idx again about this issue.

PR #6230, opened 2026-04-21
docs: simplify canister ID resolution and rename API Gateway to API Boundary Node in HTTP GW spec

r-birkner:
- Removes implementation-specific canister ID resolution logic (hardcoded hostname table, raw domain handling, DNS TXT lookup, `x-ic-canister-id` header detection) from the HTTP Gateway Protocol Spec, replacing it with a protocol-agnostic description that lists example approaches (subdomain encoding, DNS, static mapping, HTTP header).
- Renames "API Gateway" to "API Boundary Node" throughout the spec, including the section heading, body text, and cross-reference anchor link.


viviveevee (approved):

raymondk (approved):

mraszyk:
I suppose there was a high-level decision to do this, but I still wonder if there are any docs on how the canister ID resolution actually works now?

PR #6245, opened 2026-06-03
feat: allow PATCH in `http_request` for non-replicated calls

### LINKED ISSUE #6244: HTTPS outcalls: add PATCH to the `http_request` method variant
## Summary

Add `patch` to the `method` variant of `http_request_args` (the management-canister HTTPS-outcall API), so canisters can issue HTTP `PATCH` requests. Today the variant is:

```candid
method : variant { get; head; post; put; delete };
```

`PATCH` is the only common REST verb still missing after `put`/`delete` were added.

## Use case

We generate **Motoko client libraries from OpenAPI specs** (Caffeine connectors — each wraps a third-party REST API as a mops package that a canister calls via HTTPS outcalls). The generator emits one outcall per operation, mapping the spec's HTTP method straight onto the `method` variant.

The concrete trigger is a **Google Calendar API** client. Google's REST surface uses `PATCH` for every partial-update operation:

- `calendar.events.patch`
- `calendar.calendars.patch`
- `calendar.acl.patch`
- `calendar.calendarList.patch`

These simply cannot be expressed today — the generated code references `method = #patch`, which has no corresponding system variant, so those operations are un-callable. `PATCH` is pervasive across modern REST APIs (Google, GitHub, Stripe, …) wherever a partial update is preferred over a full-resource `PUT`, so this is not Calendar-specific.

## Why `PATCH` is well-behaved on the IC (and how it differs from the existing verbs)

The two IC-specific concerns for any outcall verb are **(1) how many times the remote server is hit** and **(2) whether replicas can agree on the response**. Here is how `PATCH` sits relative to the verbs already supported.

### 1. Double-request tolerance (idempotency)

In **replicated** mode an outcall is issued **once per replica** — every node on the subnet independently fires the same request, so the remote server sees N identical hits. Safety under N-fold delivery is exactly HTTP idempotency:

| Verb | Idempotent? | N copies to remote |
|------|-------------|--------------------|
| `get` / `head` | yes (nullipotent) | harmless — pure reads |
| `put` | **yes** (RFC 9110 §9.2.2) | converge to one final state |
| `delete` | **yes** | one delete |
| `post` | no | N resources / N appends |
| **`patch`** | **not guaranteed** (RFC 5789 §2) | depends on the patch document |

`PATCH` is the one verb whose idempotency is **not** guaranteed by the HTTP contract: it is idempotent when the patch body is a *field-set merge* (set these fields to these values) but **not** when it expresses a *relative* mutation (increment, append). This is the only semantic difference from `put`/`delete` — and it is **handled by exactly the same mechanism they already use**: forcing the call into non-replicated mode.

### 2. IC consensus (response agreement)

Even an idempotent verb breaks replicated consensus if the N responses don't match byte-for-byte. Mutation responses almost always echo server-assigned fields (`updated` timestamp, `etag`, generated id) that differ across the N independent calls, so without a normalizing `transform` the replicas can't agree and the call traps. This is why mutations are run **non-replicated** (`is_replicated = ?false`): a single node issues exactly one request, so there is one response and nothing to reach consensus over.

`PATCH` lands in the **same bucket as `put`/`delete`**: a mutation that should run non-replicated (`is_replicated = ?false`). The one residual difference — its weaker idempotency guarantee — only matters if the single non-replicated outcall is retried at the system level on a transient failure; clients that need that safety must already ensure their patch body is merge-shaped, exactly as they would reason about `put`.

In short: **`PATCH` introduces no new system-level concern that `put`/`delete` did not already raise**, and the existing `is_replicated` machinery already covers it.

## Proposed change

```candid
method : variant { get; head; post; put; delete; patch };
```

…with `patch` subject to the same replicated-mode handling already applied to `put`/`delete`.

## Offer to contribute

Happy to put up a PR following the **`put`/`delete` blueprint** (the change set that introduced those two verbs) if the team is open to it — it looks like a mechanical extension of the same enum + execution-path handling. Pointers to that original change set would be appreciated.

ggreif:
Resolves #6244.

Adds the `PATCH` HTTP method to the canister `http_request` (HTTPS outcalls) management-canister interface, mirroring the `PUT`/`DELETE` addition in #6199.

### What

- `ic.did`: `method` variant → `variant { get; head; post; put; delete; patch }`.
- `ic-interface-spec.md`: document `PATCH` alongside `PUT`/`DELETE` as supported in **non-replicated mode only**.
- `interface-spec-changelog.md`: new `0.62.0` entry.

### Why

`PATCH` is pervasive in modern REST APIs for partial updates (Google Calendar, GitHub, Stripe, …). It is currently the only common verb missing after `PUT`/`DELETE` landed, which blocks canisters from calling those endpoints. The concrete driver is a generated Motoko client for the Google Calendar API, whose partial-update operations (`events.patch`, `calendars.patch`, `acl.patch`, `calendarList.patch`) are otherwise un-callable.

### Why non-replicated only

Same rationale as `PUT`/`DELETE`: in replicated mode an outcall is issued once per replica, so a mutation would hit the remote N times, and mutation responses rarely agree byte-for-byte across replicas. `PATCH` additionally is not guaranteed idempotent (RFC 5789 §2) — it is idempotent only for field-merge bodies — which makes the non-replicated restriction even more appropriate. The existing `is_replicated` machinery already covers it; no new mechanism is introduced.

### Notes

- Spec/docs change only. The corresponding runtime implementation lives in `dfinity/ic`.
- Draft pending the runtime side / IC-team review.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/references/_attachments/interface-spec-changelog.md
+++ b/docs/references/_attachments/interface-spec-changelog.md
@@ -1,5 +1,8 @@
 ## Changelog {#changelog}
 
+### 0.62.0 (2025-06-03) {$0_62_0}
+* Support for the HTTP method `PATCH` in canister `http_request` in non-replicated mode.
+
 ### 0.61.0 (2025-05-18) {$0_61_0}
 * New management canister endpoint `canister_metrics`.
 
--- a/docs/references/_attachments/ic.did
+++ b/docs/references/_attachments/ic.did
@@ -344,7 +344,7 @@ type deposit_cycles_args = record {
 type http_request_args = record {
     url : text;
     max_response_bytes : opt nat64;
-    method : variant { get; head; post; put; delete };
+    method : variant { get; head; post; put; delete; patch };
     headers : vec http_header;
     body : opt blob;
     transform : opt record {
--- a/docs/references/ic-interface-spec.md
+++ b/docs/references/ic-interface-spec.md
@@ -3046,7 +3046,7 @@ In the replicated mode, the responses for all identical requests must match, too
 
 For this reason, the calling canister can supply a transformation function, which the IC uses to let the canister sanitize the responses from such unique values. The transformation function is executed separately on the corresponding response received for a request (both in replicated and non-replicated modes). Only the transformed response will be available to the calling canister.
 
-Currently, the `GET`, `HEAD`, and `POST` methods are supported for HTTP requests. Additionally, the `PUT` and `DELETE` methods are supported in non-replicated mode only. `PUT` and `DELETE` are restricted to non-replicated mode to avoid confusing race conditions that may occur with replicated execution.
+Currently, the `GET`, `HEAD`, and `POST` methods are supported for HTTP requests. Additionally, the `PUT`, `DELETE`, and `PATCH` methods are supported in non-replicated mode only. `PUT`, `DELETE`, and `PATCH` are restricted to non-replicated mode to avoid confusing race conditions that may occur with replicated execution.
 
 It is important to note the following for the usage of the `POST` method:
 
@@ -3070,7 +3070,7 @@ The following parameters should be supplied for the call:
 
 -   `max_response_bytes` - optional, specifies the maximal size of the response in bytes. If provided, the value must not exceed `2MB` (`2,000,000B`). The call will be charged based on this parameter. If not provided, the maximum of `2MB` will be used.
 
--   `method` - currently, `GET`, `HEAD`, and `POST` are supported. Additionally, `PUT` and `DELETE` are supported in non-replicated mode only.
+-   `method` - currently, `GET`, `HEAD`, and `POST` are supported. Additionally, `PUT`, `DELETE`, and `PATCH` are supported in non-replicated mode only.
 
 -   `headers` - list of HTTP request headers and their corresponding values
 


docs/references/_attachments/interface-spec-changelog.md
```diff
@@ -1,5 +1,8 @@
 ## Changelog {#changelog}
 
+### 0.62.0 (2025-06-03) {$0_62_0}
```
mraszyk, line 3: The date should be updated and this PR merged when the feature is rolled out on mainnet.

mraszyk (commented):
LGTM I'll approve when the feature is rolled out to prevent accidental merging.

ggreif:
> LGTM I'll approve when the feature is rolled out to prevent accidental merging.

Thank you @mraszyk, the implementation PR sits right here: dfinity/ic#10378!

marc0olo:
@ggreif @mraszyk we should make these changes in the https://github.com/dfinity/developer-docs repository

ggreif:
The content of this PR has been migrated to `dfinity/developer-docs` as requested: dfinity/developer-docs#289 (opened as a draft from branch `migrate-portal-6245`).

One item for human confirmation: this PR adds a `0.62.0` changelog entry, but `developer-docs` already has a `0.62.0` with different content (dated 2025-05-26). The migration used `0.63.0 (TBD)` as a placeholder — the correct version number should be confirmed with the IC team.

This PR (#6245) can be closed once the `developer-docs` PR is approved and merged.

ggreif:
Closing in favour of dfinity/developer-docs#289.

PR #6247, opened 2026-07-09
Remove security team from CODEOWNERS

tmu0:
Removes the entire `# ProdSec` section (5 `@dfinity/product-security` rules + comment) from `.github/CODEOWNERS`.

These paths are already owned by `@dfinity/editorial` via the top-level `* @dfinity/editorial` rule, so effective ownership is unchanged.


marc0olo (approved):
Approving. But please note that the `portal` repo is currently not actively used anymore. So the change is more or less irrelevant and we will deprecate/archive the repo soon.

